PR #1, opened 2023-10-02
add plot_trace

aloctavodia:
This follows the decision that `plot_trace` will be a plot without the densities (kde or histograms).

I started with this example because it seems simple enough to get familiar with the code. Please let me know what I am doing wrong. 

Also not sure what should I do to add divergences, given that divergences are defined in a separate group.

This is how the default plots look like for the `centered_eight` dataset

default

and this is with some customization 

plot_trace(post, var_names=["mu", "tau"], 
                    sample_dims=["draw"],
                    pc_kwargs={"plot_grid_kws": {"figsize": (12, 7), "sharey":"row"}, "col_wrap":4})

customized

### DIFF
--- a/src/arviz_plots/plots/__init__.py
+++ b/src/arviz_plots/plots/__init__.py
@@ -2,4 +2,6 @@
 
 from .posteriorplot import plot_posterior
 
-__all__ = ["plot_posterior"]
+from .traceplot import plot_trace
+
+__all__ = ["plot_posterior", "plot_trace"]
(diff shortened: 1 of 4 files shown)


OriolAbril:
> Also not sure what should I do to add divergences, given that divergences are defined in a separate group.

There is no clear way to go about that yet. I think the clearest way to go about it is allowing to define different data when looping/facetting than when plotting

aloctavodia:
Is this ok to merge? Or what else should I do?

OriolAbril (approved):

Issue #10, opened 2023-10-05
Rename plot_posterior and improve it


aloctavodia:
Other suggested names so far: plot_krushke, plot_book, plot_summarystats, plot_stats, plot_dist

I think `plot_dist` was the "winner" during the last discussion...

make it nicer:
* error when bad keys are present in kwargs
* make it work with datatree and dataarray
* add `var_names` argument and `filter_vars`

PR #19, opened 2023-10-13
rename plot_posterior to plot_dist and small improvements

aloctavodia:
Partially address #10 

* Rename `plot_posterior` to `plot_dist` (still not super happy with the new name) ...
* Make it work with datatree
* add `var_names` argument and `filter_vars`

----
:books: Documentation preview :books:: https://arviz-plots--19.org.readthedocs.build/en/19/


.pre-commit-config.yaml
```diff
@@ -3,6 +3,7 @@ repos:
   rev: v4.3.0
   hooks:
   - id: check-added-large-files
+    args: ['--maxkb=1500']
```
aloctavodia, line 6: I changed this to make the pre-commit pass. I am not sure what a good default could be

tests/test_plots.py
```diff
@@ -14,9 +13,7 @@ def data(seed=31):
     mu = rng.normal(size=(4, 100))
     theta = rng.normal(size=(4, 100, 7))
 
-    return xr.Dataset(
-        {"mu": (["chain", "draw"], mu), "theta": (["chain", "draw", "hierarchy"], theta)},
-    )
+    return convert_to_datatree({"mu": mu, "theta": theta})
```
aloctavodia, line 16: Not opposed to keeping the "hierarchy" name for the dims. It's just that I got an error when using `convert_to_datatree`

.pre-commit-config.yaml
OriolAbril: I am guessing this is for the notebook. I am not sure, we could exclude all notebooks or the docs folder from the check, but I think this is probably better, it will force us to avoid very big notebooks. We can start with this and modify it as necessary.

tests/test_plots.py
OriolAbril: The tuple syntax `(["chain", "draw"], mu)` is an xarray shortcut to avoid using both `data` and `dims` as arguments. We don't have that shortcut for arviz so we'd have to use `dims` as keyword argument. I'd also recommend using `from_dict` (or explicit converters) over the `convert_to_datatree` function as general rule.

PR #27, opened 2023-12-15
[WIP] Add plot_trace_dens

aloctavodia:
This is my attempt to reproduce the legacy plot_trace. 

output_00

we have ecdf too!

output_01
 

And also a lot of mistakes. 

* I added the function `scale_fig_size` (needs more works). Is this the approach we are going to use for the new ArviZ?
* I use matplotlib keywords to get/set default values, which will fail for bokeh or other backends.
* compact=true is not doing the right thing (I guess school is stacked ), see below:

output_03

* I added some logic to set the size of ticks, remove ticks and add x/y labels I am not sure I am doing the right thing, in particular for setting the size of ticks and remove ticks.

* I did not add any code for bokeh...

* Docstring and even variable names may need to be cleaned...

* Should we call this plot_trace_dist instead of plot_dens?

* To reduce redundancy and make the plots more compact, we may want to have a title per row with the variable names and spanning both columns.  Not sure how easy is that

----
📚 Documentation preview 📚: https://arviz-plots--27.org.readthedocs.build/en/27/

### DIFF
--- a/src/arviz_plots/plots/tracedensplot.py
+++ b/src/arviz_plots/plots/tracedensplot.py
@@ -96,7 +96,7 @@ def plot_trace_dens(
 
     figsize, textsize, linewidth = scale_fig_size(
         pc_kwargs.get("plot_grid_kws", {}).get("figsize", None),
-        rows=get_size_of_var(posterior, compact=compact),
+        rows=get_size_of_var(posterior, compact=compact, sample_dims=sample_dims),
         cols=2,
     )
 
@@ -175,7 +175,7 @@ def plot_trace_dens(
             ignore_aes=yticks_dens_ignore,
             coords={"__column__": 0},
             store_artist=False,
-            axis="y", # maybe also be explicit here?
+            axis="y",  # maybe also be explicit here?
         )
 
     # Add varnames as x and y labels
--- a/src/arviz_plots/plots/tracedensplot.py
+++ b/src/arviz_plots/plots/tracedensplot.py
@@ -1,0 +1,203 @@
+"""TraceDens plot code."""
+from arviz_base import rcParams
+from arviz_base.labels import BaseLabeller
+from arviz_base.utils import _var_names
+
+from arviz_plots.plot_collection import PlotCollection
+from arviz_plots.plots.utils import filter_aes
+from arviz_plots.visuals import labelled_x, labelled_y, remove_ticks, line, line_xy, ecdf_line
+
+
+def plot_trace_dens(
+    dt,
+    var_names=None,
+    filter_vars=None,
+    sample_dims=None,
+    compact=True,
+    kind=None,
+    plot_collection=None,
+    backend=None,
+    labeller=None,
+    aes_map=None,
+    dens_kwargs=None,
+    trace_kwargs=None,
+    pc_kwargs=None,
+):
+    """Plot 1D marginal densities on the first row and iteration versus sampled
+    values on the second.
+
+    Parameters
+    ----------
+    dt : DataTree
+        Input data
+    var_names: str or list of str, optional
+        One or more variables to be plotted.
+        Prefix the variables by ~ when you want to exclude them from the plot.
+    filter_vars: {None, “like”, “regex”}, optional, default=None
+        If None (default), interpret var_names as the real variables names.
+        If “like”, interpret var_names as substrings of the real variables names.
+        If “regex”, interpret var_names as regular expressions on the real variables names.
+    sample_dims : iterable, optional
+        Dimensions to reduce unless mapped to an aesthetic.
+        Defaults to ``rcParams["data.sample_dims"]``
+    compact: bool, optional
+        Plot multidimensional variables in a single plot. Defaults to True
+    kind : {"kde", "hist", "dot", "ecdf"}, optional
+        How to represent the marginal density.
+    plot_collection : PlotCollection, optional
+    backend : {"matplotlib", "bokeh"}, optional
+    labeller : labeller, optional
+    aes_map : mapping, optional
+        Mapping of artists to aesthetics that should use their mapping in `plot_collection`
+        when plotted. Defaults to only mapping properties to the density representation.
+    trace_kwargs : mapping, optional
+        Valid keys are:
+
+        * trace -> passed to visuals.line
+        * divergence -> passed to visuals.line_x
+
+    pc_kwargs : mapping
+        Passed to :class:`arviz_plots.PlotCollection`
+
+    Returns
+    -------
+    PlotCollection
+    """
+    if sample_dims is None:
+        sample_dims = rcParams["data.sample_dims"]
+    if kind is None:
+        kind = rcParams["plot.density_kind"]
+    if dens_kwargs is None:
+        dens_kwargs = {}
+    if trace_kwargs is None:
+        trace_kwargs = {}
+    if pc_kwargs is None:
+        pc_kwargs = {}
+    else:
+        pc_kwargs = pc_kwargs.copy()
+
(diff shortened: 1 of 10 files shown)


src/arviz_plots/plots/tracedensplot.py
```diff
@@ -0,0 +1,241 @@
...
+    var_names = _var_names(var_names, dt.posterior.ds, filter_vars)
+
+    if var_names is None:
+        posterior = dt.posterior.ds
+    else:
+        posterior = dt.posterior.ds[var_names]
+
+    pc_kwargs.setdefault("aes", {"color": ["chain"]})
+
+    aux_dim_list = [dim for dim in posterior.dims if dim not in {"chain", "draw"}]
```
OriolAbril, line 95:
```suggestion
    aux_dim_list = [dim for dim in posterior.dims if dim not in sample_dims]
```

src/arviz_plots/plots/tracedensplot.py
```diff
@@ -0,0 +1,241 @@
...
+        pc_kwargs = pc_kwargs.copy()
+
+    var_names = _var_names(var_names, dt.posterior.ds, filter_vars)
+
+    if var_names is None:
+        posterior = dt.posterior.ds
+    else:
+        posterior = dt.posterior.ds[var_names]
+
+    pc_kwargs.setdefault("aes", {"color": ["chain"]})
```
OriolAbril, line 93: Do we want this to happen always or only when compact=False. We could use different linestyle (current behaviour iirc) or the "overlay"/special keyword to mean loop over that but no aesthetic change.

src/arviz_plots/plots/tracedensplot.py
```diff
@@ -0,0 +1,241 @@
...
+    # Remove yticks, only for KDEs
+    if kind == "kde":
+        _, _, yticks_dens_ignore = filter_aes(plot_collection, aes_map, "yticks_dens", sample_dims)
+
+        plot_collection.map(
+            remove_ticks,
+            "yticks_dens",
+            ignore_aes=yticks_dens_ignore,
+            coords={"__column__": 0},
+        )
```
OriolAbril, line 177: This doesn't return anything, so it should not be stored:

```suggestion
        plot_collection.map(
            remove_ticks,
            "yticks_dens",
            ignore_aes=yticks_dens_ignore,
            coords={"__column__": 0},
            store_artist=False,
            axis="y", # maybe also be explicit here?
        )
```

Side note, currently, using `None` as the 2nd argument (`fun_label`) uses `fun.__name__` as default label for the stored artist. Do you think using `None` should instead mean don't store it? Then we could remove the `store_artist` argument from .map

src/arviz_plots/plots/tracedensplot.py
```diff
@@ -0,0 +1,241 @@
...
+        pc_kwargs.setdefault("cols", ["__column__"])
+        pc_kwargs.setdefault("plot_grid_kws", {"figsize": figsize})
+
+        if compact:
+            pc_kwargs.setdefault("rows", ["__variable__"])
+        else:
+            pc_kwargs.setdefault("rows", ["__variable__"] + aux_dim_list)
+
+        plot_collection = PlotCollection.grid(
+            posterior.expand_dims(__column__=2),
```
OriolAbril, line 118:
```suggestion
            posterior.expand_dims(__column__=2).assign_coords(__column__=["dens", "trace"]),
```

do you think this would make the code more clear below? We'd have `coords={__column__="dens"/"trace"}` instead of `coords={__column__=0/1}`

src/arviz_plots/plots/tracedensplot.py
```diff
@@ -0,0 +1,241 @@
...
+        ignore_aes=labels_dens_ignore,
+        coords={"__column__": 0},
+        subset_info=True,
+        labeller=labeller,
+        **labels_dens_kwargs,
+    )
+
+    plot_collection.map(
+        labelled_y,
+        "labels_dens",
```
OriolAbril, line 202: if this doesn't error out I should add a check to prevent repeated artist names.

src/arviz_plots/plots/tracedensplot.py
```diff
@@ -0,0 +1,241 @@
...
+        coords={"__column__": 1},
+        subset_info=True,
+        labeller=labeller,
+        **labels_dens_kwargs,
+    )
+
+    # Adjust ticks size
+    plot_collection.map(
+        ticks_size,
+        "labels_dens",
```
OriolAbril, line 213: also repeated artist name for storing, and this one should not be stored.

src/arviz_plots/plots/utils.py
```diff
@@ -19,3 +19,60 @@ def filter_aes(pc, aes_map, artist, sample_dims):
     _, all_loop_dims = pc.update_aes(ignore_aes=ignore_aes)
     artist_dims = [dim for dim in sample_dims if dim not in all_loop_dims]
     return artist_dims, artist_aes, ignore_aes
+
+
+def get_size_of_var(group, compact=False):
+    """Get the size of the variables in a group."""
+    coords = group.coords._names - {"chain", "draw"}  # pylint: disable=protected-access
```
OriolAbril, line 26: I think this could/should? be 

```suggestion
    coords = set(group.sizes) - set(sample_dims)
```

src/arviz_plots/plots/tracedensplot.py
```diff
@@ -0,0 +1,241 @@
...
+    else:
+        posterior = dt.posterior.ds[var_names]
+
+    pc_kwargs.setdefault("aes", {"color": ["chain"]})
+
+    aux_dim_list = [dim for dim in posterior.dims if dim not in {"chain", "draw"}]
+
+    figsize, textsize, linewidth = scale_fig_size(
+        pc_kwargs.get("plot_grid_kws", {}).get("figsize", None),
+        rows=get_size_of_var(posterior, compact=compact),
```
OriolAbril, line 99: We might need a better way to go about this for other plots. The `wrap`/`grid` classmethods already know this before creating the figure+subplots, but there are no intermediate steps right now.

src/arviz_plots/plots/utils.py
```diff
@@ -19,3 +19,60 @@ def filter_aes(pc, aes_map, artist, sample_dims):
...
+        Size of figure in inches
+    labelsize : int
+        fontsize for labels
+    linewidth : int
+        linewidth
+    """
+    # we should read figsize from rcParams or bokeh theme
+    if figsize is None:
+        width, height = 8, 2
+        height = (rows + 1) ** 1.1
```
OriolAbril, line 67:
```suggestion
        width = 8
        height = (rows + 1) ** 1.1
```

I think that is the same

src/arviz_plots/plots/utils.py
```diff
@@ -19,3 +19,60 @@ def filter_aes(pc, aes_map, artist, sample_dims):
...
+            for key in coords:
+                if key in dims_size:
+                    partial_sum *= dims_size[key]
+
+        var_size += partial_sum
+
+    return var_size
+
+
+def scale_fig_size(figsize, rows=1, cols=1):
```
OriolAbril, line 41: is this valid for both bokeh and matplotlib?

OriolAbril (commented):
I'll try to use it this weekend and see if I have any more comments.

src/arviz_plots/plots/utils.py
aloctavodia: It should

OriolAbril:
> Should we call this plot_trace_dist instead of plot_dens

me gusta más con dist, si

aloctavodia:
Merging this. We can keep working on separate PRs, I will open issues.

PR #32, opened 2024-02-23
Add legend method to PlotCollection and improve docs

OriolAbril:
I have started working on a legend method for `PlotCollection`, so far it works with matplotlib,
but still quite unclear what the ideal behaviour should be.

In parallel, I plan to extend the docs of mostly everything.

----
📚 Documentation preview 📚: https://arviz-plots--32.org.readthedocs.build/en/32/


OriolAbril:
Should be ready to merge, we can continue iterating on the legend method and on documentation.

PR #35, opened 2024-03-28
Add neutral element concept to aes generation

### LINKED ISSUE #33: Behaviour of aesthetics mapping for dimension subset

OriolAbril:
Closes #33

----
📚 Documentation preview 📚: https://arviz-plots--35.org.readthedocs.build/en/35/

### DIFF
--- a/src/arviz_plots/plot_collection.py
+++ b/src/arviz_plots/plot_collection.py
@@ -360,22 +360,26 @@ class PlotCollection:
                     )
                     aes_cumulative += n_aes
             else:
+                aes_shape = [self.data.sizes[dim] for dim in dims]
+                aes_dims_in_var = {var_name: set(dims) <= set(da.dims) for var_name, da in self.data.items()}
+                if not any(aes_dims_in_var.values()):
+                    warnings.warning("Provided mapping for {aes_key} will only use the neutral element")
                 total_aes_vals = int(
-                    np.prod([self.data.sizes[dim] for dim in dims])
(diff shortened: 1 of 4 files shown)


OriolAbril:
I have to update [`_get_aes_dict_from_dt`](https://github.com/arviz-devs/arviz-plots/blob/main/src/arviz_plots/plot_collection.py#L96) and add tests for it.

Issue #24, opened 2023-12-13
Add histogram to plot_dist


aloctavodia:

imperorrp:
I can try working on this issue

aloctavodia:
great! Feel free to ask questions if you have doubts.

imperorrp:
Sure, thank you! 

I had a closer look at the existing codebase thinking about how to approach this and did have some doubts about how to build this. 

In the existing implementation in `src/arviz-plots/plots/distplot.py`, `.map()` is called on the PlotCollection instance to create the plot elements with the visual element functions `line_xy()` and `ecdf_line()` passed to it, right? 

For an `if kind == 'hist':` condition, what visual element function should be passed- would a custom one have to be defined for histogram-ish data? The `def map()` function in the PlotCollection class only accepts one visual element function at a time, so a new visual element function for histogram plots? I suppose a combination of `line_x()` and `line()` could also work (horizontal and vertical lines), with some processing required to define intervals based on number of bins wanted. 

I checked the old ArviZ implementation of histogram plots for distplot and it seems like `ax.hist()` is used for example in the matplotlib implementation and `ax.quad()` with bokeh.

And for number of bins, I suppose the `plot_kwargs` dict could be where users could define a user given number of bins and a preset number of bins set as default?

OriolAbril:
You'd have to follow a pattern similar to the two existing kinds.

1. Compute the histogram of the data
2. use `.map` on the result to plot the histogram

In step 1 you'll have to call the histogram function on the data, with kwargs taken from stats_kwargs. As a starting point, you can use https://einstats.python.arviz.org/en/latest/api/generated/xarray_einstats.numba.histogram.html (xarray-einstats is a dependency of arviz-plots already and numba is too for now). After finishing some more urgent matters in arviz-plots I'll shift my efforts more to arviz-stats, hopefully while I work more on arviz-stats we can gather feedback on the current plots.

Then in step 2 call `.map` on a new visual element and in this particular case, this new visual element will have to call a new function in the `backend` module that would have to be added for both matplotlib and bokeh (probably bars and quad respectively). Here is where kwargs are taken from plot_kwargs.

---

Important: I have just opened a PR where I modify a bit `plot_dist` if you start working now you might have to deal with merge conflicts. They should not be a huge issue but it can be a bit annoying. I'll try to merge the PR tomorrow so you can work on plot_dist without worrying about anything else.

imperorrp:
[...]

PR #47, opened 2024-04-15
[WIP] Histogram support addition to distplot.py

imperorrp:
This PR aims to add support for histograms in `distplot.py` (Issue #24 ), allowing marginal densities to be visualized in this new form, apart from preexisting KDE and ECDF forms

----
📚 Documentation preview 📚: https://arviz-plots--47.org.readthedocs.build/en/47/

### DIFF
--- a/src/arviz_plots/backend/bokeh/__init__.py
+++ b/src/arviz_plots/backend/bokeh/__init__.py
@@ -1,4 +1,5 @@
 """Bokeh interface layer."""
+
 import warnings
 
 import numpy as np
@@ -228,6 +229,20 @@ def _float_or_str_size(size):
 
 
 # "geoms"
+def hist(
+    y, l_e, r_e, target, *, bottom=0, color=unset, facecolor=unset, edgecolor=unset, **artist_kws
+):
+    """Interface to Bokeh for a histogram bar plot."""
+    artist_kws.setdefault("level", "glyph")
+    if color is not unset:
+        if facecolor is unset:
+            facecolor = color
+        if edgecolor is unset:
+            edgecolor = color
+    kwargs = {"bottom": bottom, "fill_color": facecolor, "line_color": edgecolor}
+    return target.quad(top=y, left=l_e, right=r_e, **_filter_kwargs(kwargs, artist_kws))
+
+
 def line(x, y, target, *, color=unset, alpha=unset, width=unset, linestyle=unset, **artist_kws):
     """Interface to bokeh for a line plot."""
     kwargs = {"color": color, "alpha": alpha, "line_width": width, "line_dash": linestyle}
--- a/src/arviz_plots/backend/bokeh/__init__.py
+++ b/src/arviz_plots/backend/bokeh/__init__.py
@@ -230,7 +230,17 @@ def _float_or_str_size(size):
 
 # "geoms"
 def hist(
-    y, l_e, r_e, target, *, bottom=0, color=unset, facecolor=unset, edgecolor=unset, **artist_kws
+    y,
+    l_e,
+    r_e,
+    target,
+    *,
+    bottom=0,
+    color=unset,
+    facecolor=unset,
+    edgecolor=unset,
+    alpha=unset,
+    **artist_kws,
 ):
     """Interface to Bokeh for a histogram bar plot."""
     artist_kws.setdefault("level", "glyph")
@@ -239,7 +249,7 @@ def hist(
             facecolor = color
         if edgecolor is unset:
             edgecolor = color
-    kwargs = {"bottom": bottom, "fill_color": facecolor, "line_color": edgecolor}
+    kwargs = {"bottom": bottom, "fill_color": facecolor, "line_color": edgecolor, "alpha": alpha}
     return target.quad(top=y, left=l_e, right=r_e, **_filter_kwargs(kwargs, artist_kws))
 
 
--- a/src/arviz_plots/backend/bokeh/__init__.py
+++ b/src/arviz_plots/backend/bokeh/__init__.py
@@ -243,7 +243,6 @@ def hist(
     **artist_kws,
 ):
     """Interface to Bokeh for a histogram bar plot."""
-    artist_kws.setdefault("level", "glyph")
     if color is not unset:
         if facecolor is unset:
             facecolor = color
--- a/src/arviz_plots/backend/matplotlib/__init__.py
+++ b/src/arviz_plots/backend/matplotlib/__init__.py
@@ -211,7 +211,7 @@ def hist(
         if edgecolor is unset:
             edgecolor = color
     kwargs = {"bottom": bottom, "color": facecolor, "edgecolor": edgecolor}
-    return target.bar(l_e, y, width=widths ** _filter_kwargs(kwargs, None, artist_kws))
+    return target.bar(l_e, y, width=widths, **_filter_kwargs(kwargs, None, artist_kws))
 
 
 def line(x, y, target, *, color=unset, alpha=unset, width=unset, linestyle=unset, **artist_kws):
--- a/src/arviz_plots/backend/matplotlib/__init__.py
+++ b/src/arviz_plots/backend/matplotlib/__init__.py
@@ -5,9 +5,11 @@ Notes
 Sets ``zorder`` of all non-text "geoms" to ``2`` so that elements plotted later
 on are on top of previous ones.
 """
+
 import warnings
 from typing import Any, Dict
 
+import numpy as np
 from matplotlib.cbook import normalize_kwargs
 from matplotlib.collections import PathCollection
 from matplotlib.lines import Line2D
@@ -202,15 +204,14 @@ def hist(
 ):
     """Interface to matplotlib for a histogram bar plot."""
     artist_kws.setdefault("zorder", 2)
-    bins = list(l_e) + [r_e[-1]]
-    midpoints = [(bins[i] + bins[i + 1]) / 2 for i in range(len(bins) - 1)]
+    widths = np.asarray(r_e) - np.asarray(l_e)
     if color is not unset:
         if facecolor is unset:
             facecolor = color
         if edgecolor is unset:
             edgecolor = color
     kwargs = {"bottom": bottom, "color": facecolor, "edgecolor": edgecolor}
-    return target.bar(midpoints, y, **_filter_kwargs(kwargs, None, artist_kws))
+    return target.bar(l_e, y, width=widths ** _filter_kwargs(kwargs, None, artist_kws))
 
 
 def line(x, y, target, *, color=unset, alpha=unset, width=unset, linestyle=unset, **artist_kws):
--- a/src/arviz_plots/backend/matplotlib/__init__.py
+++ b/src/arviz_plots/backend/matplotlib/__init__.py
@@ -216,7 +216,9 @@ def hist(
     artist_kws.setdefault("zorder", 2)
     widths = np.asarray(r_e) - np.asarray(l_e)
     if np.any(bottom != 0):
-        y = y - bottom  # making y the top coordinate and not height
+        height = y - bottom
+    else:
+        height = y
     if color is not unset:
         if facecolor is unset:
             facecolor = color
@@ -224,7 +226,7 @@ def hist(
             edgecolor = color
     kwargs = {"bottom": bottom, "color": facecolor, "edgecolor": edgecolor, "alpha": alpha}
     return target.bar(
-        l_e, y, width=widths, align="edge", **_filter_kwargs(kwargs, None, artist_kws)
+        l_e, height, width=widths, align="edge", **_filter_kwargs(kwargs, None, artist_kws)
     )
 
 
--- a/src/arviz_plots/backend/matplotlib/__init__.py
+++ b/src/arviz_plots/backend/matplotlib/__init__.py
@@ -197,17 +197,20 @@ def _filter_kwargs(kwargs, artist, artist_kws):
 
 
 # "geoms"
-def hist(y, l_e, r_e, target, *, bottom=None, color=None, facecolor=None, edgecolor=None, **kwargs):
+def hist(
+    y, l_e, r_e, target, *, bottom=None, color=unset, facecolor=unset, edgecolor=unset, **artist_kws
+):
     """Interface to matplotlib for a histogram bar plot."""
-    kwargs.setdefault("zorder", 2)
+    artist_kws.setdefault("zorder", 2)
     bins = list(l_e) + [r_e[-1]]
     midpoints = [(bins[i] + bins[i + 1]) / 2 for i in range(len(bins) - 1)]
-    if color is not None:
-        if facecolor is None:
+    if color is not unset:
+        if facecolor is unset:
             facecolor = color
-        if edgecolor is None:
+        if edgecolor is unset:
             edgecolor = color
-    return target.bar(midpoints, y, bottom=bottom, color=facecolor, edgecolor=edgecolor, **kwargs)
+    kwargs = {"bottom": bottom, "color": facecolor, "edgecolor": edgecolor}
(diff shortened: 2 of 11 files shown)


imperorrp:
@OriolAbril I had to dig a bit deeper into understanding how the plotting actually works as coded, but I think I've managed to get step 1 (computing the histogram) roughly done- I'll probably have to combine these DataArrays into a Dataset and then create the visual element for mapping. Also, I'll remove the `print` statements I've added in later commits- just added them to help with seeing the outputs.  Is something like this alright?- Looping through the data variables in `distribution` to get DataArrays and then passing that plus the dimensions to be reduced to the xarray_einstats.histogram() function. 

Also, I was going through the flow of how density was computed for KDE as an example (in Arviz-Stats). I saw bandwidth computing was done for KDE- but that won't be necessary for plotting histograms, right?

imperorrp:
Combined the `xarray_einstats.numba.histogram` produced DataArrays into a Dataset: 

Dataset-image

(This is with the Schools dataset) 

It's in the style of the KDE dataset returned by `distribution.azstats.kde` to density

imperorrp:
Added a visual element function, backend interface and matplotlib implementation with `matplotlib.pyplot.bar`:

image

The bar widths are different due to the different scales of the x-axis. I'm not sure how this can be fixed yet if keeping the width a consistent value is desired though. Also, the number of bins is set to 20 by default. Some bins have too few values to be visible as their respective bars are too low in height compared to the taller ones.

src/arviz_plots/plots/distplot.py
```diff
@@ -250,6 +257,33 @@ def plot_dist(
...
+            hist_dict = {}
+            # loops through the data variables in distribution and calls histogram() for each
+            for var_name in distribution.data_vars:
+                # print(f"    Var name: {var_name!r}\n")
+                var_data = distribution[var_name]
+                # print(f"    Var data: {var_data!r}\n")
+
+                # number of bins is set to 20 by default
+                hist = histogram(
+                    da=var_data, dims=density_dims, bins=20, **stats_kwargs.get("density", {})
```
OriolAbril, line 272: the default number of bins should be taken care of by `histogram` function. Moreover, if we believed a different default were needed, it would have to happen via `stats_kwargs` and `setdefault` so a user could modify that value.

src/arviz_plots/backend/__init__.py
```diff
@@ -99,6 +99,29 @@ def create_plotting_grid(
 
 
 # "geoms"
+def hist(x, height, target, *, width=1.6, bottom=None, align="center", color=None):
```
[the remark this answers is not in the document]
OriolAbril: This will require some changes to the data restructure utility function but I am not too worried about this, at some point we'll have a function in arviz-stats to take care of all this.

src/arviz_plots/backend/__init__.py
```diff
@@ -99,6 +99,29 @@ def create_plotting_grid(
...
+    """Interface for a hist plot.
+
+    Parameters
```
OriolAbril, line 105: We haven't been completely consistent adding the docstrings to this module, but the idea is for the documentation here so define the behaviour **all** backends have to comply with. It also isn't a general plotting backend interface but a common interface _for our own plots to use_ which means we can be quite restrictive about the features.

For example, I would only accept array_like for x, height, width... I don't think the extra edge case of floats would be used anywhere. It also means the descriptions should not depend on any backend.

src/arviz_plots/backend/__init__.py
```diff
@@ -99,6 +99,29 @@ def create_plotting_grid(
...
+    target : matplotlib.axes._axes.Axes
+        The target Axes object where the plot will be drawn.
+    width : float or array-like, default 0.8
+        The width(s) of the bars.
+    bottom : float or array-like, optional
+        The y coordinate(s) of the bottom side(s) of the bars.
+    align : {'center', 'edge'}, default 'center'
+        Alignment of the bars to the x coordinates.
+    color : color or array-like, optional
+        The colors of the bar faces.
```
OriolAbril, line 120: I might start with the same color for all rectangles, mostly because I am not sure how easy it will be to support the different color rectangles in all backends. If we then see it is feasible and of interest we can add that extra case.

There should also be a facecolor and an edgecolor arguments given there is a color argument and the generated form is "fillable" to match the expectations from https://arviz-plots.readthedocs.io/en/latest/api/backend/index.html#backend-interface-arguments

OriolAbril (commented):
It is looking very good already. The computation of the histogram is a bit trickier and would not be part of the gsoc project, but I think it will be helpful to get you more familiar with xarray.

I have tried to focus on the plotting and contributing workflow with the comments. Let me know if you have any doubts related to testing.

imperorrp:
> It is looking very good already. The computation of the histogram is a bit trickier and would not be part of the gsoc project, but I think it will be helpful to get you more familiar with xarray.
> 
> I have tried to focus on the plotting and contributing workflow with the comments. Let me know if you have any doubts related to testing.

Thank you, I will take a look at your reviews and advice!

src/arviz_plots/plots/distplot.py
imperorrp: Just modified this in the latest commit to do so 

Now by default, with no bins provided (`azp.plot_dist(schools, kind='hist')`) and the same example dataset, this is produced: 
distplot-hist-matplotlib

And with bins=40 provided by a user (`azp.plot_dist(schools, kind='hist', stats_kwargs={"density": {"bins":40}})`), this is produced:
distplot-hist-matplotlib

Will look at the other comments and make changes too now

imperorrp:
@OriolAbril Latest commit has changes incorporating advice from your comments. Modified data restructure utility function to keep left and right edges info now until the matplotlib backend calculates midpoints from these at the end to create the bars, and I've updated the docstring and color/facecolor/edgecolor interface

src/arviz_plots/backend/matplotlib/__init__.py
```diff
@@ -200,6 +200,22 @@ def _filter_kwargs(kwargs, artist, artist_kws):
...
+    artist_kws.setdefault("zorder", 2)
+    bins = list(l_e) + [r_e[-1]]
+    midpoints = [(bins[i] + bins[i + 1]) / 2 for i in range(len(bins) - 1)]
+    if color is not unset:
+        if facecolor is unset:
+            facecolor = color
+        if edgecolor is unset:
+            edgecolor = color
+    kwargs = {"bottom": bottom, "color": facecolor, "edgecolor": edgecolor}
+    return target.bar(midpoints, y, **_filter_kwargs(kwargs, None, artist_kws))
```
OriolAbril, line 216: To minimize computation, I would do the following:

```
widths = np.asarray(r_e) - np.asarray(l_e)
...
return target.bar(l_e, y, width=widths, align="edge", _**filter_...)
```

The bin edges of a histogram don't necessarly need to be equispaced, so we do need to use array-like widths. We don't need to compute the midpoints though, using `align="edge"` we can use the left edges which we already have as x values.

OriolAbril (commented):
Left a comment on how to use matplotlib bar function to generate histogram looking plot. Let me know if you need any help to generate the bokeh equivalent

src/arviz_plots/backend/matplotlib/__init__.py
imperorrp: Okay, I'll add this change and push

imperorrp: This one definitely works better, the gaps between histogram bars seem eliminated now 

image



image

OriolAbril:
By the way, I commented on the ess and mcse issues but forgot to comment here. There is a histogram function available in `arviz-stats` now. I think the output will have a similar (if not the same) shape to what you are using

imperorrp:
> By the way, I commented on the ess and mcse issues but forgot to comment here. There is a histogram function available in `arviz-stats` now. I think the output will have a similar (if not the same) shape to what you are using

Altered the code to make use of this function. Also, it seems like point estimate and credible intervals are being plotted so close to the x axis for histogram plots because the default 'y' aesthetic value/s returned by plot collection are pretty low in relation. The maximum y-axis values for the KDE/ECDF plots are way lower so it's not a problem for them. I initially thought there was some issue with this aesthetic mapping when type hist is picked, but it works fine- it's just really small values compared to the histogram heights. Should we mention this in the documentation, asking users to define custom 'y' aesthetic values (in `pc_kwargs`)when using kind=histogram for better visibility? 

For the plot_dist example where one subplot is created for each variable:
```
azp.plot_dist(
        data,
        kind="kde",
        pc_kwargs={
            "cols": ["__variable__"],
            "aes": {"color": ["school"], "y": ["school"]},
            "y": np.linspace(0, 0.06, 9),
        },
        aes_map={
            "kde": ["color"],
            "point_estimate": ["color", "y"],
            "credible_interval": ["y"],
        },
    )
    plt.savefig("distplot-kde-cols-matplotlib.png")
```
image

Point estimates when switching from kind="kde" to kind="hist":
```
azp.plot_dist(
        data,
        kind="hist",
        pc_kwargs={
            "cols": ["__variable__"],
            "aes": {"color": ["school"], "y": ["school"]},
            "y": np.linspace(0, 0.06, 9),
        },
        aes_map={
            "kde": ["color"],
            "point_estimate": ["color", "y"],
            "credible_interval": ["y"],
        },
    )
    plt.savefig("distplot-hist-cols-matplotlib.png")
```
image

This is easily fixable with editing the `np.linspace` line though-
```
"y": np.linspace(0, 600, 9),
```
image

We could maybe add this in the documentation so users know to keep this in mind.

src/arviz_plots/plots/distplot.py
```diff
@@ -250,6 +253,16 @@ def plot_dist(
                 **density_kwargs,
             )
 
+        elif kind == "hist":
+            density = distribution.azstats.histogram(
+                dims=density_dims, **stats_kwargs.get("density", {})
+            )
```
OriolAbril, line 259: we should pass `density=True` as default here. This will normalize the histogram so its area is 1, and will therefore generate histograms with heights similar to those of the kde. That will should take care of the non-reusability of the overlayed example when changing hist/kde. For ecdf/dotplots it will probably break too, but I don't think it matters, ecdf always goes 0-1 whereas dotplots are particularly strange, they should not be expected to act as drop-in replacements of either kde nor hist.

imperorrp: Yeah adding that seems to work for the overlay example, without having to alter the `np.linspace` parameters.

src/arviz_plots/plots/distplot.py
```diff
@@ -97,6 +99,7 @@ def plot_dist(
 
           * "kde" -> passed to :func:`~arviz_plots.visuals.line_xy`
           * "ecdf" -> passed to :func:`~arviz_plots.visuals.ecdf_line`
+          * "hist" -> passed to :func: `~WIP`
```
OriolAbril, line 102: update this and add the new visual function to https://github.com/arviz-devs/arviz-plots/blob/main/docs/source/api/visuals.rst

src/arviz_plots/plots/distplot.py
```diff
@@ -250,6 +253,16 @@ def plot_dist(
                 **density_kwargs,
             )
 
+        elif kind == "hist":
+            density = distribution.azstats.histogram(
+                dims=density_dims, density=True, **stats_kwargs.get("density", {})
```
OriolAbril, line 258: `density=True` should be added to `stats_kwargs` with `setdefault`. Then, if a user wanted to they can set it to False, keep the axis and use the y axis to look at counts for example. The task itself isn't really important, what is is that it should be possible

src/arviz_plots/plots/distplot.py
```diff
@@ -352,7 +374,7 @@ def plot_dist(
             labeller=labeller,
             **title_kwargs,
         )
-    if (kind == "kde") and (plot_kwargs.get("remove_axis", True) is not False):
+    if (kind in ("kde", "hist")) and (plot_kwargs.get("remove_axis", True) is not False):
```
OriolAbril, line 377: I would modify this logic. Instead, I would set the value of the `remove_axis` depending on `kind` at the top, right after copying `plot_kwargs`.

```python
if kind in ("hist", "ecdf"):
    plot_kwargs.setdefault("remove_axis", False)
```

I think as a default we'll want to remove the axis for kde and quantile dotplot but keep it for the others.

src/arviz_plots/visuals/__init__.py
```diff
@@ -13,6 +13,21 @@
 from arviz_base.labels import BaseLabeller
 
 
+def plot_hist(da, target, backend, **kwargs):
```
OriolAbril, line 16: I would use `hist` only to "reserve" the `plot_xyz` for batteries included functions, otherwise it might get confusing.

src/arviz_plots/backend/__init__.py
```diff
@@ -99,6 +99,33 @@ def create_plotting_grid(
 
 
 # "geoms"
+def hist(y, l_e, r_e, target, *, bottom=None, color=None, facecolor=None, edgecolor=None):
```
OriolAbril, line 102: I went over all potential keywords in https://arviz-plots.readthedocs.io/en/latest/api/backend/index.html#backend-interface-arguments and `alpha` should be added here too.

OriolAbril (commented):
As always, go over all comments before starting to address any single one of them

imperorrp:
Just pushed changes per last review

tests/test_plots.py
```diff
@@ -85,29 +85,32 @@ def datatree_sample(seed=31):
...
-        assert "kde" in pc.viz["mu"]
+        assert kind in pc.viz["mu"]
         assert "hierarchy" not in pc.viz["mu"].dims
         assert "hierarchy" in pc.viz["theta"].dims
         assert "hierarchy" not in pc.viz["mu"]["point_estimate"].dims
         assert "hierarchy" in pc.viz["theta"]["point_estimate"].dims
 
-    def test_plot_dist_models(self, datatree, datatree2, backend):
-        pc = plot_dist({"c": datatree, "n": datatree2}, backend=backend)
+    @pytest.mark.parametrize("kind", ["kde", "hist"])
```
OriolAbril, line 108:
```suggestion
    @pytest.mark.parametrize("kind", ["kde"])
```

src/arviz_plots/backend/bokeh/__init__.py
```diff
@@ -235,6 +236,30 @@ def _float_or_str_size(size):
...
+    *,
+    bottom=0,
+    color=unset,
+    facecolor=unset,
+    edgecolor=unset,
+    alpha=unset,
+    **artist_kws,
+):
+    """Interface to Bokeh for a histogram bar plot."""
+    artist_kws.setdefault("level", "glyph")
```
OriolAbril, line 253: is this necessary?

src/arviz_plots/plots/distplot.py
```diff
@@ -178,6 +181,8 @@ def plot_dist(
         kind = rcParams["plot.density_kind"]
     if plot_kwargs is None:
         plot_kwargs = {}
+    if kind in ("hist", "ecdf"):
```
OriolAbril, line 184:
```suggestion
    if kind in ("kde", "dot"):
```

src/arviz_plots/plots/utils.py
```diff
@@ -87,3 +89,56 @@ def filter_aes(pc, aes_map, artist, sample_dims):
     _, all_loop_dims = pc.update_aes(ignore_aes=ignore_aes)
     artist_dims = [dim for dim in sample_dims if dim not in all_loop_dims]
     return artist_dims, artist_aes, ignore_aes
+
+
+# WIP: dropping bin_midpoints calculation and keeping bin_edges
+def restructure_hist_data(hist_dict):
```
OriolAbril, line 95: this can be removed

OriolAbril:
This will need to be rebased and double checked to make it compatible with plotly. Take care of these last two comments left and after that I'll rebase and add more commits to the PR so it works with plotly

Issue #11, opened 2023-10-05
Add plot_ppc


aloctavodia:
Until explicit indexes are supported it won’t be possible to have the groupby like behaviour

OriolAbril:
I also propose to completely get rid of the "mean" line. Nobody understands what it actually means and there are many cases where it's range is much larger than each individual sample ending up in very strange/ugly plots.

aloctavodia:
Can we keep it, and make it default to False?

OriolAbril:
maybe if we rename the argument and legend entry?

aloctavodia:
what name(s) do you have in mind?

OriolAbril:
I think we should avoid the word "mean" because it makes people think about the mean of the posterior predictive. "combined" seems to be more intuitive in plot_trace, but here we _combine_ over samples instead of over chain only, "aggregate" might also help with avoiding relating that to the actual mean.

aloctavodia:
Aggregate can be a good choice

imperorrp:
Plot_ppc support for discrete values might have to be thought of as well: https://github.com/arviz-devs/arviz/issues/1103

imperorrp:
Some more plot_ppc related issues with legacy Arviz plot_ppc:

A bug when zooming in: 
https://github.com/arviz-devs/arviz/issues/1756  

A feature request for adding HDI-style credible intervals for plot_ppc:
https://github.com/arviz-devs/arviz/issues/1654

OriolAbril:
> Plot_ppc support for discrete values might have to be thought of as well: https://github.com/arviz-devs/arviz/issues/1103

Discussion related to better support for discrete values should go to #53.

PR #55, opened 2024-06-21
[WIP] Adding PPC plot to Arviz-Plots 

imperorrp:
Adding plot_ppc (issue https://github.com/arviz-devs/arviz-plots/issues/11). 

This is a PPC plot draft that currently plots the actual observed values and prior/posterior predictive values, flattening across all dimensions except sample dims (chain, draw) by default 

### WIP: 
- The prior/posterior predictive mean
- Cumulative and scatter visualization 

----
📚 Documentation preview 📚: https://arviz-plots--55.org.readthedocs.build/en/55/

### DIFF
--- a/src/arviz_plots/plots/ppcplot.py
+++ b/src/arviz_plots/plots/ppcplot.py
@@ -186,6 +186,15 @@ def plot_ppc(
             plot element"""
         )
 
+    observed_rug_kwargs = copy(plot_kwargs.get("observed_rug", {}))
+    if observed_rug_kwargs is False:
+        raise ValueError(
+            """plot_kwargs['observed_rug'] can't be False, use observed_rug=False to remove 
+            observed_rug plot element"""
+        )
+    if observed is False and observed_rug is True:
+        raise ValueError("""observed_rug=True is only valid when observed=True""")
+
     aggregate_kwargs = copy(plot_kwargs.get("aggregate", {}))
     if aggregate_kwargs is False:
         raise ValueError(
--- a/src/arviz_plots/plots/ppcplot.py
+++ b/src/arviz_plots/plots/ppcplot.py
@@ -234,12 +234,6 @@ def plot_ppc(
         coords=coords,
     )
 
-    # creating random number generator
-    if random_seed is not None:
-        rng = np.random.default_rng(random_seed)
-    else:
-        rng = np.random.default_rng()
-
     # picking random pp dataset sample indexes and subsetting pp_distribution accordingly
     total_pp_samples = np.prod(
         [pp_distribution.sizes[dim] for dim in sample_dims if dim in pp_distribution.dims]
@@ -257,19 +251,40 @@ def plot_ppc(
     ):
         raise TypeError(f"`num_pp_samples` must be an integer between 1 and {total_pp_samples}.")
 
-    # stacking sample dimensions and selecting randomly if sample_dims length>1 or
-    # num_pp_samples=total_pp_samples
-    if num_pp_samples != total_pp_samples or len(sample_dims) == 1:
-        pp_sample_ix = rng.choice(total_pp_samples, size=num_pp_samples, replace=False)
+    # if num_pp_samples==1, no stacking or subselecting required
 
-        # stacking sample dims into a new 'ppc_dim' dimension
-        pp_distribution = pp_distribution.stack(ppc_dim=sample_dims)
+    # stacking sample dimensions and subselecting from
+    # if len(sample_dims)>1 and num_pp_samples!=total_pp_samples
 
-        # Select the desired samples
-        pp_distribution = pp_distribution.isel(ppc_dim=pp_sample_ix)
+    # subselecting from without stacking if len(sample_dims)==1 and
+    # num_pp_samples!=total_pp_samples
 
-        # renaming sample_dims so that rest of plot will consider this as sample_dims
-        sample_dims = ["ppc_dim"]
+    if num_pp_samples != total_pp_samples:
+        # creating random number generator for random subselecting
+        if random_seed is not None:
+            rng = np.random.default_rng(random_seed)
+        else:
+            rng = np.random.default_rng()
+
+        if len(sample_dims) > 1:
+            # stacking into one dim (ppc_dim) required before subselecting
+            pp_sample_ix = rng.choice(total_pp_samples, size=num_pp_samples, replace=False)
+
+            # stacking sample dims into a new 'ppc_dim' dimension
+            pp_distribution = pp_distribution.stack(ppc_dim=sample_dims)
+
+            # Select the desired samples
+            pp_distribution = pp_distribution.isel(ppc_dim=pp_sample_ix)
+
+            # renaming sample_dims so that rest of plot will consider this as sample_dims
+            sample_dims = ["ppc_dim"]
+        elif len(sample_dims) == 1:
+            # subselecting without stacking
+            pp_sample_ix = rng.choice(total_pp_samples, size=num_pp_samples, replace=False)
+
+            # Select the desired samples
+            sample_dim = sample_dims[0]  # provided dimension used for subsetting
+            pp_distribution = pp_distribution.isel({sample_dim: pp_sample_ix})
 
     # wrap plot collection with pp distribution
     if plot_collection is None:
--- a/src/arviz_plots/plots/ppcplot.py
+++ b/src/arviz_plots/plots/ppcplot.py
@@ -40,7 +40,7 @@ def plot_ppc(
     stats_kwargs=None,
     pc_kwargs=None,
 ):
-    """Plot prior/posterior predictive and observed values as kde, cumulative or scatter plots.
+    """Plot prior/posterior predictive and observed values as kde, ecdf, hist, or scatter plots.
 
     Parameters
     ----------
@@ -69,25 +69,24 @@ def plot_ppc(
         Dimensions to loop over in plotting posterior/prior predictive.
         Note: Dims not in sample_dims or facet_dims (below) will be reduced by default.
         Defaults to ``rcParams["data.sample_dims"]``
-    kind : {"kde", "cumulative", "scatter"}, optional
+    kind : {"kde", "ecdf", "hist, "scatter"}, optional
         How to represent the marginal density. Defaults to ``rcParams["plot.density_kind"]``.
     facet_dims : list, optional
         Dimensions to facet over (for which multiple plots will be generated).
-        Defaults to empty list. A warning is raised if `pc_kwargs` is also used to define
-        dims to facet over with the `cols` key and `pc_kwargs` takes precedence.
+        Defaults to empty list.
     data_pairs : dict, optional
         Dictionary containing relations between observed data and posterior/prior predictive data.
-        Dictionary structure:
-            * key = observed data var_name
-            * value = posterior/prior predictive var_name
+        Dictionary keys are variable names corresponding to observed data and dictionary values
+        are variable names corresponding to posterior/prior predictive.
         For example, data_pairs = {'y' : 'y_hat'}.
-        If None, it will assume that the observed data and the posterior/prior predictive data
-        have the same variable name
-    aggregate: bool, optional
+        By default, it will assume that the observed data and the posterior/prior predictive data
+        have the same variable names
+    aggregate: bool, default False
         If True, predictive data will be aggregated over both sample_dims and reduce_dims.
-        Defaults to False.
     num_pp_samples : int, optional
         Number of prior/posterior predictive samples to plot.
+        Defaults to the total sample size (product of sample_dim dimension lengths) or minimum
+        between total sample size and 5 in case of kind='scatter'
     random_seed : int, optional
         Random number generator seed passed to numpy.random.seed to allow reproducibility of the
         plot.
@@ -108,15 +107,16 @@ def plot_ppc(
     plot_kwargs : mapping of {str : mapping or False}, optional
         Valid keys are:
 
-        * "predictive" -> Passed to either of "kde", "cumulative", "scatter" based on `kind`
-        * "observed" -> passed to either of "kde", "cumulative", "scatter" based on `kind`
-        * "aggregate" -> passed to either of "kde", "cumulative", "scatter" based on `kind`
+        * "predictive" -> Passed to either of "kde", "ecdf", "hist", "scatter" based on `kind`
+        * "observed" -> passed to either of "kde", "ecdf", "hist", "scatter" based on `kind`
+        * "aggregate" -> passed to either of "kde", "ecdf", "hist", "scatter" based on `kind`
 
-        Values of the above plot_kwargs keys are passed to one of "kde", "cumulative", "scatter",
+        Values of the above plot_kwargs keys are passed to one of "kde", "ecdf", "hist", "scatter",
         matching the `kind` argument.
 
           * "kde" -> passed to :func:`~arviz_plots.visuals.line_xy`
-          * "cumulative" -> passed to :func:`~arviz_plots.visuals.ecdf_line`
+          * "ecdf" -> passed to :func:`~arviz_plots.visuals.ecdf_line`
+          * "hist" -> passed to :func:`~arviz_plots.visuals.hist_line`
           * "scatter" -> passed to :func: `~arviz_plots.visuals.scatter_x`
 
         * "observed_rug" -> passed to :func:`arviz_plots.visuals.trace_rug`
@@ -126,9 +126,9 @@ def plot_ppc(
     stats_kwargs : mapping, optional
         Valid keys are:
 
-        * predictive -> passed to kde, cumulative, ...
-        * aggregate -> passed to kde, cumulative, ...
-        * observed -> passed to kde, cumulative, ...
+        * predictive -> passed to kde, ecdf, ...
+        * aggregate -> passed to kde, ecdf, ...
+        * observed -> passed to kde, ecdf, ...
 
     pc_kwargs : mapping
         Passed to :class:`arviz_plots.PlotCollection.wrap`
@@ -175,50 +175,22 @@ def plot_ppc(
         raise TypeError("`group` argument must be either `posterior` or `prior`")
 
     predictive_data_group = f"{group}_predictive"
-    default_observed = ""
     if observed is None:
-        default_observed = f"(set as default for chosen group {group})"
         observed = group == "posterior"  # by default true if posterior, false if prior
 
-    # checking if plot_kwargs["observed"] or plot_kwargs["aggregate"] is not inconsistent with
-    # top level bool arguments `observed`` and `aggregate`
-
-    # observed args logic check:
-    # observed will be True/False depending on user-input/group- true for posterior, false for prior
-    # if observed and no plot_kwargs['observed'], no prob
-    # if observed = True and plot_kwargs['observed'] = True, no prob (observed is plotted)
-    # if observed = True and plot_kwargs['observed'] = False, prob (error/warning raised)
-    # if observed = False and plot_kwargs['observed'] = True, prob (error/warning raised)
-    # if observed = False and plot_kwargs['observed'] = False, no prob (observed is not plotted)
-    if (
-        observed
-        and plot_kwargs.get("observed", True) is False
-        or not observed
-        and plot_kwargs.get("observed", False) is not False
-    ):
-        # raise warning or error
+    # checking to make sure plot_kwargs["observed"] or plot_kwargs["aggregate"] are not False
+    observed_kwargs = copy(plot_kwargs.get("observed", {}))
+    if observed_kwargs is False:
         raise ValueError(
-            f"""
-            `observed` and `plot_kwargs["observed"]` inconsistency detected.
-            `observed` = {observed}{default_observed}
-            `plot_kwargs["observed"]` = {plot_kwargs["observed"]}
-            Please make sure `observed` and `plot_kwargs["observed"]` have the same value."""
+            """plot_kwargs['observed'] can't be False, use observed=False to remove observed 
+            plot element"""
         )
 
-    # same check for aggregate:
-    if (
-        aggregate
-        and plot_kwargs.get("aggregate", True) is False
-        or not aggregate
-        and plot_kwargs.get("aggregate", False) is not False
-    ):
-        # raise warning or error
+    aggregate_kwargs = copy(plot_kwargs.get("aggregate", {}))
+    if aggregate_kwargs is False:
         raise ValueError(
-            f"""
-            `aggregate` and `plot_kwargs["aggregate"]` inconsistency detected.
-            `aggregate` = {aggregate}
-            `plot_kwargs["aggregate"]` = {plot_kwargs["aggregate"]}
-            Please make sure `aggregate` and `plot_kwargs["aggregate"]` have the same value."""
+            """plot_kwargs['aggregate'] can't be False, use aggregate=False to remove 
+            aggregate plot element"""
         )
 
     # making sure both posterior/prior predictive group and observed_data group exists in
@@ -269,7 +241,6 @@ def plot_ppc(
         rng = np.random.default_rng()
 
     # picking random pp dataset sample indexes and subsetting pp_distribution accordingly
-    # total_pp_samples = plot_collection.data.sizes["chain"] * plot_collection.data.sizes["draw"]
     total_pp_samples = np.prod(
         [pp_distribution.sizes[dim] for dim in sample_dims if dim in pp_distribution.dims]
     )
@@ -286,18 +257,19 @@ def plot_ppc(
     ):
         raise TypeError(f"`num_pp_samples` must be an integer between 1 and {total_pp_samples}.")
 
-    pp_sample_ix = rng.choice(total_pp_samples, size=num_pp_samples, replace=False)
+    # stacking sample dimensions and selecting randomly if sample_dims length>1 or
+    # num_pp_samples=total_pp_samples
+    if num_pp_samples != total_pp_samples or len(sample_dims) == 1:
+        pp_sample_ix = rng.choice(total_pp_samples, size=num_pp_samples, replace=False)
 
-    # print(f"\npp_sample_ix: {pp_sample_ix!r}")
+        # stacking sample dims into a new 'ppc_dim' dimension
+        pp_distribution = pp_distribution.stack(ppc_dim=sample_dims)
 
-    # stacking sample dims into a new 'ppc_dim' dimension
-    pp_distribution = pp_distribution.stack(ppc_dim=sample_dims)
+        # Select the desired samples
+        pp_distribution = pp_distribution.isel(ppc_dim=pp_sample_ix)
 
-    # Select the desired samples
-    pp_distribution = pp_distribution.isel(ppc_dim=pp_sample_ix)
-
-    # renaming sample_dims so that rest of plot will consider this as sample_dims
-    sample_dims = ["ppc_dim"]
+        # renaming sample_dims so that rest of plot will consider this as sample_dims
+        sample_dims = ["ppc_dim"]
 
     # wrap plot collection with pp distribution
     if plot_collection is None:
@@ -379,11 +351,11 @@ def plot_ppc(
             plot_collection=plot_collection,
             labeller=labeller,
             aes_map={
-                kind: value for key, value in aes_map.items() if key == "predictive"
+                kind: aes_map.get("predictive", {})
             },  # aes_map[kind] is set to "predictive" aes_map
             plot_kwargs=plot_kwargs_dist,  # plot_kwargs[kind] is set to "predictive" plot_kwargs
             stats_kwargs={
-                "density": value for key, value in stats_kwargs.items() if key == "predictive"
+                "density": stats_kwargs.get("predictive", {})
             },  # stats_kwargs["density"] is set to "predictive" stats_kwargs
             # via plot_dist
         )
@@ -424,11 +396,11 @@ def plot_ppc(
             plot_collection=plot_collection,
             labeller=labeller,
             aes_map={
-                kind: value for key, value in aes_map.items() if key == "aggregate"
+                kind: aes_map.get("aggregate", {})
             },  # aes_map[kind] is set to "aggregate" aes_map
             plot_kwargs=plot_kwargs_dist,  # plot_kwargs[kind] is set to "aggregate" plot_kwargs
             stats_kwargs={
-                "density": value for key, value in stats_kwargs.items() if key == "aggregate"
+                "density": stats_kwargs.get("aggregate", {})
             },  # stats_kwargs["density"] is set to "aggregate" stats_kwargs
         )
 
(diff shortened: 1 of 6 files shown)


imperorrp:
Plots generated by passing in the `centered_eight` datatree as `data`:

It seems like multiple colors are cycled through even though no aesthetic mapping was defined in kwargs for the kde curves for the ppc samples so the observed values curve isn't standing out. Not sure how to fix this-

`azp.plot_ppc(data, data_pairs={"y":"y"}, num_pp_samples=50)`
image

`azp.plot_ppc(data, data_pairs={"y":"y"}, num_pp_samples=500)`
image

When only 1 pp sample is selected though, the observed curve becomes clear (the darker one)-
`azp.plot_ppc(data, data_pairs={"y":"y"}, num_pp_samples=1)`
image

As a comparison for checking the visualization accuracy, this is the plot generated for the same `centered_eight` data by the legacy Arviz plot_ppc-
image

src/arviz_plots/plots/ppcplot.py
```diff
@@ -0,0 +1,385 @@
...
+        pp_density_dims = {dim}.union(
+            density_dims
+        )  # dims to be reduced now includes the flatten ones and not just sample dims
+    # ppc_dim should not be flattened- this is because subselections of this correspond to
+    # pp samples
+
+    pp_densities = []
+    pp_xs = []
+
+    for i in range(predictive_distribution.sizes["ppc_dim"]):
```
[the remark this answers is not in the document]
imperorrp: So this is sort the approach that should be taken when calling map on plot_ppc as well, right - as in sample_dims should be all of the non chain and draw dimensions (by default, though a user could select less) since these are the dims being flattened and used to generate KDEs, while the actual pp sample dims chain and draw are set as "overlay" in the aesthetic mapping?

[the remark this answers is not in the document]
OriolAbril: In the example `sample_dims` has nothing to do with `.map`, the need for `sample_dims="school"` is because `sample_dims` are the dimensions being reduced by `plot_dist` when computing the kde.

In plot_ppc, we split the dimensions into 3 non-overlapping groups:
* overlay dimensions -> here `sample_dims`
* facet dimensions -> what used to be `flatten` argument/now modifiable also via pc_kwargs
* rest/leftover/reduce (defined by elimination) -> these are the ones that should be reduced when computing the kde

In this particular example, they are respectively `[chain, draw]`, `[]` and `[school]`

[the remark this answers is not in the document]
imperorrp: Got it, this way of splitting makes sense. I've modified the code in my latest commit to take into account these three kinds of dimensions (for overlay/facetting/reducing)

imperorrp:
Latest commit takes into account your suggestions and comments @OriolAbril. The `mean` plotting might need some more work as the current implementation just takes into account all of the data to generate the mean. If the plot is facetted, generating the mean curves on the relevant subselections of the data would then be required I suppose.

Here is the current output when plot_ppc is called like:
```
azp.plot_ppc(data, num_pp_samples=500)
```
image

src/arviz_plots/plots/ppcplot.py
```diff
@@ -0,0 +1,479 @@
...
+            coords=coords,
+            sample_dims=reduce_dims,
+            kind=kind,
+            plot_collection=plot_collection,
+            labeller=labeller,
+            aes_map={
+                kind: value
+                for key, value in aes_map.items()
+                if key == "predictive" and key not in pp_density_ignore
+            },
```
OriolAbril, line 337: I don't understand what is happening here. The keys of the `aes_map` dict need to be ids (aka artist names) for the visual elements in plot_dist, the values are sets of aesthetic names. It indicates which aesthetics should be applied to each artist.

No key can be in `pp_density_ignore` as `pp_density_ignore` contains aesthetic names. That means only the "predictive" key will be used, which makes sense as it is what we want to plot, and it is "renamed" to `kind`, but also `kind` is hardcoded even if inside the dict comprehension, so if the loop were to actually do anything then it would break as it would attempt to generate a dictionary with repeated keys.

src/arviz_plots/plots/ppcplot.py
```diff
@@ -0,0 +1,479 @@
...
+            plot_kwargs={
+                **{key: False for key in plot_dist_artists},
+                kind: dict(pp_kwargs.items()),
+            },
+            stats_kwargs={key: value for key, value in stats_kwargs.items() if key == "predictive"},
```
OriolAbril, line 342: Somewhat similar comment here. Make sure the code makes sense and that it is consistent with the documentation.

src/arviz_plots/plots/ppcplot.py
```diff
@@ -0,0 +1,479 @@
...
+        )
+
+    # ---------STEP 2 (PPC AGGREGATE)-------------
+
+    aggregate_kwargs = copy(plot_kwargs.get("aggregate", {}))
+    if aggregate and pp_kwargs is not False and aggregate_kwargs is not False:
```
OriolAbril, line 348: This should not depend on `pp_kwargs`, only on `aggregate` related kwargs. Also, this should be consistent with whatever happens with `observed` and `plot_kwargs["observed"]` if one raises a warning when there is duplication, the other should too, if it is an error instead, both should raise an error (ideally if an error at the top of the function before any computation or plotting happens).

src/arviz_plots/plots/ppcplot.py
```diff
@@ -0,0 +1,479 @@
...
+            aggregate_kwargs.setdefault("color", aggregate_default_color)
+
+        print(f"\n aggregate reduce_dims = {reduce_dims!r}")
+        print(f"\n aggregate sample_dims = {sample_dims!r}")
+        print(f"\n aggregate_kwargs = {aggregate_kwargs}")
+        aggregate_reduce_dims = reduce_dims + list(sample_dims)
+        print(f"\n aggregate_reduce_dims = {aggregate_reduce_dims}")
+
+        # setting aggregate aes_map to `[]` so `overlay` isn't applied
+        aes_map.setdefault("aggregate", [])
```
OriolAbril, line 372: This should happen in the aes_map defaults (it doesn't really matter if we set that and then aggregate=False)

src/arviz_plots/plots/ppcplot.py
```diff
@@ -0,0 +1,479 @@
...
+        obs_distribution = process_group_variables_coords(
+            dt,
+            group=observed_data_group,
+            var_names=var_names,
+            filter_vars=filter_vars,
+            coords=coords,
+        )
+        obs_kwargs = copy(plot_kwargs.get("observed", {}))
+
+        if obs_kwargs is not False:
```
OriolAbril, line 410: This still has the double level logic with `observed` and `obs_kwargs`

src/arviz_plots/plots/ppcplot.py
```diff
@@ -0,0 +1,479 @@
...
+                Both in the the top level function arguments and in `pc_kwargs`. 
+                The `cols` key in `pc_kwargs` will take precedence.
+                                        
+                facet_dims = {facet_dims}
+                pc_kwargs["cols"] = {pc_kwargs["cols"]}
+                Duplicated dimensions: {duplicated_dims}.""",
+            )
+        # setting facet_dims to pc_kwargs defined values since it is used
+        # later to calculate dims to reduce
+        facet_dims = list(set(pc_kwargs["col"]).difference({"__variable__"}))
```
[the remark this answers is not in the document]
OriolAbril: Extra note: reviewing the plot_ridge PR, I realized that a similar thing already happens in `plot_forest` and we raise errors, so for consistency I think we should raise errors here too. ref https://github.com/arviz-devs/arviz-plots/blob/main/src/arviz_plots/plots/forestplot.py#L242-L249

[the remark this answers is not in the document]
imperorrp: Ah my bad, I just used the error bit from `plot_dist` as a template without checking it properly. I'll replace with the `raise ValueError(...)` you pointed out from `plot_forest`. 

Also, I think the first option sounds good- raise an error if both are provided and ask the user to only provide one of the args.

src/arviz_plots/plots/ppcplot.py
```diff
@@ -0,0 +1,479 @@
...
+    if "observed" in plot_kwargs:
+        if plot_kwargs["observed"] != observed:
+            with warnings.catch_warnings():
+                warnings.filterwarnings(
+                    "ignore",
+                    message="""`plot_kwargs['observed']` inconsistency detected.
+                    It is not the same as with the top level `observed` argument. 
+                    `plot_kwargs['observed']` will take precedence.""",
+                )
+            observed = plot_kwargs["observed"]
```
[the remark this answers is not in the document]
imperorrp: I'm a bit confused by this logic (Wouldn't this imply if both are true, the error is raised?) Did you mean 
```if plot_kwargs.get("observed", False) is not False and not observed:```? (implying `plot_kwargs["observed"]` is set to true but `observed` is false, so an error should be raised as they contradict)

Also what about the condition when `observed` is true but `plot_kwargs["observed"]` is false- should we display the error for this condition too due to contradiction, or just let the top-level argument `observed` take precedence? 

Because it might be the case that `observed` is not set at all by the user but defaults to true (which happens in the code before this check) because of the posterior predictive group being plotted, but the user has set `plot_kwargs["observed"]` to false to not have it be displayed- in this case should the top level argument still take preference even though it took a default value and wasn't explicitly defined in the function call?

File: src/arviz_plots/plots/ppcplot.py (hunk 2 above)
Comment (imperorrp) [reply src/arviz_plots/plots/ppcplot.py:337]: I think I meant to put `if key == "predictive" and value not in pp_density_ignore` so it checks the aesthetic names against those in `pp_density_ignore`. About the dist comprehension yeah it doesn't seem to useful in retrospect. I'll refactor this to how it's done in `plot_trace_dist` when `plot_dist` is called intrinsically

File: src/arviz_plots/plots/ppcplot.py (hunk 8 above)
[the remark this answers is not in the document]
Comment (OriolAbril) [reply src/arviz_plots/plots/ppcplot.py:189]: You are right, the code makes no sense, sorry. I started writing with an idea on mind on how to do it then changed my mind but not the text.

The two options should be considered:

```python
if (
    observed and plot_kwargs.get("observed", True) is False
    or not observed and plot_kwargs.get("observed", False) is not False
):
    # raise warning or error
```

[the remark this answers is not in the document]
OriolAbril: Whenever they are incompatible we should raise an error, thus, none will ever take precedence over the other because they either mean the same thing or an error is thrown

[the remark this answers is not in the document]
OriolAbril: Actually, depending on #66 we cpuld simplify it to " `plot_kwargs["observed"] ` can't be False, use observed top level kwarg for that"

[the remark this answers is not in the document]
imperorrp: I've used the first incompatibility logic for now- could we keep a note of this and make modifications after the proposals in #66 are finalized?

imperorrp:
Made some modifications/fixes. Some changes related to input kwargs (the incompatibility checks between top level args and `plot_kwargs` keys) might have to be further modified later though (in relation to issue #66) in the future. 

Updated the docstring as well so users are aware that the values they pass into `plot_ppc` related keys ("predictive", "aggregate", and "observed") in `plot_kwargs` and `aes_map` are transferred to `plot_dist`'s 'kind' type artists internally. In the code for this, I've used `plot_trace_dist` conventions as far as possible but with some changes since `plot_dist` is only called once in that plot and not for multiple artists.

`stats_kwargs` is passed to all internally called `plot_dist`s as is- so there is an assumption being made that users probably will want the same kind of statistical computation run for each `plot_ppc` artist they want to plot for consistency.

OriolAbril:
I'll try to review the code tomorrow, in the meantime some quick amswers to the questions above.

I don't think it matters to users that plot_dist is called internally, they only care about the valid keys and where are their values passed to eventually (so they know what values are valid). This will also match how things are documented in plot_trace_dist.

stats_kwargs should allow for different kwargs being passed to the different datasets, see https://github.com/arviz-devs/arviz-plots/pull/55#discussion_r1648995395

imperorrp:
> I'll try to review the code tomorrow, in the meantime some quick amswers to the questions above.
> 
> I don't think it matters to users that plot_dist is called internally, they only care about the valid keys and where are their values passed to eventually (so they know what values are valid). This will also match how things are documented in plot_trace_dist.
> 
> stats_kwargs should allow for different kwargs being passed to the different datasets, see [#55 (comment)](https://github.com/arviz-devs/arviz-plots/pull/55#discussion_r1648995395)

Okay, I'll adjust the docstring to reflect that so its simpler- I could make mention of it as a comment in the code though so if looking at the source code one could make out where it's being passed internally. 

Also sorry, missed that. I'll adjust the stats_kwargs usage logic to allow different ones for "predictive", "aggregate" and "observed".

imperorrp:
Added observed_rug to plot the observed values' distribution. Since no masking is required as divergences aren't being plotted here, the `trace_rug` visual element was also modified a bit to allow for mask=None. 

image

src/arviz_plots/plots/ppcplot.py
```diff
@@ -23,16 +23,14 @@ def plot_ppc(
     filter_vars=None,
     group="posterior",
     observed=None,
+    observed_rug=False,
     coords=None,
     sample_dims=None,
     kind=None,
     facet_dims=None,
-    data_pairs=None,
```
OriolAbril: data_pairs argument should be kept

src/arviz_plots/plots/ppcplot.py
```diff
@@ -0,0 +1,507 @@
...
+        if plot_collection is None:
+            backend = rcParams["plot.backend"]
+        else:
+            backend = plot_collection.backend
+    # getting plot backend to get default colors to put in default kwargs for visual elements
+    plot_bknd = import_module(f".backend.{backend}", package="arviz_plots")
+
+    # pp distribution group plotting logic
+    pp_distribution = process_group_variables_coords(
+        dt, group=predictive_data_group, var_names=var_names, filter_vars=filter_vars, coords=coords
```
OriolAbril, line 239: `data_pairs` should be taken into account here:

```suggestion
        dt, group=predictive_data_group, var_names=None if var_names is None else [data_pairs.get(var_name, var_name) for var_name in var_names], filter_vars=filter_vars, coords=coords
```

OriolAbril (commented):
This needs tests so we can check the code does what we want to happen in all the different possible scenarios

imperorrp:
About the tests- the 'kind' and 'group' arguments I've parametrized both via pytest and via hypothesis.strategies. Should only one be picked or this kept? 

The hypothesis tests are also failing a few times. At least one seems to be because of trying to toggle the 'observed' artist off with plot_kwargs but not in the top level arg, so the ValueError set for this gets raised. The others I am still unsure of why they're failing.

src/arviz_plots/plots/ppcplot.py
```diff
@@ -0,0 +1,529 @@
...
+        Whether or not to plot a rug plot of the observed data. Only valid if observed=True.
+    coords : dict, optional
+        Dictionary mapping dimensions to selected coordinates to be plotted.
+        Dimensions without a mapping specified will include all coordinates for that dimension.
+        Defaults to including all coordinates for all dimensions if None.
+    sample_dims : str or sequence of hashable, optional
+        Dimensions to loop over in plotting posterior/prior predictive.
+        Note: Dims not in sample_dims or facet_dims (below) will be reduced by default.
+        Defaults to ``rcParams["data.sample_dims"]``
+    kind : {"kde", "cumulative", "scatter"}, optional
```
OriolAbril, line 72:
```suggestion
    kind : {"kde", "cumulative", "scatter"}, optional
```

I see you did some changes like renaming "cumulative" to "ecdf" internally, but that will be confusing for users. There should be 4 possible values accepted by `kind`: kde, ecdf, hist and scatter

src/arviz_plots/plots/ppcplot.py
```diff
@@ -0,0 +1,529 @@
...
+        Defaults to empty list. A warning is raised if `pc_kwargs` is also used to define
+        dims to facet over with the `cols` key and `pc_kwargs` takes precedence.
+    data_pairs : dict, optional
+        Dictionary containing relations between observed data and posterior/prior predictive data.
+        Dictionary structure:
+            * key = observed data var_name
+            * value = posterior/prior predictive var_name
+        For example, data_pairs = {'y' : 'y_hat'}.
+        If None, it will assume that the observed data and the posterior/prior predictive data
+        have the same variable name
```
OriolAbril, line 85:
```suggestion
        Dictionary containing relations between observed data and posterior/prior predictive data.
        Dictionary keys are variable names corresponding to observed data and dictionary values
        are variable names corresponding to posterior/prior predictive. For example, data_pairs = {'y' : 'y_hat'}.
        By default, it will assume that the observed data and the posterior/prior predictive data
        have the same variable names
```

src/arviz_plots/plots/ppcplot.py
```diff
@@ -0,0 +1,529 @@
...
+    aggregate: bool, optional
+        If True, predictive data will be aggregated over both sample_dims and reduce_dims.
+        Defaults to False.
```
OriolAbril, line 88:
```suggestion
    aggregate: bool, default False
        If True, predictive data will be aggregated over both sample_dims and reduce_dims.
```

When the default is the value that is used it should be documented here directly. If it is a placeholder or if it depends on other values then use `optional` and explain the behaviour in the description

src/arviz_plots/plots/ppcplot.py
```diff
@@ -0,0 +1,529 @@
...
+    sample_dims : str or sequence of hashable, optional
+        Dimensions to loop over in plotting posterior/prior predictive.
+        Note: Dims not in sample_dims or facet_dims (below) will be reduced by default.
+        Defaults to ``rcParams["data.sample_dims"]``
+    kind : {"kde", "cumulative", "scatter"}, optional
+        How to represent the marginal density. Defaults to ``rcParams["plot.density_kind"]``.
+    facet_dims : list, optional
+        Dimensions to facet over (for which multiple plots will be generated).
+        Defaults to empty list. A warning is raised if `pc_kwargs` is also used to define
+        dims to facet over with the `cols` key and `pc_kwargs` takes precedence.
```
OriolAbril, line 77:
```suggestion
        Defaults to empty list.
```

src/arviz_plots/plots/ppcplot.py
```diff
@@ -0,0 +1,529 @@
...
+    # print(f"\npp_sample_ix: {pp_sample_ix!r}")
+
+    # stacking sample dims into a new 'ppc_dim' dimension
+    pp_distribution = pp_distribution.stack(ppc_dim=sample_dims)
+
+    # Select the desired samples
+    pp_distribution = pp_distribution.isel(ppc_dim=pp_sample_ix)
+
+    # renaming sample_dims so that rest of plot will consider this as sample_dims
+    sample_dims = ["ppc_dim"]
```
OriolAbril, line 300: I think we should skip all these steps if num_pp_samples == total_pp_samples. I think nothing within the function itself should change (test probably will) and it would allow using aesthetics for chain/draw when plotting all samples

src/arviz_plots/plots/ppcplot.py
```diff
@@ -0,0 +1,529 @@
...
+        plot_dist(
+            pp_dt,
+            group="pp_dt",
+            sample_dims=reduce_dims,
+            kind=kind,
+            plot_collection=plot_collection,
+            labeller=labeller,
+            aes_map={
+                kind: value for key, value in aes_map.items() if key == "predictive"
+            },  # aes_map[kind] is set to "predictive" aes_map
```
OriolAbril, line 383:
```suggestion
            aes_map={kind: aes_map.get("predictive", {})},
```

Not sure if you were getting errors when "predictive" key wasn't in `aes_map` but this should not be a loop, we are always selecting only one key and we know which one beforehand. Same for the other similar situations.

src/arviz_plots/visuals/__init__.py
```diff
@@ -59,9 +60,13 @@ def trace_rug(da, target, backend, mask, xname=None, y=None, **kwargs):
             xvalues = da[xname]
         if y is None:
             y = da.min().item()
+    if flatten is not False:
+        xvalues = xvalues.values.flatten()  # flatten xvalues
     if len(xvalues.shape) != 1:
         raise ValueError(f"Expected unidimensional data but got {xvalues.sizes}")
```
OriolAbril, line 66: can you try what happens when xvalues is always flattened and no error is risen?

src/arviz_plots/plots/ppcplot.py
```diff
@@ -0,0 +1,529 @@
...
+        and plot_kwargs.get("aggregate", False) is not False
+    ):
+        # raise warning or error
+        raise ValueError(
+            f"""
+            `aggregate` and `plot_kwargs["aggregate"]` inconsistency detected.
+            `aggregate` = {aggregate}
+            `plot_kwargs["aggregate"]` = {plot_kwargs["aggregate"]}
+            Please make sure `aggregate` and `plot_kwargs["aggregate"]` have the same value."""
+        )
```
OriolAbril, line 222: Seeing the tests and having reviewed plot_forest, I realized this is different behaviour from `plot_forest`. It might make sense to mimic the behaviour completely, which would be raising an error if `plot_kwargs["aggregate"]` (or observed) is `False`, continue on ahead otherwise.

This would mean we could have a dict in that key that gets ignored and the element is not plotted, but that is also what happens in plot_forest and I think is good enough behaviour.

tests/test_hypothesis_plots.py
```diff
@@ -172,3 +178,43 @@ def test_plot_ridge(datatree, combined, plot_kwargs, labels_shade_label):
...
+            "predictive": plot_kwargs_value,
+            "observed": plot_kwargs_value,
+            "aggregate": plot_kwargs_value,
+            "observed_rug": plot_kwargs_value,
+            "title": plot_kwargs_value,
+            "remove_axis": st.just(False),
+        },
+    ),
+    kind=ppc_kind_value,
+    group=ppc_group,
```
OriolAbril, line 202: All relevant plot specific arguments should be here, especially if they affect what gets plotted or how the data is processed _within the function_. Thus, `observed` and `aggregate` (toggling `observed_rug` on too if `observed` is true), `facet_dims` and `num_pp_samples` should be here.

tests/test_plots.py
```diff
@@ -246,3 +272,52 @@ def test_plot_ridge_aes_labels_shading(self, backend, datatree_4d, pseudo_dim):
         if pseudo_dim != "__variable__":
             assert all(0 in child["alpha"] for child in pc.aes.children.values())
             assert any(pseudo_dim in child["shade"].dims for child in pc.viz.children.values())
+
+    @pytest.mark.parametrize("group", ("prior", "posterior"))
```
OriolAbril, line 276: No need to parametrize by group. It plays an extremely minor role in the function and it is also being checked in the hypothesis tests.

src/arviz_plots/plots/ppcplot.py
OriolAbril: we might also want to distinguish between length 1 sample_dims (in which case it is not necessary to stack nor to rename and overwrite its value) and >1 length sample_dims (in which case stacking is necessary)

tests/test_plots.py
```diff
@@ -246,3 +272,52 @@ def test_plot_ridge_aes_labels_shading(self, backend, datatree_4d, pseudo_dim):
...
+        if kind == "kde":
+            assert "kde" in pc.viz["obs"]
+        elif kind == "cumulative":
+            assert "ecdf" in pc.viz["obs"]
+        assert "overlay" in pc.aes["obs"].data_vars
+
+    @pytest.mark.parametrize("group", ("prior", "posterior"))
+    @pytest.mark.parametrize("kind", ("kde", "cumulative"))
+    @pytest.mark.parametrize("facet_dims", (["group"], ["hierarchy"], None))
+    def test_plot_ppc_4d(self, datatree_4d, facet_dims, kind, group, backend):
```
OriolAbril, line 307: there should be assertions to check `facet_dims` is working as expected. For example, making sure that dimension is present in `pc.viz["obs"]["plot"]`

imperorrp:
- Updated plot_ppc with changes: docstring updates, a check to only flatten and stack if `num_pp_samples` isn't `total_pp_samples` or `len(sample_dims)`>1, plot_forest-like checks for `plot_kwargs['observed']` and `plot_kwargs['aggregate']` to ensure they aren't False.

- The `trace_rug` visual element now flattens by default 

- Parametrizations for `group` are removed from test_plots, parametrizations for `facet_dims` are added and assert checks for  `facet-dims` in `pc.viz["obs"]["plot"].dims` were added. Since the flattening/stacking and updating of sample_dims to ppc_dims was made conditional on some logic now, the 
`assert "ppc_dim" in pc.viz["obs"].dims`
statement doesn't work anymore- should we keep it with some logic added? 

- Also test_hypothesis_plots now has `observed`, `observed_rug`, `aggregate`, `facet_dims`, `sample_dims`, and a composite strategy for `num_pp_samples` (dependent on draws from `group` and `sample_dims`) added and `plot_kwargs["observed"]` and `plot_kwargs["aggregate"]` are always set to True. 

Modified the `datatree` fixture in test_hypothesis_plots too to name the first two dims in each variable so they can be referenced by `sample_dims` later

There are still 2 failing conditions in the hypothesis tests though -I'm not sure why yet- where the datatree seems to be interpreted as a function for some reason

OriolAbril:
>  Since the flattening/stacking and updating of sample_dims to ppc_dims was made conditional on some logic now, the
assert "ppc_dim" in pc.viz["obs"].dims
statement doesn't work anymore- should we keep it with some logic added?

Some checks on this should still be present

tests/test_hypothesis_plots.py
```diff
@@ -18,13 +18,23 @@ def datatree(seed=31):
...
             "posterior": {"mu": mu, "theta": theta, "tau": tau},
             "sample_stats": {"diverging": diverging},
+            "observed_data": {"obs": obs},
+            "prior_predictive": {"obs": prior_predictive},
+            "posterior_predictive": {"obs": posterior_predictive},
+        },
+        dims={
+            "theta": ["chain", "draw", "hierarchy", "group"],
+            "tau": ["chain", "draw", "hierarchy"],
+            "obs": ["chain", "draw", "hierarchy", "group"],
```
OriolAbril, line 36: adding chain and draw is actually breaking the tests. `from_dict` assumes the `sample_dims` rcparams prepend the provided dimensions _in some of the groups_, in others, where it doesn't make sense for these dimensions to be they are not prepended. When you say `obs: chain, draw, hierarchy, group` you are breaking this behaviour, for the posterior predictive group it is not really an issue because if they are already present arviz won't prepend them again, but in observed_data they are not present at all, yet the code now says they should

tests/test_hypothesis_plots.py
```diff
@@ -172,3 +182,86 @@ def test_plot_ridge(datatree, combined, plot_kwargs, labels_shade_label):
...
+
+@given(
+    plot_kwargs=st.fixed_dictionaries(
+        {},
+        optional={
+            "kind": plot_kwargs_value,
+            "predictive": plot_kwargs_value,
+            "observed": plot_kwargs_value,
+            "aggregate": plot_kwargs_value,
+            "observed_rug": plot_kwargs_value,
```
OriolAbril, line 217: I think none of these can be false, so they should not have their values be `plot_kwargs_value`, see what other tests like plot_forest are doing

src/arviz_plots/plots/ppcplot.py
```diff
@@ -0,0 +1,500 @@
...
+        pp_sample_ix = rng.choice(total_pp_samples, size=num_pp_samples, replace=False)
+
+        # stacking sample dims into a new 'ppc_dim' dimension
+        pp_distribution = pp_distribution.stack(ppc_dim=sample_dims)
+
+        # Select the desired samples
+        pp_distribution = pp_distribution.isel(ppc_dim=pp_sample_ix)
+
+        # renaming sample_dims so that rest of plot will consider this as sample_dims
+        sample_dims = ["ppc_dim"]
```
OriolAbril, line 272: This is not the logic we discussed.

If `num_pp_samples` is already `total_pp_samples` there is no need to subset (so no rng.choice nor .isel). Consequently, it is not necessary to stack anyhting either.

If len(sample_dims)=1, it might still be necessary to subset (this will depend on whether num_pp_samples is the same than the length of the provided dimension). But there is no need to stack anything, instead, the provided dimension should be maintained and used for subsetting.

imperorrp:
Resolved latest reviews with these last commits- the predictive values stacking/subselection logic was updated and the the tests modified. The Hypothesis tests still seem to be failing for 2 conditions though

imperorrp:
Just rebased this PR too

tests/test_hypothesis_plots.py
```diff
@@ -192,3 +202,91 @@ def test_plot_ridge(datatree, combined, plot_kwargs, labels_shade_label):
...
+    kind=ppc_kind_value,
+    group=ppc_group,
+    observed=ppc_observed,
+    observed_rug=ppc_observed,
+    aggregate=ppc_aggregate,
+    facet_dims=ppc_facet_dims,
+    sample_dims=ppc_sample_dims,
+    drawed_samples=draw_num_pp_samples(ppc_group, ppc_sample_dims),
+)
+def test_plot_ppc(
```
OriolAbril, line 251:
```suggestion
def test_plot_ppc(
    datatree,
```

imperorrp:
Added support for kind="scatter" type plots in `plot_ppc`. 

- Some work on the 'y' aesthetic that is mapped to all of the predictive rugs needs to be done
- When `observed_rug=True` and `observed=True` for kind='scatter', maybe turning one off would be good because they essentially have the same function 

```
azp.plot_ppc(
        data,
        kind="scatter",
        num_pp_samples=10,
        aggregate=True,
        observed_rug=True,
    )
```
image


```
azp.plot_ppc(
        data,
        kind="scatter",
        num_pp_samples=5,
        aggregate=True,
    )
```
image

imperorrp:
Fixed some more test related issues (new errors still persisting though). The 'alpha' keyword was missing from the 'none' backend for the 'hist' function, unlike the matplotlib and bokeh backends, which is why it hadn't been detected yet. 

Another fixed test fail was due to the fact that different artists have different dimensions (for example, the 'predictive' artists are overlayed along sample dims but the 'observed' artist does not have these dimensions and the 'aggregate' artist has these reduced instead of overlayed) so the dimensions were varying when the 'predictive' key in `plot_kwargs` was being turned off. 

Suggestion by @OriolAbril to deal with the above is to check dimensions of the artists in the tests. Current issue with this though is since a lot of the artists `plot_ppc` creates are done so via internal calls to `plot_dist`, they are named based on their 'kind' and thus can have the same name despite different plot-related visual concerns. 

Current tests bypass this if 'predictive' is False in `plot_kwargs`

#### Todo: 

- Renaming of the artists to match `plot_ppc` `plot_kwargs` keys
- Modifying the hypothesis tests to check dimensions of each of these artists 
- Fix new test errors

aloctavodia:
Closing in favor of #138

PR #57, opened 2024-06-30
Adding Ridgeplot to Arviz-Plots 

imperorrp:
First commit for adding ridge plot. As per issue #7, the already implemented forest plot structure was taken as base logic for this plot and the conventions for adding a new plot (https://arviz-plots.readthedocs.io/en/latest/contributing/new_plot.html) have been followed. This plot plots kde ridges instead of credible intervals and point estimates. 

I also added two top-level arguments to this plot since I thought they might be useful- 

- `rescale_kde` (rescales kde by raising the 'y' coord values in the density dataset to a power, set to 1 by default. Using exponent here to amplify the peaks so they're more prominently visible because there's a tendency for kdes to appear very flat when plot_ridge() has a lot of subplots stacked)

- `plot_ridge_base` (boolean to choose whether or not to plot a baseline for each ridge. I think this helps in legibility if the ridges overlap each other or have high peaks, especially if the kde was rescaled)

Also modified the `line_xy` visual element function to accept an additional `y` argument (set to None by default) which is added to `da.sel(plot_axis='y')` if it is not None, so that each ridge is plotted at the correct height corresponding to its labels on the left column. Since the functionality is the same as previously if `y` is not provided, this ensures the function works where used already (like in `plot_dist()`) without deprecation. 

----
📚 Documentation preview 📚: https://arviz-plots--57.org.readthedocs.build/en/57/

### DIFF
--- a/src/arviz_plots/plots/ridgeplot.py
+++ b/src/arviz_plots/plots/ridgeplot.py
@@ -1,0 +1,505 @@
+"""Ridge plot code."""
+
+import warnings
+from copy import copy
+from importlib import import_module
+
+import arviz_stats  # pylint: disable=unused-import
+import numpy as np
+import xarray as xr
+from arviz_base import rcParams
+from arviz_base.labels import BaseLabeller
+
+from arviz_plots.plot_collection import PlotCollection
+from arviz_plots.plots.utils import filter_aes, process_group_variables_coords
+from arviz_plots.visuals import annotate_label, fill_between_y, line_x, line_xy, remove_axis
+
+
+def plot_ridge(
+    dt,
+    var_names=None,
+    filter_vars=None,
+    group="posterior",
+    coords=None,
+    sample_dims=None,
+    rescale_kde=None,
+    plot_ridge_base=None,
+    combined=False,
+    labels=None,
+    shade_label=None,
+    plot_collection=None,
+    backend=None,
+    labeller=None,
+    aes_map=None,
+    plot_kwargs=None,
+    stats_kwargs=None,
+    pc_kwargs=None,
+):
+    """Plot 1D marginal densities in a single plot, akin to a forest plot.
+
+    Parameters
+    ----------
+    dt : DataTree or dict of {str : DataTree}
+        Input data. In case of dictionary input, the keys are taken to be model names.
+        In such cases, a dimension "model" is generated and can be used to map to aesthetics.
+
+        ``plot_ridge`` uses the dimension "column" (creating it if necessary) to generate the grid
+        then adds the intervals+point estimates to its "ridge" coordinate
+        and labels to its "labels" coordinates. The data used to plot is then the subset
+        ``column="ridge"``.
+    var_names : str or list of str, optional
+        One or more variables to be plotted.
+        Prefix the variables by ~ when you want to exclude them from the plot.
+    filter_vars : {None, “like”, “regex”}, default None
+        If None, interpret var_names as the real variables names.
+        If “like”, interpret var_names as substrings of the real variables names.
+        If “regex”, interpret var_names as regular expressions on the real variables names.
+    group : str, default "posterior"
+        Group to be plotted.
+    coords : dict, optional
+    sample_dims : str or sequence of hashable, optional
+        Dimensions to reduce unless mapped to an aesthetic.
+        Defaults to ``rcParams["data.sample_dims"]``
+    rescale_kde : float, optional
+        Rescale the ridge kde density by this factor for better legibility. Defaults to 1.
+    plot_ridge_base : bool, default False.
+        Whether to plot the base of the ridge plot or not for better legibility.
+        If None, defaults to false.
+    combined : bool, default False
+        Whether to plot intervals for each chain or not. Ignored when the "chain" dimension
+        is not present.
+    labels : sequence of str, optional
+        Sequence with the dimensions to be labelled in the plot. By default all dimensions
+        except "chain" and "model" (if present). The order of `labels` is ignored,
+        only elements being present in it matters.
+        It can include the special "__variable__" indicator, and does so by default.
+    shade_label : str, default None
+        Element of `labels` that should be used to add shading horizontal strips to the plot.
+        Note that labels and credible intervals are plotted in different :term:`plots`.
+        The shading is applied to both plots, and the spacing between them is set to 0
+        *if possible*, which is not always the case (one notable example being matplotlib's
+        constrained layout).
+    plot_collection : PlotCollection, optional
+    backend : {"matplotlib", "bokeh"}, optional
+    labeller : labeller, optional
+    aes_map : mapping of {str : sequence of str or False}, optional
+        Mapping of artists to aesthetics that should use their mapping in `plot_collection`
+        when plotted. Valid keys are the same as for `plot_kwargs` except "ticklabels"
+        which doesn't apply.
+
+        By default, aesthetic mappings are generated for: y, alpha, overlay and color
+        (if multiple models are present). All aesthetic mappings but alpha are applied
+        to both the ridge and ridge base; overlay is applied
+        to labels; and both overlay and alpha are applied to the shade.
+
+        "overlay" is a dummy aesthetic to trigger looping over variables and/or
+        dimensions using all aesthetics in every iteration. "alpha" gets two
+        values (0, 0.3) in order to trigger the alternate shading effect.
+    plot_kwargs : mapping of {str : mapping or False}, optional
+        Valid keys are:
+
+        * ridge -> passed to :func:`~visuals.line_xy`
+        * ridge_base -> passed to :func:`~.visuals.line_x`
+        * labels -> passed to :func:`~.visuals.annotate_label`
+        * shade -> passed to :func:`~.visuals.fill_between_y`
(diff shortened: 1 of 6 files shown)


imperorrp:
Some examples of the plot output with the `centered_eight` data.

## Without the extra top-level args for rescaling/adding a baseline:
`azp.plot_ridge(data)`
image

`azp.plot_ridge(data, var_names=["mu", "tau"])`
image

---------
## With rescaling:

`azp.plot_ridge(data, rescale_kde=0.5)`
image

`azp.plot_ridge(data, var_names=["mu", "tau"], rescale_kde=0.5)`
image

`azp.plot_ridge(data, rescale_kde=1.5)`
image

`azp.plot_ridge(data, var_names=["mu", "tau"], rescale_kde=1.5)`
image

-------------
## With base line added:

`azp.plot_ridge(data, plot_ridge_base=True)`
image

`azp.plot_ridge(data, var_names=["mu", "tau"], plot_ridge_base=True)`
image

(Looks a bit cluttered when baselines are added when there's a lot of ridges but for the latter I think helps in seeing the density, especially if some ridges are overlapping.)

src/arviz_plots/plots/ridgeplot.py
```diff
@@ -0,0 +1,505 @@
...
+            plot_collection.get_target(None, {"column": "labels"}),
+            **ticklabel_kwargs,
+        )
+
+    # plot kde
+    default_color = plot_bknd.get_default_aes("color", 1, {})[0]
+    # rescaling kde
+    print(f"\n kde density pre rescaled = {density!r}")
+    for var in density.data_vars:
+        density[var].loc[{"plot_axis": "y"}] = density[var].sel(plot_axis="y") ** rescale_kde
```
OriolAbril, line 467: This type of transformation should not be an option. kdes are already hard enough to interpret (even if they look simple and we believe we can read them with ease that is not the case). I would not add this as this makes it basically impossible to actually read anything from the kde other than the range and the position of the peak.

By rescaling I meant multiplying all kdes by a common scale factor in order to preserve all proportions in the curve but having the tallest one peak at height 1 (or something else, this could be user modified, similarly to the overlap control argument there is in current `plot_forest`).

```suggestion
    for var in density.data_vars:
        density.loc[{"plot_axis": "y"}] = density.sel(plot_axis="y") / density.sel(plot_axis="y").max().to_array().max() * ridge_height
```

where `ridge_height` regulates the height of the ridge (that is, the height of the tallest peak in the ridge). If `ridge_height=1` (or lower) there should be no overlap at all (when combined=True and there aren't multiple models) with higher values there will be some area overlap (>1 but smallish should mean area overlap but no edge overlap generally, >>1 will generally mean both are and edge overlap). Once both area and edge plotting are working, we should try several values like 1, 1.5, 2 and see with the example arviz datasets how each looks.

src/arviz_plots/visuals/__init__.py
```diff
@@ -13,13 +13,20 @@
...
+    dataset sliced along plot_axis='y'.
     """
     plot_backend = import_module(f"arviz_plots.backend.{backend}")
-    return plot_backend.line(da.sel(plot_axis="x"), da.sel(plot_axis="y"), target, **kwargs)
+    da_has_y = "plot_axis" in da.dims and "y" in da.plot_axis
+    if da_has_y:
+        y = da.sel(plot_axis="y") if y is None else da.sel(plot_axis="y") + y
+    if y is None:
+        raise ValueError("Unable to find values for y in `da`")
+    return plot_backend.line(da.sel(plot_axis="x"), y, target, **kwargs)
```
OriolAbril, line 29: This should use the same pattern as `annotate_label` or `point_estimate_text`. That is

1. Both x and y are arguments of the function, both default to `None`
2. the actual x and y values are obtained with `x, y = _process_da_x_y(da, x, y)` (combining values from the function arguments and the different dataarray coordinate values as needed).

OriolAbril (commented):
There should be two visual elements that generate the ridge: the edge (already implemented and working) and the face (filling the area of the kde). This will match the behaviour of current plot_ridge (via plot_forest), see for example: https://python.arviz.org/en/stable/examples/plot_forest_ridge.html.

As always either or both can be turned off. Regarding defaults, I think having the edge have the default color is great, for the area I think using the same color as the default would mean having an data-ink ratio too low, we could use `color="white", alpha=0.4` or something like that to help with visualizing overlap between ridges but keeping a "clean" look (that would look quite weird for backgrounds other than white though like `arviz-darkgrid` style), or we could use `plot_kwargs.setdefault("face", False)` to skip it by default but giving users the option to turn it on.

API question (needs discussion before any implementation): I think it would probably be nice to have a shortcut/syntactic sugar to have both edge and face use the same properties, but I am not sure how much worth it it would be.

imperorrp:
Renamed the ridge artist to edge and added the face artist. I had to modify the `fill_between_y` visual element function a bit as well so that the 'y' aesthetic argument passed to it by `pc.map` is considered and also updated the `line_xy` visual element with the pattern those other visual element functions use. 

Also moved the kde rescaling code to just after statistical computation so that a new dataset `face_density` can be created and later fed to the `fill_between_y` function. 

About the default color for face- I've just used the same one as for edge for now but with the alpha=0.4 added. Have left the bit about face/edge using the same properties for now as well, for after discussion like you said. 

Also added some tests, with style drawn from the existing ones for `plot_dist` and `plot_forest`. Tried adding some hypothesis tests but there were issues with that so I haven't committed those changes yet. 

Here's how the plots look like now: 

### When combined==True (by default)
image

image

### When combined==False

image

image

OriolAbril (commented):
> Tried adding some hypothesis tests but there were issues with that so I haven't committed those changes yet.

Try manually generating a plot_ridge without the edge and another without the face. The hypothesis test if you copied it from the plot_forest one is failing at least due to that

src/arviz_plots/visuals/__init__.py
```diff
@@ -77,8 +84,12 @@ def ecdf_line(values, target, backend, **kwargs):
     return plot_backend.line(values.sel(plot_axis="x"), values.sel(plot_axis="y"), target, **kwargs)
 
 
-def fill_between_y(da, target, backend, *, x=None, y_bottom=None, y_top=None, **kwargs):
+def fill_between_y(da, target, backend, *, x=None, y_bottom=None, y=None, y_top=None, **kwargs):
     """Fill the region between to given y values."""
+    if y_bottom is None and y:
```
OriolAbril, line 89: Side comment is that `if y_bottom is None and y` would evaluate `y` as a boolean, thus, `y_bottom=None, y=0` would not enter the if. Not relevant with the other behavioural comment though

src/arviz_plots/visuals/__init__.py
```diff
@@ -77,8 +84,12 @@ def ecdf_line(values, target, backend, **kwargs):
...
+        y_bottom = y
+    if y_top is None and y:
+        y_top = y
```
OriolAbril, line 92: I think I would modify this to make sure things behave as expected. `x` and `y` are mostly used so we can have aesthetics `x` and `y` that control shifting the visual elements within a plot. Therefore, I would expect that if I pass `y`, everything is shifted along the vertical axes. With these checks, if I give both `y_top` and `y` this won't happen.

I would use

```python
if y is not None:
    y_top += y
    y_bottom += y
```

once the `if "kwarg" in da.dims:` finishes as then both y_top and y_bottom should be array_like elements and not None.

OriolAbril (commented):
I have added a couple comments for the visuals, to make sure we can combine `x` and `y` as aesthetics along with the data provided by the plots.

imperorrp:
> > Tried adding some hypothesis tests but there were issues with that so I haven't committed those changes yet.
> 
> Try manually generating a plot_ridge without the edge and another without the face. The hypothesis test if you copied it from the plot_forest one is failing at least due to that

Yeah I just checked this- since the current code calls the density calculation function only if `edge_kwargs is not False`, it doesn't work when edge is turned off with plot_kwargs. That might be the reason behind the test failing. I could move the density computation outside to fix this, with perhaps a check for if `edge_kwargs or face_kwargs`.

imperorrp:
Modified density calculation to occur if either one of edge_kwargs or face_kwargs is not False, indicating that it is required to be calculated for plotting to be possible. Also added the visual element function changes and hypothesis tests code, which appears to pass now for plot_ridge

imperorrp:
### Current plot outputs:

#### With both edge and face:
image

#### With edge only: 
image

#### With face only: 
image

src/arviz_plots/visuals/__init__.py
```diff
@@ -13,13 +13,20 @@
...
+    If additional x and y arguments are provided, x and y are added to the values
+    in the `da` dataset sliced along plot_axis='x' and plot_axis='y'.
     """
     plot_backend = import_module(f"arviz_plots.backend.{backend}")
-    return plot_backend.line(da.sel(plot_axis="x"), da.sel(plot_axis="y"), target, **kwargs)
+    if x is not None or y is not None:  # if either x or y is provided as an arg
+        x, y = _process_da_x_y(da, x, y)
+    else:
+        x = da.sel(plot_axis="x")
+        y = da.sel(plot_axis="y")
```
OriolAbril, line 28:
```suggestion
    x, y = _process_da_x_y(da, x, y)
```

PR #58, opened 2024-07-03
Add ESS Plot

imperorrp:
Adding ESS plot (#5 ) 

Currently implemented for kind='local' only, using the new `scatter_xy` visual element function also added as part of this commit. The ess data ('y' values) obtained from the ess statistical computation via Arviz-Stats is combined with xdata ('x' values) generated via `np.linspace` (like the legacy ess plot in old Arviz does) after this xdata is broadcasted to fit the shape of the ess data. These are concatenated along a new plot_axis dimension (coords 'x' and 'y') which the `scatter_xy` visual element function then splits and plots accordingly. 

### Outputs:

`azp.plot_ess(data)`
image

`azp.plot_ess(data, var_names=["mu", "tau"])`
image

----
📚 Documentation preview 📚: https://arviz-plots--58.org.readthedocs.build/en/58/

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -474,7 +474,7 @@ max-locals=15
 max-parents=7
 
 # Maximum number of public methods for a class (see R0904).
-max-public-methods=25
+max-public-methods=20
 
 # Maximum number of return / yield for function / method body
 max-returns=6
--- a/.pylintrc
+++ b/.pylintrc
@@ -474,7 +474,7 @@ max-locals=15
 max-parents=7
 
 # Maximum number of public methods for a class (see R0904).
-max-public-methods=20
+max-public-methods=25
 
 # Maximum number of return / yield for function / method body
 max-returns=6
--- a/.pylintrc
+++ b/.pylintrc
@@ -67,6 +67,7 @@ disable=missing-docstring,
         no-member,
         import-error,
         possibly-used-before-assignment,
+        too-many-positional-arguments,
         fixme
 
 
--- a/docs/source/api/plots.rst
+++ b/docs/source/api/plots.rst
@@ -23,3 +23,4 @@ A complementary introduction and guide to ``plot_...`` functions is available at
    plot_ridge
    plot_trace
    plot_trace_dist
+   plot_ess
--- a/docs/source/api/plots.rst
+++ b/docs/source/api/plots.rst
@@ -19,8 +19,8 @@ A complementary introduction and guide to ``plot_...`` functions is available at
 
    plot_compare
    plot_dist
+   plot_ess
    plot_forest
    plot_ridge
    plot_trace
    plot_trace_dist
-   plot_ess
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ classifiers = [
 dynamic = ["version", "description"]
 dependencies = [
   "arviz-base==0.2",
-  "arviz-stats[xarray] @ git+https://github.com/arviz-devs/arviz-stats",
+  "arviz-stats[xarray] @ git+https://github.com/arviz-devs/arviz-stats@1d_ess",
 ]
 
 [tool.flit.module]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ classifiers = [
 dynamic = ["version", "description"]
 dependencies = [
   "arviz-base==0.2",
-  "arviz-stats[xarray] @ git+https://github.com/arviz-devs/arviz-stats@rankdata",
+  "arviz-stats[xarray] @ git+https://github.com/arviz-devs/arviz-stats",
 ]
 
 [tool.flit.module]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ classifiers = [
 dynamic = ["version", "description"]
 dependencies = [
   "arviz-base==0.2",
-  "arviz-stats[xarray]==0.2",
+  "arviz-stats[xarray] @ git+https://github.com/arviz-devs/arviz-stats@rankdata",
 ]
 
 [tool.flit.module]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ classifiers = [
 dynamic = ["version", "description"]
 dependencies = [
   "arviz-base==0.2",
-  "arviz-stats[xarray] @ git+https://github.com/arviz-devs/arviz-stats@1d_ess",
+  "arviz-stats[xarray] @ git+https://github.com/arviz-devs/arviz-stats",
 ]
 
 [tool.flit.module]
--- a/src/arviz_plots/plots/essplot.py
+++ b/src/arviz_plots/plots/essplot.py
@@ -11,13 +11,8 @@ from arviz_base import rcParams
 from arviz_base.labels import BaseLabeller
 
 from arviz_plots.plot_collection import PlotCollection
-from arviz_plots.plots.utils import filter_aes, process_group_variables_coords
-from arviz_plots.visuals import (  # trace_rug,; line_x,; line_fixed_x,
-    labelled_title,
-    labelled_x,
-    labelled_y,
-    scatter_xy,
-)
+from arviz_plots.plots.utils import filter_aes, get_group, process_group_variables_coords
+from arviz_plots.visuals import labelled_title, labelled_x, labelled_y, scatter_xy, trace_rug
 
 
 # function signature
@@ -32,8 +27,8 @@ def plot_ess(
     # plot specific arguments
     kind="local",
     relative=False,
-    # rug=False,
-    # rug_kind="diverging",
+    rug=False,
+    rug_kind="diverging",
     n_points=20,
     # extra_methods=False,
     # min_ess=400,
@@ -159,8 +154,12 @@ def plot_ess(
             ["__variable__"]
             + [dim for dim in distribution.dims if dim not in {"model"}.union(sample_dims)],
         )
+        pc_kwargs["aes"] = pc_kwargs.get("aes", {}).copy()
+        if "chain" in distribution:
+            pc_kwargs["aes"].setdefault("overlay", ["chain"])  # so rug for each chain is overlaid
+            # doing this^ sets overlay: chain for each artist. But we only want overlay for the
+            # rug so .difference() has to be be applied to the aes_map defaults if not wanted
         if "model" in distribution:
-            pc_kwargs["aes"] = pc_kwargs.get("aes", {}).copy()
             pc_kwargs["aes"].setdefault("color", ["model"])
             # setting x aesthetic to np.linspace(-x_diff/3, x_diff/3, length of 'model' dim)
             # x_diff = span of x axis (1) divided by number of points to be plotted (n_points)
@@ -168,10 +167,17 @@ def plot_ess(
             if "x" not in pc_kwargs:
                 pc_kwargs["x"] = np.linspace(-x_diff / 3, x_diff / 3, distribution.sizes["model"])
             pc_kwargs["aes"].setdefault("x", ["model"])
+        aux_dim_list = [dim for dim in pc_kwargs["cols"] if dim != "__variable__"]  # for divergence
         plot_collection = PlotCollection.wrap(
             distribution,
             backend=backend,
             **pc_kwargs,
+        )
+    else:
+        aux_dim_list = list(
+            set(
+                dim for child in plot_collection.viz.children.values() for dim in child["plot"].dims
+            )
         )
 
     # set plot collection dependent defaults (like aesthetics mappings for each artist)
@@ -179,11 +185,12 @@ def plot_ess(
         aes_map = {}
     else:
         aes_map = aes_map.copy()
-    aes_map.setdefault(kind, plot_collection.aes_set)
+    aes_map.setdefault(kind, plot_collection.aes_set.difference({"overlay"}))
     aes_map.setdefault("mean", plot_collection.aes_set)
     aes_map.setdefault("mean", ["linestyle", "-"])
     aes_map.setdefault("sd", plot_collection.aes_set)
     aes_map.setdefault("sd", ["linestyle", "-"])
+    aes_map.setdefault("divergence", {"overlay"})
     if labeller is None:
         labeller = BaseLabeller()
 
@@ -230,6 +237,7 @@ def plot_ess(
             ess_dataset = xr.concat([xdata_dataset, ess_y_dataset], dim="plot_axis").assign_coords(
                 plot_axis=["x", "y"]
             )
+            print(f"\n distribution = {distribution!r}")
             print(f"\n ess_dataset = {ess_dataset!r}")
 
             # step 4
@@ -241,7 +249,6 @@ def plot_ess(
                 scatter_xy, "local", data=ess_dataset, ignore_aes=ess_ignore, **ess_kwargs
             )
 
-        # WIP: repeat previous pattern for ess method kind = 'quantile'
         if kind == "quantile":
             probs = np.linspace(1 / n_points, 1 - 1 / n_points, n_points)
             xdata = probs
@@ -286,13 +293,57 @@ def plot_ess(
                 scatter_xy, "quantile", data=ess_dataset, ignore_aes=ess_ignore, **ess_kwargs
             )
 
-        # all the ess methods supported in arviz stats:
-        # valid_methods = {
-        #        "bulk", "tail", "mean", "sd", "quantile", "local", "median", "mad",
-        #        "z_scale", "folded", "identity"
-        #    }
-
     # plot rug
+    # overlaying divergences for each chain
+    if rug:
+        sample_stats = get_group(dt, "sample_stats", allow_missing=True)
+        divergence_kwargs = copy(plot_kwargs.get("divergence", {}))
+        if (
+            sample_stats is not None
+            and "diverging" in sample_stats.data_vars
+            and np.any(sample_stats[rug_kind])  # 'diverging' by default
+            and divergence_kwargs is not False
+        ):
+            divergence_mask = dt.sample_stats[rug_kind]  # 'diverging' by default
+            print(f"\n divergence_mask = {divergence_mask!r}")
+            _, div_aes, div_ignore = filter_aes(plot_collection, aes_map, "divergence", sample_dims)
+            if "color" not in div_aes:
+                divergence_kwargs.setdefault("color", "black")
+            if "marker" not in div_aes:
+                divergence_kwargs.setdefault("marker", "|")
+            # if "width" not in div_aes: # should this be hardcoded?
+            #    divergence_kwargs.setdefault("width", linewidth)
+            if "size" not in div_aes:
+                divergence_kwargs.setdefault("size", 30)
+            div_reduce_dims = [dim for dim in distribution.dims if dim not in aux_dim_list]
+
+            # xname is used to pick subset of dataset in map() to be masked
+            xname = None  # xname logic from traceplot
+            default_xname = sample_dims[0] if len(sample_dims) == 1 else "draw"
+            if (default_xname not in distribution.dims) or (
+                not np.issubdtype(distribution[default_xname].dtype, np.number)
+            ):
+                default_xname = None
+            xname = divergence_kwargs.get("xname", default_xname)
+            divergence_kwargs["xname"] = xname
+            print(f"\n div_reduce_dims = {div_reduce_dims!r}")
+            print(f"\n xname = {xname}")
+
+            draw_length = distribution.sizes["draw"]  # used to scale xvalues to between 0-1
+
+            # print(f"\n distribution = {distribution}")
+
+            plot_collection.map(
+                trace_rug,
+                "divergence",
+                data=distribution,
+                ignore_aes=div_ignore,
+                # xname=xname,
+                y=distribution.min(div_reduce_dims),
+                mask=divergence_mask,
+                scale=draw_length,
+                **divergence_kwargs,
+            )
 
     # WIP: plot mean and sd (to be done after plot_ridge PR merge for line_xy update)
     a = """xyz.
--- a/src/arviz_plots/plots/essplot.py
+++ b/src/arviz_plots/plots/essplot.py
@@ -12,7 +12,12 @@ from arviz_base.labels import BaseLabeller
 
 from arviz_plots.plot_collection import PlotCollection
 from arviz_plots.plots.utils import filter_aes, process_group_variables_coords
-from arviz_plots.visuals import labelled_title, scatter_xy
+from arviz_plots.visuals import (  # trace_rug,; line_x,; line_fixed_x,
+    labelled_title,
+    labelled_x,
+    labelled_y,
+    scatter_xy,
+)
 
 
 # function signature
@@ -97,18 +102,23 @@ def plot_ess(
     plot_kwargs : mapping of {str : mapping or False}, optional
         Valid keys are:
 
-        * One of "local", "quantile", "evolution", matching the `kind` argument.
+        * One of "local" or "quantile", matching the `kind` argument.
             * "local" -> passed to :func:`~arviz_plots.visuals.scatter_xy`
-            * "quantile" -> passed to :func:`~arviz_plots.visuals.line_xy`
-            * "evolution" -> passed to :func:`~arviz_plots.visuals.line_xy`
+            * "quantile" -> passed to :func:`~arviz_plots.visuals.scatter_xy`
 
         * divergence -> passed to :func:`~.visuals.trace_rug`
         * title -> passed to :func:`~arviz_plots.visuals.labelled_title`
+        * label -> passed to :func:`~arviz_plots.visuals.labelled_x` and
+          :func:`~arviz_plots.visuals.labelled_y`
+        * mean ->
+        * sd ->
 
     stats_kwargs : mapping, optional
         Valid keys are:
 
-        * ess -> passed to ess
+        * ess -> passed to ess, method = 'local' or 'quantile' based on `kind`
+        * mean -> passed to ess, method='mean'
+        * sd -> passed to ess, method='sd'
 
     pc_kwargs : mapping
         Passed to :class:`arviz_plots.PlotCollection.wrap`
@@ -152,6 +162,11 @@ def plot_ess(
         if "model" in distribution:
             pc_kwargs["aes"] = pc_kwargs.get("aes", {}).copy()
             pc_kwargs["aes"].setdefault("color", ["model"])
+            # setting x aesthetic to np.linspace(-x_diff/3, x_diff/3, length of 'model' dim)
+            # x_diff = span of x axis (1) divided by number of points to be plotted (n_points)
+            x_diff = 1 / n_points
+            if "x" not in pc_kwargs:
+                pc_kwargs["x"] = np.linspace(-x_diff / 3, x_diff / 3, distribution.sizes["model"])
             pc_kwargs["aes"].setdefault("x", ["model"])
         plot_collection = PlotCollection.wrap(
             distribution,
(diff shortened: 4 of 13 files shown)


src/arviz_plots/plots/essplot.py
```diff
@@ -0,0 +1,251 @@
...
+
+            # broadcasting xdata to match ess_y_dataset's shape
+            xdata_da = xr.DataArray(xdata, dims="ess_dim")
+            print(f"\n xdata_da ={xdata_da}")
+
+            # broadcasting xdata_da to match shape of each variable in ess_y_dataset
+            xdata_broadcasted_dict = {}
+            for var in ess_y_dataset.data_vars:
+                _, xdata_broadcasted = xr.broadcast(ess_y_dataset[var], xdata_da)
+                xdata_broadcasted_dict[var] = xdata_broadcasted
```
OriolAbril, line 205:
```suggestion
            xdata_broadcasted_dict = {var_name: xdata_da.broadcast_like(da) for var_name, da in ess_y_dataset.items()}
```

This should generate the same dictionary being slightly more efficient

could also be merged with the `xdata_dataset` line so it is `xdata_dataset = xr.Dataset({var_name: ...})`

src/arviz_plots/plots/essplot.py
```diff
@@ -0,0 +1,251 @@
...
+            backend = rcParams["plot.backend"]
+        pc_kwargs.setdefault("col_wrap", 5)
+        pc_kwargs.setdefault(
+            "cols",
+            ["__variable__"]
+            + [dim for dim in distribution.dims if dim not in {"model"}.union(sample_dims)],
+        )
+        if "model" in distribution:
+            pc_kwargs["aes"] = pc_kwargs.get("aes", {}).copy()
+            pc_kwargs["aes"].setdefault("color", ["model"])
```
OriolAbril, line 154:
```suggestion
        if "model" in distribution:
            pc_kwargs["aes"] = pc_kwargs.get("aes", {}).copy()
            pc_kwargs["aes"].setdefault("color", ["model"])
```

For this plot I think it would look better to also add `x` to the default aesthetic mappings when there are multiple models.

src/arviz_plots/plots/essplot.py
```diff
@@ -0,0 +1,251 @@
...
+
+            # creating a new dataset from dict of broadcasted xdata
+            xdata_dataset = xr.Dataset(xdata_broadcasted_dict)
+            print(f"\n xdata_dataset = {xdata_dataset}")
+
+            # concatenating xdata_dataset and ess_y_dataset along plot_axis
+            ess_dataset = xr.concat([xdata_dataset, ess_y_dataset], dim="plot_axis")
+
+            # assigning 'x' and 'y' coordinates to the 'plot_axis' dimension
+            ess_dataset["plot_axis"] = ["x", "y"]
```
OriolAbril, line 215:
```suggestion
            ess_dataset = xr.concat([xdata_dataset, ess_y_dataset], dim="plot_axis").assign_coords(plot_axis=["x", "y"])
```

src/arviz_plots/visuals/__init__.py
```diff
@@ -71,6 +71,15 @@ def scatter_x(da, target, backend, y=None, **kwargs):
     return plot_backend.scatter(da, y, target, **kwargs)
 
 
+def scatter_xy(da, target, backend, **kwargs):
+    """Plot a scatter plot x vs y.
+
+    The input argument `da` is split into x and y using the dimension ``plot_axis``.
+    """
+    plot_backend = import_module(f"arviz_plots.backend.{backend}")
```
OriolAbril, line 79: not yet, but after merging the plot_ridge PR, this should use `_process_da_x_y` too. See https://github.com/arviz-devs/arviz-plots/pull/57#discussion_r1664857354. The `x` as aesthetic I mentioned above won't work otherwise.

OriolAbril (commented):
Looks very good so far

imperorrp:
Fixed the docstring switch-up for type 'local' and 'quantile' and incorporated other suggestions including the x aesthetic addition for the 'model' dimension. I made a temporary addition to `scatter_xy` to add the extra x arg passed due to this new aesthetic. The output ends up looking like this though- 

```
pc = azp.plot_ess(
        {"centered": data, "non centered": data2}, var_names=["mu", "tau"]
    )
pc.add_legend("model")
```
image

The default x arg value generated for the second model is '1',  which makes the data get severely skewed out of the original x axis range. Should we hardcode some smaller 'x' values with something like `pc_kwargs['x'].setdefault(np.linspace(0, 0.01, 9))` maybe?

imperorrp:
Updated x aesthetic mapping for multiple-model cases: 

image

The logic followed is as below. The x_diff is calculated, and currently one-third of that is taken as the range within which points of different models can be plotted. `np.linspace` then ensures an even distribution of the points of whatever number of models to plot within the aforementioned range. 
```
  # setting x aesthetic to np.linspace(-x_diff/3, x_diff/3, length of 'model' dim)
  # x_diff = span of x axis (1) divided by number of points to be plotted (n_points)
  x_diff = 1 / n_points
  if "x" not in pc_kwargs:
      pc_kwargs["x"] = np.linspace(-x_diff / 3, x_diff / 3, distribution.sizes["model"])
  pc_kwargs["aes"].setdefault("x", ["model"])
```

And quantile plots added- 
image

Labelling, along the y axes ('ESS for small intervals' for kind='sample', 'ESS for quantiles' for type='quantile') and x axis ('quantile') has also been added

imperorrp:
Added rugplot to `plot_ess`. 

Like `plot_trace`, I've set the 'overlay' aesthetic for the 'chain' dimension but is ignored with `.difference()` when the default aes_map is generated for the local and quantile artists. 

Also made a modification to the `trace_rug` visual element to add a new arg `scale`, which helps in controlling the `xvalues` range. It is set to the length of the draw dimension so that the xvalues range is kept between 0-1. 

#### Plot output:

`azp.plot_ess(data, var_names=["mu", "tau"], rug=True)`
image

imperorrp:
To do: Apply `rankdata` function to the data once available in Arviz-Stats for proper rug generation.

src/arviz_plots/plots/essplot.py
```diff
@@ -0,0 +1,438 @@
...
+            pc_kwargs["aes"].setdefault("overlay", ["chain"])  # so rug for each chain is overlaid
+            # doing this^ sets overlay: chain for each artist. But we only want overlay for the
+            # rug so .difference() has to be be applied to the aes_map defaults if not wanted
+        if "model" in distribution:
+            pc_kwargs["aes"].setdefault("color", ["model"])
+            # setting x aesthetic to np.linspace(-x_diff/3, x_diff/3, length of 'model' dim)
+            # x_diff = span of x axis (1) divided by number of points to be plotted (n_points)
+            x_diff = 1 / n_points
+            if "x" not in pc_kwargs:
+                pc_kwargs["x"] = np.linspace(-x_diff / 3, x_diff / 3, distribution.sizes["model"])
```
OriolAbril, line 168:
```suggestion
            n_models = distribution.sizes["model"]
            x_diff = min(1 / n_points / 3, 1 / n_points * n_models / 10)
            pc_kwargs.setdefault("x", np.linspace(-x_diff, x_diff, n_models))
```

Seeing the plot now, 1/3 would probably look good for several models, but for only 2 of them the points end up equispaced (the model1 point corresponding to quantile 0 is as far from the model0 point corresponding to quantile 0 than from the model0 point corresponding to quantile 0.05). This change will result in smaller spacing with small number of models, but never going over 1/3 of the available space.

src/arviz_plots/plots/essplot.py
```diff
@@ -0,0 +1,438 @@
...
+            print(f"\n ess_dataset = {ess_dataset!r}")
+
+            # step 4
+            # if "color" not in ess_aes:
+            #    ess_kwargs.setdefault("color", "gray")
+
+            # step 5
+            plot_collection.map(
+                scatter_xy, "local", data=ess_dataset, ignore_aes=ess_ignore, **ess_kwargs
+            )
```
OriolAbril, line 250: all this can be removed

src/arviz_plots/plots/essplot.py
```diff
@@ -0,0 +1,438 @@
...
+            print(f"\n ess_dataset = {ess_dataset!r}")
+
+            # step 4
+            # if "color" not in ess_aes:
+            #    ess_kwargs.setdefault("color", "gray")
+
+            # step 5
+            plot_collection.map(
+                scatter_xy, "quantile", data=ess_dataset, ignore_aes=ess_ignore, **ess_kwargs
+            )
```
OriolAbril, line 294: If all this is moved outside the if. To get the same exact behaviour using `.map(scatter_xy, kind, ...)` would be needed, but from #66 I think it should instead be `.map(scatter_xy, "ess", ...)`

OriolAbril (commented):
go over plot_ess_evolution review first

imperorrp:
Just made updates. This takes into account #66 and plots the mean, sd and min_ess as well using `line_xy`. Users can also set custom text for xlabel and ylabel now:
image
image



Annotating these lines is left as well as fixing an issue when the model dimension exists, where the output is like this:
image

src/arviz_plots/plots/essplot.py
```diff
@@ -0,0 +1,432 @@
...
+            mean_ess = distribution.azstats.ess(
+                dims=mean_dims, method="mean", relative=relative, **stats_kwargs.get("mean", {})
+            )
+            print(f"\n mean_ess = {mean_ess}")
+
+            if "linestyle" not in mean_aes:
+                if backend == "matplotlib":
+                    mean_kwargs.setdefault("linestyle", "--")
+                elif backend == "bokeh":
+                    mean_kwargs.setdefault("linestyle", "dashed")
```
OriolAbril, line 309: you have to get the 2nd linestyle 

```
linestyle = plot_bknd.get_default_aes("linestyle", 2, {})[1]
```

or alternatively, the same way we have defined `|` must be a valid marker for all backends because it is part of the "common plotting interface" we could do the same for solid/dashed linestyles. For now I think I prefer the approach above though

src/arviz_plots/plots/essplot.py
```diff
@@ -0,0 +1,432 @@
...
+                min_ess_kwargs.setdefault("linestyle", "dashed")
+
+        plot_collection.map(
+            line_xy,
+            "min_ess",
+            data=min_ess_dataset,
+            ignore_aes=min_ess_ignore,
+            x=x_range,
+            **min_ess_kwargs,
+        )
```
OriolAbril, line 373: can't you use `x=x_range, y=min_ess` 1d array of length 2 for x and a scalar for y? if both x and y are provided and da doesn't have `plot_axis` (which it doesn't) then `da` is used for looping but ignored for plotting which is what we want here.

src/arviz_plots/plots/essplot.py
```diff
@@ -0,0 +1,432 @@
...
+            )
+        )
+
+    # set plot collection dependent defaults (like aesthetics mappings for each artist)
+    if aes_map is None:
+        aes_map = {}
+    else:
+        aes_map = aes_map.copy()
+    aes_map.setdefault(kind, plot_collection.aes_set.difference({"overlay"}))
+    aes_map.setdefault("rug", {"overlay"})
```
OriolAbril, line 186: sd and mean should get the `color` aesthetic mapped to them, at least when `model` dimension is present. Otherwise you get n_model values of ess_mean for each plot which can't be used to plot a horizontal line at that y coordinate.

Moreover, if color is mapped to the model, then it can be used to distinguish the mean line from the sd line, they should get different linestyles, or there should be different aesthetic mappings defined depending on whether there are multiple models or not.

src/arviz_plots/plots/essplot.py
```diff
@@ -0,0 +1,432 @@
...
+        min_ess_data = {}
+        for var in distribution.data_vars:
+            reduced_data = distribution[var].mean(
+                dim=[dim for dim in distribution[var].dims if dim in min_ess_dims]
+            )
+            min_ess_data[var] = xr.full_like(reduced_data, min_ess)
+
+        min_ess_dataset = xr.Dataset(min_ess_data)
+        print(f"\n min_ess = {min_ess_dataset}")
+
```
OriolAbril, line 359: this should get "black", "red", or "gray" as default color. Probably gray.

src/arviz_plots/plots/essplot.py
imperorrp: Linestyles -> mean, sd, min
Colors -> for diff models

src/arviz_plots/plots/essplot.py
```diff
@@ -0,0 +1,432 @@
...
+            scatter_xy, "ess", data=ess_dataset, ignore_aes=ess_ignore, **ess_kwargs
+        )
+
+    # plot rug
+    # overlaying divergences(or other 'rug_kind') for each chain
+    if rug:
+        sample_stats = get_group(dt, "sample_stats", allow_missing=True)
+        rug_kwargs = copy(plot_kwargs.get("rug", {}))
+        if rug_kwargs is False:
+            raise ValueError("plot_kwargs['rug'] can't be False, use rug=False to remove the rug")
```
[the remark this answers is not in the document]
imperorrp: Adding this `plot_forest` style check and moving it higher up for mcseplot as well

imperorrp:
Plot_ess output is now like this:

image

image

image


Each of 'mean', 'sd', and 'min_ess' has their own linestyle to distinguish between them. Maybe having another legend for them could be of help though as it might not be evident to a user which is which unless the the linestyle cycle and order of plotting is known by looking at the source code. 

The 'min_ess' line is set to gray by default and 'mean' and 'sd' to the first color in the color cycle by default. If the 'model' dim is present, then all three elements get the same color for a model.

OriolAbril:
I think annotating the lines is he only way to clearly indicate which is which. If you set `extra_methods=True` in current arviz, you'll see the output looks like this:

imatge

I would add these two annotations as visual elements here too

imperorrp:
> I think annotating the lines is he only way to clearly indicate which is which. If you set `extra_methods=True` in current arviz, you'll see the output looks like this:
> 
> imatge
> 
> I would add these two annotations as visual elements here too

Todo: Add a new visual element for these annotations

imperorrp:
Rebased essplot commits

imperorrp:
Added new `annotate_xy` visual element with in-built logic to determine whether to vertically align the intended annotating text top or bottom based on an `extra_da` arg passed to it. This helps when two lines like mean and sd both are being plotted. Defaults for them both are also set- 'bottom' for mean and 'top' for sd. 

New `plot_kwargs` keys `mean_text` and `sd_text` are also added and their default aesthetics from the `aes_map`s for `mean` and `sd` are also used for these if not set. 

#### Outputs:
image

image

imperorrp:
- Some examples to the docstring and to the example gallery for plot_ess were added. Documentation seems to build fine when tried locally. 

- Tests were added in `test_plots.py` and `test_hypothesis_plots.py`. In the former, testing with the `datatree_sample` fixture causes the ess computing function from Arviz-Stats to raise an error due to number of sample_dims not being equal to 2, which it seems to require. In the latter, a couple of errors are being raised that have to be looked into. One is with the time limit being exceeded.

imperorrp:
Fixed hypothesis test errors for ESS plot

pyproject.toml
```diff
@@ -27,7 +27,7 @@ classifiers = [
 dynamic = ["version", "description"]
 dependencies = [
   "arviz-base==0.2",
-  "arviz-stats[xarray]==0.2",
+  "arviz-stats[xarray] @ git+https://github.com/arviz-devs/arviz-stats@rankdata",
```
OriolAbril, line 30:
```suggestion
  "arviz-stats[xarray] @ git+https://github.com/arviz-devs/arviz-stats",
```

src/arviz_plots/plots/essplot.py
```diff
@@ -0,0 +1,550 @@
...
+                mean_text_kwargs.setdefault("color", "black")
+
+            mean_text_kwargs.setdefault("x", 1)
+            mean_text_kwargs.setdefault("horizontal_align", "right")
+
+            # pass the mean vertical_align data for vertical alignment setting
+            if mean_va_align is not None:
+                vertical_align = mean_va_align
+            else:
+                vertical_align = "bottom"
```
OriolAbril, line 427: also use setdefault

src/arviz_plots/plots/essplot.py
```diff
@@ -0,0 +1,550 @@
...
+                sd_text_kwargs.setdefault("color", "black")
+
+            sd_text_kwargs.setdefault("x", 1)
+            sd_text_kwargs.setdefault("horizontal_align", "right")
+
+            # pass the sd vertical_align data for vertical alignment setting
+            if sd_va_align is not None:
+                vertical_align = sd_va_align
+            else:
+                vertical_align = "top"
```
OriolAbril, line 457: also setdefault

src/arviz_plots/visuals/__init__.py
```diff
@@ -47,6 +47,7 @@ def line(da, target, backend, xname=None, **kwargs):
 
 def trace_rug(da, target, backend, mask, xname=None, y=None, **kwargs):
     """Create a rug plot with the subset of `da` indicated by `mask`."""
+    # print(f'\n da = {da}')
```
OriolAbril, line 50: delete

src/arviz_plots/visuals/__init__.py
```diff
@@ -125,6 +138,19 @@ def _ensure_scalar(*args):
     return tuple(arg.item() if hasattr(arg, "item") else arg for arg in args)
 
 
+def annotate_xy(da, target, backend, *, text, x=None, y=None, vertical_align=None, **kwargs):
+    """Annotate a point (x, y) in a plot."""
+    if vertical_align is not None:
+        # print(f"\n vertical_align.item() = {vertical_align.item()}")
```
OriolAbril, line 144: delete

src/arviz_plots/plots/essplot.py
```diff
@@ -0,0 +1,550 @@
...
+            n_models = distribution.sizes["model"]
+            x_diff = min(1 / n_points / 3, 1 / n_points * n_models / 10)
+            pc_kwargs.setdefault("x", np.linspace(-x_diff, x_diff, n_models))
+            pc_kwargs["aes"].setdefault("x", ["model"])
+        aux_dim_list = [dim for dim in pc_kwargs["cols"] if dim != "__variable__"]
+        plot_collection = PlotCollection.wrap(
+            distribution,
+            backend=backend,
+            **pc_kwargs,
+        )
```
OriolAbril, line 250: add automatic figure sizing, see https://github.com/arviz-devs/arviz-plots/blob/main/src/arviz_plots/plots/forestplot.py#L258 and https://github.com/arviz-devs/arviz-plots/blob/main/src/arviz_plots/plots/traceplot.py#L114

tests/test_hypothesis_plots.py
```diff
@@ -192,3 +192,73 @@ def test_plot_ridge(datatree, combined, plot_kwargs, labels_shade_label):
...
+
+ess_kind_value = st.sampled_from(("local", "quantile"))
+ess_relative = st.booleans()
+ess_rug = st.booleans()
+ess_extra_methods = st.booleans()
+
+
+@st.composite
+def ess_n_points(draw):
+    return draw(st.integers(min_value=1, max_value=50))  # should this range be changed?
```
OriolAbril, line 205: range 1-5

tests/test_hypothesis_plots.py
```diff
@@ -192,3 +192,73 @@ def test_plot_ridge(datatree, combined, plot_kwargs, labels_shade_label):
...
+            "title": plot_kwargs_value,
+            "remove_axis": st.just(False),
+        },
+    ),
+    kind=ess_kind_value,
+    relative=ess_relative,
+    rug=ess_rug,
+    n_points=ess_n_points(),
+    extra_methods=ess_extra_methods,
+    min_ess=ess_min_ess(),
```
OriolAbril, line 235: use `st.xyz` directly here

tests/test_hypothesis_plots.py
```diff
@@ -192,3 +192,73 @@ def test_plot_ridge(datatree, combined, plot_kwargs, labels_shade_label):
...
+def ess_min_ess(draw):
+    return draw(st.integers(min_value=10, max_value=150))  # max samples = 3 x 50 = 150
+
+
+@given(
+    plot_kwargs=st.fixed_dictionaries(
+        {},
+        optional={
+            "ess": plot_kwargs_value,
+            "rug": st.sampled_from(({}, {"color": "red"})),
```
OriolAbril, line 218: This can be defined outside in line 43 as `plot_kwargs_value_no_false`

tests/test_hypothesis_plots.py
```diff
@@ -192,3 +192,73 @@ def test_plot_ridge(datatree, combined, plot_kwargs, labels_shade_label):
                 assert all(key in child for child in pc.viz.children.values())
         elif key not in ("remove_axis", "ticklabels"):
             assert all(key in child for child in pc.viz.children.values())
+
+
+ess_kind_value = st.sampled_from(("local", "quantile"))
```
OriolAbril, line 197: also move to line 43

tests/test_plots.py
```diff
@@ -246,3 +254,51 @@ def test_plot_ridge_aes_labels_shading(self, backend, datatree_4d, pseudo_dim):
...
+        pc = plot_ess(datatree, backend=backend, rug=True)
+        assert "chart" in pc.viz.data_vars
+        assert "plot" not in pc.viz.data_vars
+        assert "ess" in pc.viz["mu"]
+        assert "min_ess" in pc.viz["mu"]
+        assert "title" in pc.viz["mu"]
+        assert "rug" in pc.viz["mu"]
+        assert "hierarchy" not in pc.viz["mu"].dims
+        assert "hierarchy" in pc.viz["theta"].dims
+        assert pc.viz["mu"].rug.shape == (4,)  # checking rug artist shape (4 chains overlaid)
```
OriolAbril, line 268:
```suggestion
        assert "chain" in pc.viz["mu"].rug.dims
```

tests/test_plots.py
```diff
@@ -246,3 +254,51 @@ def test_plot_ridge_aes_labels_shading(self, backend, datatree_4d, pseudo_dim):
...
+        print("a")  # just to temporarily suspend linting error
+
+    def test_plot_ess_models(self, datatree, datatree2, backend):
+        pc = plot_ess({"c": datatree, "n": datatree2}, backend=backend, rug=False)
+        assert "chart" in pc.viz.data_vars
+        assert "plot" not in pc.viz.data_vars
+        assert "ess" in pc.viz["mu"]
+        assert "min_ess" in pc.viz["mu"]
+        assert "title" in pc.viz["mu"]
+        # assert "rug" in pc.viz["mu"]
```
OriolAbril, line 295:
```suggestion
        assert "rug" not in pc.viz["mu"]
```

tests/test_plots.py
```diff
@@ -246,3 +254,51 @@ def test_plot_ridge_aes_labels_shading(self, backend, datatree_4d, pseudo_dim):
...
+        assert "hierarchy" not in pc.viz["mu"].dims
+        assert "hierarchy" in pc.viz["theta"].dims
+        assert "model" in pc.viz["mu"].dims
+        # assert pc.viz["mu"].rug.shape == (2, 4)  # since there are 2 Ms, 4 Cs in datatree
```
OriolAbril, line 299: delete

OriolAbril (commented):
very close to merging

imperorrp:
Rebased the PR

docs/source/api/plots.rst
```diff
@@ -23,3 +23,4 @@ A complementary introduction and guide to ``plot_...`` functions is available at
    plot_ridge
    plot_trace
    plot_trace_dist
+   plot_ess
```
OriolAbril, line 26: can you keep the entries sorted alphabetically?

src/arviz_plots/plots/essplot.py
```diff
@@ -0,0 +1,585 @@
+"""ess plot code."""
+
+# imports
+# import warnings
```
OriolAbril, line 4:
```suggestion
```

src/arviz_plots/plots/essplot.py
```diff
@@ -0,0 +1,585 @@
...
+    )
+
+    # ensuring plot_kwargs['rug'] is not False
+    rug_kwargs = copy(plot_kwargs.get("rug", {}))
+    if rug_kwargs is False:
+        raise ValueError("plot_kwargs['rug'] can't be False, use rug=False to remove the rug")
+
+    if backend is None:
+        backend = rcParams["plot.backend"]
+    plot_bknd = import_module(f".backend.{backend}", package="arviz_plots")
```
OriolAbril, line 228: Use same pattern as plot_forest: https://github.com/arviz-devs/arviz-plots/blob/main/src/arviz_plots/plots/forestplot.py#L217-L221

src/arviz_plots/plots/essplot.py
```diff
@@ -0,0 +1,585 @@
...
+        n_cols = leaf_dataset(plot_collection.viz, "col").max().to_array().max().item()
+
+    plot_bknd = import_module(f".backend.{backend}", package="arviz_plots")
+
+    figsize = plot_bknd.scale_fig_size(
+        figsize,
+        rows=n_rows,
+        cols=n_cols,
+        figsize_units=figsize_units,
+    )
```
OriolAbril, line 262: This is the initial version of figsize scaling, it should follow the same pattern as plot_forest. The only thing to take from plot_trace is the computation of number of rows and columns (instead of using the one from plot_forest which would make no sense here).

https://github.com/arviz-devs/arviz-plots/blob/main/src/arviz_plots/plots/forestplot.py#L258-L278, changing https://github.com/arviz-devs/arviz-plots/blob/main/src/arviz_plots/plots/forestplot.py#L271-L272 to the computation plot_trace does to get the number of columns and rows in the generated grid.

imperorrp:
Made those modifications. Also, pylint was indicating an error regarding the number of tests under the TestPlots class in testplots.py so I raised the limit from 20 to 25. 

The output is like this right now though: 
image

^All 10 subplots in one row despite the print statements indicating the figsizing logic computes 2 rows and sticks to the 5 max subplots per row constraint. This has to be looked into if I'm missing something.

.pylintrc
```diff
@@ -474,7 +474,7 @@ max-locals=15
 max-parents=7
 
 # Maximum number of public methods for a class (see R0904).
-max-public-methods=20
+max-public-methods=25
```
OriolAbril, line 477: Add a comment in the `test_plots` file instead. https://github.com/arviz-devs/arviz-plots/blob/main/tests/test_plots.py#L1

src/arviz_plots/plots/essplot.py
```diff
@@ -0,0 +1,589 @@
...
+                figsize,
+                rows=n_rows,
+                cols=n_cols,
+                figsize_units=figsize_units,
+            )
+            print(f"\n figsize = {figsize!r}")
+            figsize_units = "dots"
+        pc_kwargs["plot_grid_kws"]["figsize"] = figsize
+        pc_kwargs["plot_grid_kws"]["figsize_units"] = figsize_units
+        plot_collection = PlotCollection.grid(
```
OriolAbril, line 286: You are using `.grid` instead of `.wrap` so the grid no longer "wraps" the 1sr row into multiple ones when it has too many columns.

src/arviz_plots/plots/essplot.py
```diff
@@ -0,0 +1,589 @@
...
+            pc_kwargs["aes"].setdefault("color", ["model"])
+            n_models = distribution.sizes["model"]
+            x_diff = min(1 / n_points / 3, 1 / n_points * n_models / 10)
+            pc_kwargs.setdefault("x", np.linspace(-x_diff, x_diff, n_models))
+            pc_kwargs["aes"].setdefault("x", ["model"])
+        aux_dim_list = [dim for dim in pc_kwargs["cols"] if dim != "__variable__"]
+        figsize = pc_kwargs.get("plot_grid_kws", {}).get("figsize", None)
+        figsize_units = pc_kwargs.get("plot_grid_kws", {}).get("figsize_units", "inches")
+        if figsize is None:
+            coeff = 0.2
```
OriolAbril, line 261: All the coeff related part was plot_forest specific, here the variable isn't used anywhere

src/arviz_plots/plots/essplot.py
imperorrp: Yeah using `.wrap` worked: 

image

imperorrp:
Todo: Add `plot_ess_evolution`, `plot_mcse` reference in 'see also' once both this and their PRs (#71, #79) are merged.

Issue #56, opened 2024-06-21
Set default color in plot_dist


OriolAbril:
Right after https://github.com/arviz-devs/arviz-plots/blob/6f8b7507888cacfc6fcab488411d179da0b853ab/src/arviz_plots/plots/distplot.py#L229 we should check if color has an aesthetic mapped to it and if not srt a default color. Otherwise due to matplotlib default property cycle we get lines with different colors without any aesthetic mapping.

We already do something similar for plot_forest: https://github.com/arviz-devs/arviz-plots/blob/main/src/arviz_plots/plots/forestplot.py#L489-L493

PR #61, opened 2024-07-03
plotly backend proof of concept

OriolAbril:


OriolAbril:
Ready to merge! 🎉  

I am quite happy with the example gallery. It has a lot of common elements with the original one: cool css thanks to Sarina, tabs for different backends, link to api docs and to download the script. And in addition it has several improvements:

* Thanks to the refactoring the script is now shared, the only thing that changes is the `backend="none"` part. The line with that content has a comment to indicate so, and is also highlighted so it gets extra emphasis.
* Consequently, only the plots themselves are inside the tab, and tab management has gotten much better, both due to new features in sphinx-design extension and due to better use of them on our side. Now tabs should sync across pages. That is, if I select the plotly backend on the plot_forest page and them move to the plot_dist one, the active tab should continue to be the plotly one. Moreover, it is possible to share links that will open the page with a specific tab as the active one using `?backend=plotly#synchronised-tabs` suffix for the html (these html are also present in the bokeh and plotly tabs so users are aware of the possibility).
* Everything continues to be in a single python file with a docstring at the bottom, but processing of the content has changed and is now a bit simpler. First the content of the python file is splitted between code and docstring, then the docstring is splitted in two using `---` as separator. When putting the page together, the content above the `---` goes at the top, then comes the tabs and code snippet (properly processed automatically to generate all backends and such), then the content below the `---` and eventually the download button at the bottom. Other than the `---` split, the content of the docstring is left as is and should be valid myst markdown.

Regarding plotly, all currenly implemented plots already work, and I have also added initial support for the `arviz-clean` style. Now `az.style.use` in no longer an alias to matplotlib `plt.style.use` but instead sets that style as the default for all installed backends with style/template support. This still needs to be documented though.

aloctavodia:
This is really awesome, so I am merging. I have a few nitpicks, that I am listing here. I can open issues later.

Marginal distribution comparison plot --- point-estimate overlaps (plotly okish, but not matplotlib), titles overlaps (plotly) 
ECDF plot --- plotly each variable has a different color, but the same for matplotlib and bokeh
Trace and distribution plot -- plotly looks very tall compared to matplotlib
Trace plot -- plotly titles overlap
My first impression is that we should extend style.py so we have some of the functionalities available in matplotlib, at least being able to list the available styles.

OriolAbril:
Issues will be great. Some of them will be fixed by #67 I think (as once implemented all backends should get roughly the same size, and definitely the same aspect ratio).

The title overlap is most probably bad plotly usage on my side 😅, and the different colors already have an open issue https://github.com/arviz-devs/arviz-plots/issues/56 as this is also an issue in matplotlib when we use plot dist to overlay things on the same plot (automatic cycling works differently in the backends: matplotlib cycles automatically within axes, bokeh does no automatic cycling and plotly cycles automatically within figures)

Issue #5, opened 2023-10-05
Add plot_ess


aloctavodia:

OriolAbril:
`ess` is already available in `arviz-stats` it is possible to implement this plot already

OriolAbril:
Should we split that into `plot_ess` (local or quantile) and `plot_ess_evolution`? The first two are quite similar whereas the 3rd one has nothing to do (in terms of plotting).

aloctavodia:
Yes, that makes sense.

Issue #66, opened 2024-07-06
Define and document conventions on artist naming and keys for xyz_kwargs in plots module


OriolAbril:
Functions in `plots` module have `aes_kwargs`, `plot_kwargs`, `stats_kwargs` whose keys we need to define. We should define how we want those to behave and document whatever naming conventions we are using to then be able to ensure consistency across plots. Otherwise we'll end up having too many plots and it will no longer be possible to use something intuitive-ish and consistent.

Things I think we have been using quite consistently:

* Full names are preferred over abbreviations or acronyms
* Each visual element has a name which is a valid key in `plot_kwargs`. This allows users to deactivate/skip any operation the plot does on.
* Keys in `aes_kwargs` are keys in `plot_kwargs`. Some might only make sense for `plot_kwargs` and can be skipped (e.g. "remove_axis") and there can be some exceptions for different visual elements that share the same aesthetic mappings, but this should be done sparsely and only when the different visual elements end up calling the same backend function.
* There aren't really rules or conventions on `stats_kwargs` so far. Also related to #40

Open questions:

* what should happen if there are invalid keys?

  I lean towards raising an error.

* what should happen if there are valid keys but that will be ignored?

  I think here the best is probably to ignore them and go ahead, at most print a warning. This would allow for example: `plot_dist(..., stats_kwargs={"kde": {"bw": 4}, "hist": {"bins": 20}})` to work and correctly process the kwargs with both `kind="kde"` and `kind="hist"` making the switch between different visualizations as painless as possible (as opposed to changing kind and modifying stats_kwargs).

* Should we try to match the keys in `stats_kwargs` to the key in `plot_kwargs` where that data is more prominently encoded?

OriolAbril:
Should there be some kind of syntactic sugar to set multiple keys to the same value?

For example, to disable the ci, point estimate and title in plot_dist, something like:

```python
plot_kwargs = {key: False for key in ("credible_interval", "title", "point_estimate")}
```

One option could be to allow "," or "+" in the keys, so we then internally "explode" those composed keys into the multiple ones. Then, setting ci and title to color red and disabling remove axis and point estimate would be something like:

```python
plot_kwargs = {"title,credible_interval": {"color": "red"}, "point_estimate,remove_axis": False}
```

OriolAbril:
[...]

PR #71, opened 2024-07-16
Adding Plot ESS Evolution

imperorrp:
First commit for `plot_ess_evolution()`, as a branched off plot from legacy `plot_ess` for kind = 'evolution' (https://github.com/arviz-devs/arviz-plots/issues/5). The `line_xy` and `scatter_xy` visual elements were used for the plotting, with rug plot, axes labelling and title labelling functionality taken from ESS plot functionality. Since the ESS plot PR hasn't been merged yet, the `scatter_xy` visual element was duplicated for use here. 

#### WIP: 
Marking the 'bulk' and 'tail' artists with a legend 

#### Current plot outputs:

`azp.plot_ess_evolution(data)`
image

`azp.plot_ess_evolution(data, var_names=["mu", "tau"])`
image

----
📚 Documentation preview 📚: https://arviz-plots--71.org.readthedocs.build/en/71/

### DIFF
--- a/src/arviz_plots/plots/evolutionplot.py
+++ b/src/arviz_plots/plots/evolutionplot.py
@@ -1,0 +1,640 @@
+"""evolution ess plot."""
+
+# imports
+# import warnings
+from copy import copy
+from importlib import import_module
+
+import arviz_stats  # pylint: disable=unused-import
+import numpy as np
+import xarray as xr
+from arviz_base import rcParams
+from arviz_base.labels import BaseLabeller
+
+from arviz_plots.plot_collection import PlotCollection, process_facet_dims
+from arviz_plots.plots.utils import filter_aes, process_group_variables_coords
+from arviz_plots.visuals import (
+    annotate_xy,
+    labelled_title,
+    labelled_x,
+    labelled_y,
+    line_xy,
+    scatter_xy,
+)
+
+
+def plot_ess_evolution(
+    dt,
+    var_names=None,
+    filter_vars=None,
+    group="posterior",
+    coords=None,
+    sample_dims=None,
+    relative=False,
+    n_points=20,
+    extra_methods=False,
+    min_ess=400,
+    plot_collection=None,
+    backend=None,
+    labeller=None,
+    aes_map=None,
+    plot_kwargs=None,
+    stats_kwargs=None,
+    pc_kwargs=None,
+):
+    """Plot estimated effective sample size plots for increasing number of iterations.
+
+    Parameters
+    ----------
+    dt : DataTree or dict of {str : DataTree}
+        Input data. In case of dictionary input, the keys are taken to be model names.
+        In such cases, a dimension "model" is generated and can be used to map to aesthetics.
+    var_names : str or sequence of str, optional
+        One or more variables to be plotted.
+        Prefix the variables by ~ when you want to exclude them from the plot.
+    filter_vars : {None, “like”, “regex”}, default None
+        If None, interpret `var_names` as the real variables names.
+        If “like”, interpret `var_names` as substrings of the real variables names.
+        If “regex”, interpret `var_names` as regular expressions on the real variables names.
+    group : str, default "posterior"
+        Group to be plotted.
+    coords : dict, optional
+    sample_dims : str or sequence of hashable, optional
+        Dimensions to reduce unless mapped to an aesthetic.
+        Defaults to ``rcParams["data.sample_dims"]``
+    relative : bool, default False
+        Show relative ess in plot ``ress = ess / N``.
+    n_points : int, default 20
+        Number of subsets in the evolution plot.
+    extra_methods : bool, default False
+        Plot mean and sd ESS as horizontal lines.
+    min_ess : int, default 400
+        Minimum number of ESS desired. If ``relative=True`` the line is plotted at
+        ``min_ess / n_samples`` as a curve following the ``min_ess / n`` dependency
+    plot_collection : PlotCollection, optional
+    backend : {"matplotlib", "bokeh"}, optional
+    labeller : labeller, optional
+    aes_map : mapping of {str : sequence of str or False}, optional
+        Mapping of artists to aesthetics that should use their mapping in `plot_collection`
+        when plotted. Valid keys are the same as for `plot_kwargs`.
+
+    plot_kwargs : mapping of {str : mapping or False}, optional
+        Valid keys are:
+        * ess_bulk -> passed to :func:`~arviz_plots.visuals.scatter_xy`
+        * ess_bulk_line -> passed to :func:`~arviz_plots.visuals.line_xy`
+        * ess_tail -> passed to :func:`~arviz_plots.visuals.scatter_xy`
+        * ess_tail_line -> passed to :func:`~arviz_plots.visuals.line_xy`
+        * title -> passed to :func:`~arviz_plots.visuals.labelled_title`
+        * xlabel -> passed to :func:`~arviz_plots.visuals.labelled_x`
+        * ylabel -> passed to :func:`~arviz_plots.visuals.labelled_y`
+        * mean -> passed to :func:`~arviz_plots.visuals.line_xy`
+        * sd -> passed to :func:`~arviz_plots.visuals.line_xy`
+        * mean_text -> passed to :func:`~arviz.plots.visuals.annotate_xy`
+        * sd_text -> passed to :func:`~arviz.plots.visuals.annotate_xy`
+        * min_ess -> passed to :func:`~arviz_plots.visuals.line_xy`
+
+    stats_kwargs : mapping, optional
+        Valid keys are:
+        * ess_bulk -> passed to ess, method = 'bulk'
+        * ess_tail -> passed to ess, method = 'tail'
+        * mean -> passed to ess, method='mean'
+        * sd -> passed to ess, method='sd'
+
+    pc_kwargs : mapping
+        Passed to :class:`arviz_plots.PlotCollection.wrap`
+
+    Returns
+    -------
+    PlotCollection
+
+    See Also
+    --------
+    :ref:`plots_intro` :
+        General introduction to batteries-included plotting functions, common use and logic overview
+
+    Examples
+    --------
+    When adding a mapping for color across variables, the same color for a variable gets
+    applied to both the 'bulk' and 'tail' ess. In such a case, if separate linestyles for
+    'bulk' and 'tail' are desired to distinguish them instead of colors (which is what is
+    used by default), then this can be implemented:
+
+    .. plot::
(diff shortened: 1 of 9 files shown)


src/arviz_plots/plots/evolutionplot.py
```diff
@@ -0,0 +1,449 @@
...
+    if aes_map is None:
+        aes_map = {}
+    else:
+        aes_map = aes_map.copy()
+    aes_map.setdefault("bulk", plot_collection.aes_set.difference({"overlay"}))
+    aes_map.setdefault("tail", plot_collection.aes_set.difference({"overlay"}))
+    aes_map.setdefault("mean", plot_collection.aes_set)
+    aes_map.setdefault("mean", ["linestyle", "-"])
+    aes_map.setdefault("sd", plot_collection.aes_set)
+    aes_map.setdefault("sd", ["linestyle", "-"])
```
OriolAbril, line 182: I think by default we don't want any aesthetic mapping to be applied to either mean nor sd, so I would remove these 4 lines. Also note the lines 2 and 4 are not doing anything because the lines right before already define that key, but if they did, it would be incorrect. Values in the `aes_map` dict should be sets of strings that represent aesthetics.

This seems a confusion with setting a default linestyle in `plot_kwargs`. If you want an aesthetic property to always have the same value, use `plot_kwargs` if instead you want it to depend on a dimension/variable use `aes` in `pc_kwargs` to define the mapping and `aes_map` to indicate to which visual elements should the mapping be applied too.

There is nothing about this concepts in https://arviz-plots.readthedocs.io/en/latest/contributing/new_plot.html because that doc is for writing plots and these are usage questions which are therefore out of scope. The usage docs are instead https://arviz-plots.readthedocs.io/en/latest/tutorials/plots_intro.html and to a much lesser extent https://arviz-plots.readthedocs.io/en/latest/tutorials/intro_to_plotcollection.html

src/arviz_plots/plots/evolutionplot.py
```diff
@@ -0,0 +1,449 @@
...
+    aes_map : mapping of {str : sequence of str or False}, optional
+        Mapping of artists to aesthetics that should use their mapping in `plot_collection`
+        when plotted. Valid keys are the same as for `plot_kwargs`.
+    plot_kwargs : mapping of {str : mapping or False}, optional
+        Valid keys are:
+        * "bulk" -> passed to :func:`~arviz_plots.visuals.scatter_xy` and
+        :func:`~arviz_plots.visuals.line_xy`
+        * "tail" -> passed to :func:`~arviz_plots.visuals.scatter_xy` and
+        :func:`~arviz_plots.visuals.line_xy`
+        * divergence -> passed to :func:`~.visuals.trace_rug`
```
OriolAbril, line 96: This key should be "rug" instead. The most common thing to use (and therefore default for `rug_kind` are divergences) but they are not the only option. As we agreed on smaller set of fixed keys in #66, I think this key should be "rug" instead of `rug_kind`.

Note this will require multiple changes in the function code

src/arviz_plots/plots/evolutionplot.py
```diff
@@ -0,0 +1,449 @@
...
+    if labeller is None:
+        labeller = BaseLabeller()
+
+    # compute and add ess evolution subplots for 'bulk' and 'tail'
+
+    n_draws = distribution.sizes["draw"]
+    n_samples = n_draws * distribution.sizes["chain"]
+    plot_bknd = import_module(f".backend.{backend}", package="arviz_plots")
+    default_bulk_color = plot_bknd.get_default_aes("color", 2, {})[0]
+    default_tail_color = plot_bknd.get_default_aes("color", 2, {})[1]
```
OriolAbril, line 193:
```suggestion
    default_bulk_color, default_tail_color = plot_bknd.get_default_aes("color", 2, {})
```

src/arviz_plots/plots/evolutionplot.py
```diff
@@ -0,0 +1,449 @@
...
+    aes_map.setdefault("sd", plot_collection.aes_set)
+    aes_map.setdefault("sd", ["linestyle", "-"])
+    aes_map.setdefault("divergence", {"overlay"})
+    if labeller is None:
+        labeller = BaseLabeller()
+
+    # compute and add ess evolution subplots for 'bulk' and 'tail'
+
+    n_draws = distribution.sizes["draw"]
+    n_samples = n_draws * distribution.sizes["chain"]
```
OriolAbril, line 190: This won't work when `sample_dims` are not `chain` and `draw`, it should work at least with a single `sample` dimension without users needing to set any extra parameter manually (more complex cases should also be possible but might require users to provide specific `pc_kwargs` or things like that).

src/arviz_plots/plots/evolutionplot.py
```diff
@@ -0,0 +1,449 @@
...
+                    **stats_kwargs.get("ess_bulk", {}),
+                )
+                for draw_div in draw_divisions
+            ],
+            dim="ess_dim",
+        )
+
+        # ----------------------------
+
+        xdata = np.linspace(0, n_samples, n_points)
```
OriolAbril, line 219: This is overwriting the xdata above but I think we want the one above, so I'd remove this line. Otherwise we get ess not necessarly 0 at 0 samples which makes no sense

src/arviz_plots/plots/evolutionplot.py
```diff
@@ -0,0 +1,449 @@
...
+        if "color" not in bulk_aes:
+            bulk_kwargs.setdefault("color", default_bulk_color)
+
+        # step 5
+        plot_collection.map(
+            scatter_xy, "bulk", data=ess_dataset, ignore_aes=bulk_ignore, **bulk_kwargs
+        )
+        plot_collection.map(
+            line_xy, "bulk", data=ess_dataset, ignore_aes=bulk_ignore, **bulk_kwargs
+        )
```
OriolAbril, line 251: These are two different visual elements which should have different names and consequently different kwargs too. It should be possible to remove one but not the other, and it should be possible to change the marker of the scatter for example, but if we use the same kwargs, these are then passed to the line plotting function which doesn't accept the marker kwarg.

src/arviz_plots/plots/evolutionplot.py
```diff
@@ -0,0 +1,449 @@
...
+            pc_kwargs["aes"].setdefault("overlay", ["chain"])  # so rug for each chain is overlaid
+            # doing this^ sets overlay: chain for each artist. But we only want overlay for the
+            # rug so .difference() has to be be applied to the aes_map defaults if not wanted
+        if "model" in distribution:
+            pc_kwargs["aes"].setdefault("color", ["model"])
+            # setting x aesthetic to np.linspace(-x_diff/3, x_diff/3, length of 'model' dim)
+            # x_diff = span of x axis (1) divided by number of points to be plotted (n_points)
+            x_diff = 1 / n_points
+            if "x" not in pc_kwargs:
+                pc_kwargs["x"] = np.linspace(-x_diff / 3, x_diff / 3, distribution.sizes["model"])
```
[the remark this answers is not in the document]
OriolAbril: I thought I was reviewing the plot_ess PR. Having x with an aesthetic mapping here makes no sense and should be completely removed.

src/arviz_plots/plots/evolutionplot.py
```diff
@@ -0,0 +1,449 @@
...
+        ):
+            divergence_mask = dt.sample_stats[rug_kind]  # 'diverging' by default
+            print(f"\n divergence_mask = {divergence_mask!r}")
+            _, div_aes, div_ignore = filter_aes(plot_collection, aes_map, "divergence", sample_dims)
+            if "color" not in div_aes:
+                divergence_kwargs.setdefault("color", "black")
+            if "marker" not in div_aes:
+                divergence_kwargs.setdefault("marker", "|")
+            # if "width" not in div_aes: # should this be hardcoded?
+            #    divergence_kwargs.setdefault("width", linewidth)
```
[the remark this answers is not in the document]
imperorrp: I tried this with bokeh, seems like there definitely is some default that isn't zero when nothing is used- 

image

src/arviz_plots/plots/evolutionplot.py
```diff
@@ -0,0 +1,449 @@
...
+        label_kwargs.setdefault("color", "black")
+
+    # label_kwargs.setdefault("size", textsize)
+
+    # formatting ylabel and setting xlabel
+    if relative is not False:
+        ylabel = ylabel.format("Relative ESS")
+    else:
+        ylabel = ylabel.format("ESS")
+    xlabel = "Total Number of Draws"
```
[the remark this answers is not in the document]
imperorrp: So "xlabel", "ylabel" as two separate keys, right?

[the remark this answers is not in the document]
OriolAbril: yes

src/arviz_plots/plots/evolutionplot.py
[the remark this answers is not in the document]
OriolAbril: then no need to set it, you can delete these lines

imperorrp:
Updated plot_ess_evolution. The points and lines for both 'bulk' and 'tail' methods can now be turned on and off via `plot_kwargs`, the mean, sd, and min_ess were added following the current plot_ess logic, and xlabel and ylabel texts can also now be set separately. 

Also added a common ess_dataset computing function since it can be called upto 4 times (one for each of the cartesian product of 'bulk', 'tail' and points, lines). 

The rug plot element is still WIP.

The outputs without the rug are now like: 
image

image

image

#### An aesthetics issue with the color coding of the mean/sd plot elements: 

For the mean/sd plotting (and this'd similarly apply to plot_ess too) it looks like setting mean and sd to one color by default as I'm doing currently makes it hard to interpret that they have been overlapped when they coincide: 
image

When they're not set to the same color, it's visible that they coincide and are overlapped but then the colors might be confusing as they correspond to the 'bulk' and 'tail' plotting elements' colors too:
image

OriolAbril:
I think annotating the lines is he only way to clearly indicate which is which. If you set `extra_methods=True` in current arviz, you'll see the output looks like this:

imatge

I would add these two annotations as visual elements here too

imperorrp:
Rebased plot ess evolution commits

imperorrp:
The mean and sd lines are now annotated:

image

Should the min_ess line also be annotated? (This also applies for ESS plot #58 ). Current Arviz ESS plot doesn't have it annotated (https://github.com/arviz-devs/arviz/blob/main/arviz/plots/backends/matplotlib/essplot.py#L150) but it would probably be nice to have and is easy to add.

imperorrp:
Fixed hypothesis tests errors and documentation build

OriolAbril (approved):

PR #75, opened 2024-07-30
expand style functionality

aloctavodia:
This adds two new functions. 

* style.available(), list all available styles/templates per backed
* style.get(), return the style/template by name

----
📚 Documentation preview 📚: https://arviz-plots--75.org.readthedocs.build/en/75/

### DIFF
--- a/src/arviz_plots/style.py
+++ b/src/arviz_plots/style.py
@@ -73,7 +73,7 @@ def get(name, backend=None):
     """
     if backend is None:
         backend = rcParams["plot.backend"]
-    elif backend not in ["matplotlib", "plotly"]:
+    if backend not in ["matplotlib", "plotly"]:
         raise ValueError(f"Default styles/templates are not supported for Backend {backend}")
 
     if backend == "matplotlib":
@@ -88,4 +88,4 @@ def get(name, backend=None):
         if name in pio.templates:
             return pio.templates[name]
 
-    return ValueError(f"Style {name} not found.")
+    raise ValueError(f"Style {name} not found.")
--- a/src/arviz_plots/style.py
+++ b/src/arviz_plots/style.py
@@ -1,5 +1,7 @@
 """Style/templating helpers."""
 
+from arviz_base import rcParams
+
 
 def use(name):
     """Set an arviz style as the default style/template for all available backends.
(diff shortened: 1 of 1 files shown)


src/arviz_plots/style.py
```diff
@@ -13,16 +13,79 @@ def use(name):
...
+            pio.templates.default = name
+            ok = True
+    except ImportError:
+        pass
+
+    if not ok:
+        raise ValueError(f"Style {name} not found.")
+
+
+def available():
```
OriolAbril, line 40: How about returning a dictionary {backend: list of style names} instead of printing? It could be printed later on but would allow checking if a style is available programatically

src/arviz_plots/style.py
```diff
@@ -13,16 +13,79 @@ def use(name):
...
+    try:
+        import plotly.io as pio
+
+        print("Plotly templates:")
+        print(list(pio.templates))
+    except ImportError:
+        pass
+
+
+def get(name, backend="matplotlib"):
```
OriolAbril, line 60: backend should be none and read the rcParam as default

src/arviz_plots/style.py
```diff
@@ -13,16 +13,79 @@ def use(name):
...
+    name : str
+        Name of the style/template to get.
+    backend : str
+        Name of the backend to use. Options are 'matplotlib' (default) and 'plotly'.
+    """
+    if backend not in ["matplotlib", "plotly"]:
+        raise ValueError(f"Default styles/templates are not supported for Backend {backend}")
+
+    if backend == "matplotlib":
+        try:
```
OriolAbril, line 74: Given the backend is an input, I think we can skip the try except here. Otherwise not having matplotlib installed but using matplotlib here would result in the function not doing anything instead of raising an error

src/arviz_plots/style.py
```diff
@@ -39,53 +41,51 @@ def use(name):
...
         Name of the style/template to get.
     backend : str
-        Name of the backend to use. Options are 'matplotlib' (default) and 'plotly'.
+        Name of the backend to use. Options are 'matplotlib' and 'plotly'.
+        Defaults to ``rcParams["plot.backend"]``.
     """
-    if backend not in ["matplotlib", "plotly"]:
+    if backend is None:
+        backend = rcParams["plot.backend"]
+    elif backend not in ["matplotlib", "plotly"]:
```
OriolAbril, line 76:
```suggestion
    if backend not in ["matplotlib", "plotly"]:
```

I think it should still fail with this error if running wih bakcend=None and plot.backend rcparam set to bokeh

src/arviz_plots/style.py
```diff
@@ -39,53 +41,51 @@ def use(name):
...
+        import plotly.io as pio
 
-            if name in pio.templates:
-                return pio.templates[name]
-        except ImportError:
-            pass
+        if name in pio.templates:
+            return pio.templates[name]
 
     return ValueError(f"Style {name} not found.")
```
OriolAbril, line 91:
```suggestion
    raise ValueError(f"Style {name} not found.")
```

OriolAbril (approved):
wanted to make the last one approve review but forgot to click

PR #77, opened 2024-08-02
Add plot_compare

### LINKED ISSUE #76: Add plot_compare
There isn't any compare function in `arviz-stats` yet, but the idea is that when it is added it will basically return the same output as the current `arviz.compare` function. Moreover it is dataframe to "row-wise" plot, there aren't really aesthetics nor facetting at play and is independent of `PlotCollection`

aloctavodia:
closes #76  

There are two major points I want to discuss for this PR (or we can split the discussion into more than one PR).

* Is this the correct approach to implement this plot? I mean I am following the proper "new arviz logic"?
*  Modifications relative to the legacy `plot_compare`

about the second point I would like to simplify the plot and only show the ELPD point estimates + SE. Then we will get rid of the arguments `insample_dev`, `plot_standard_error`, `plot_ic_diff`, `order_by_rank`, `legend`. The main reason for this is that these are the most useful elements for the plot and this is how most people use and interpret this plot (including the equivalent plot in the Stan world).

I also would like to add two new arguments (already in the docstring), that I think could help make the plot easier to interpret.

*  similar_band : If True, a band is drawn to indicate models with similar predictive performance to the best model. (essentially 4 units from the point estimate of the best model) see figure below. I think this should be True by default
* relative_scale :  If True scale the ELPD values relative to the best model, so best model will be 0 and all the rest will be below 0 in the log-scale. Not sure if this should be true or false.

Captura desde 2024-08-02 11-23-08

**Update** 

This is how it looks like now in

matplotlib
output_matplotlib

plotly  (the band is missing from plotly, not sure why yet)
output_plotly

bokeh
output_bokeh

### DIFF
--- a/docs/source/gallery/model_comparison/plot_compare.py
+++ b/docs/source/gallery/model_comparison/plot_compare.py
@@ -1,0 +1,34 @@
+"""
+(gallery_forest_pp_obs)=
+# Posterior predictive and observations forest plot
+
+Overlay of forest plot for the posterior predictive samples and the actual observations
+
+---
+
+:::{seealso}
+API Documentation: {func}`~arviz_plots.plot_forest`
+
+Other gallery examples using `plot_forest`: {ref}`gallery_forest`, {ref}`gallery_forest_shade`
+:::
+"""
+from importlib import import_module
+
+from arviz_base import load_arviz_data
+
+import arviz_plots as azp
+
+azp.style.use("arviz-clean")
+
+backend="none"  # change to preferred backend
+
+cmp_df = pd.DataFrame({"elpd_loo": [-4.575778, -14.309050, -16], 
+                       "p_loo": [2.646204, 2.399241, 2], 
+                       "elpd_diff": [0.000000, 9.733272, 11], 
+                       "weight": [1.000000e+00, 3.215206e-13, 0], 
+                       "se": [2.318739, 2.673219, 2], 
+                       "dse": [0.00000, 2.68794, 2], 
+                       "warning": [False, False, False], 
+                       "scale": ["log", "log", "log"]}, index=["modelo_p", "modelo_l", "modelo_d"])
+
+azp.plot_compare(cmp_df, backend=backend)
--- a/docs/source/gallery/model_comparison/plot_compare.py
+++ b/docs/source/gallery/model_comparison/plot_compare.py
@@ -9,8 +9,9 @@ Compare multiple models using predictive accuracy estimates like  {abbr}`LOO-CV 
 API Documentation: {func}`~arviz_plots.plot_compare`
 :::
 """
-import arviz_plots as azp
 import pandas as pd
+
+import arviz_plots as azp
 
 azp.style.use("arviz-clean")
 
--- a/docs/source/gallery/model_comparison/plot_compare.py
+++ b/docs/source/gallery/model_comparison/plot_compare.py
@@ -1,22 +1,18 @@
 """
-(gallery_forest_pp_obs)=
-# Posterior predictive and observations forest plot
+# Predicive model comparison plot
 
-Overlay of forest plot for the posterior predictive samples and the actual observations
+Compare multiple models using predictive accuracy estimates like  {abbr}`LOO-CV (leave one out cross validation)` or {abbr}`WAIC (widely applicable information criterion)`
 
 ---
 
 :::{seealso}
-API Documentation: {func}`~arviz_plots.plot_forest`
-
-Other gallery examples using `plot_forest`: {ref}`gallery_forest`, {ref}`gallery_forest_shade`
+API Documentation: {func}`~arviz_plots.plot_compare`
 :::
 """
 import arviz_plots as azp
 
 azp.style.use("arviz-clean")
 
-backend="none"  # change to preferred backend
 
 cmp_df = pd.DataFrame({"elpd_loo": [-4.5, -14.3, -16.2], 
                        "p_loo": [2.6, 2.3, 2.1], 
@@ -27,4 +23,5 @@ cmp_df = pd.DataFrame({"elpd_loo": [-4.5, -14.3, -16.2],
                        "warning": [False, False, False], 
                        "scale": ["log", "log", "log"]}, index=["Model B", "Model A", "Model C"])
 
-azp.plot_compare(cmp_df, backend=backend)
+pc = azp.plot_compare(cmp_df, backend="none")  # change to preferred backend
+pc.show()
--- a/docs/source/gallery/model_comparison/plot_compare.py
+++ b/docs/source/gallery/model_comparison/plot_compare.py
@@ -10,6 +10,7 @@ API Documentation: {func}`~arviz_plots.plot_compare`
 :::
 """
 import arviz_plots as azp
+import pandas as pd
 
 azp.style.use("arviz-clean")
 
--- a/docs/source/gallery/model_comparison/plot_compare.py
+++ b/docs/source/gallery/model_comparison/plot_compare.py
@@ -12,23 +12,19 @@ API Documentation: {func}`~arviz_plots.plot_forest`
 Other gallery examples using `plot_forest`: {ref}`gallery_forest`, {ref}`gallery_forest_shade`
 :::
 """
-from importlib import import_module
-
-from arviz_base import load_arviz_data
-
 import arviz_plots as azp
 
 azp.style.use("arviz-clean")
 
 backend="none"  # change to preferred backend
 
-cmp_df = pd.DataFrame({"elpd_loo": [-4.575778, -14.309050, -16], 
-                       "p_loo": [2.646204, 2.399241, 2], 
-                       "elpd_diff": [0.000000, 9.733272, 11], 
-                       "weight": [1.000000e+00, 3.215206e-13, 0], 
-                       "se": [2.318739, 2.673219, 2], 
-                       "dse": [0.00000, 2.68794, 2], 
+cmp_df = pd.DataFrame({"elpd_loo": [-4.5, -14.3, -16.2], 
+                       "p_loo": [2.6, 2.3, 2.1], 
+                       "elpd_diff": [0, 9.7, 11.3], 
+                       "weight": [0.9, 0.1, 0], 
+                       "se": [2.3, 2.7, 2.3], 
+                       "dse": [0, 2.7, 2.3], 
                        "warning": [False, False, False], 
-                       "scale": ["log", "log", "log"]}, index=["modelo_p", "modelo_l", "modelo_d"])
+                       "scale": ["log", "log", "log"]}, index=["Model B", "Model A", "Model C"])
 
 azp.plot_compare(cmp_df, backend=backend)
--- a/src/arviz_plots/plots/compareplot.py
+++ b/src/arviz_plots/plots/compareplot.py
@@ -97,6 +97,7 @@ def plot_compare(
     figsize_units = pc_kwargs.get("plot_grid_kws", {}).get("figsize_units", "inches")
     chart, target = p_be.create_plotting_grid(1, figsize=figsize, figsize_units=figsize_units)
 
+    # Create plot collection
     plot_collection = PlotCollection(
         Dataset({}),
         viz_dt=DataTree.from_dict(
@@ -106,6 +107,7 @@ def plot_compare(
                 )
             }
         ),
+        backend=backend,
     )
 
     # Set scale relative to the best model
--- a/src/arviz_plots/plots/compareplot.py
+++ b/src/arviz_plots/plots/compareplot.py
@@ -93,8 +93,8 @@ def plot_compare(
 
     # Get figure params and create figure and axis
     pc_kwargs["plot_grid_kws"] = pc_kwargs.get("plot_grid_kws", {}).copy()
-    figsize = pc_kwargs.get("plot_grid_kws", {}).get("figsize", (10, len(cmp_df)))
-    figsize_units = pc_kwargs.get("plot_grid_kws", {}).get("figsize_units", "inches")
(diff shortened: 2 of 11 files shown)


src/arviz_plots/plots/compareplot.py
```diff
@@ -0,0 +1,111 @@
...
+
+        target = figure()
+    elif backend == "matplotlib":
+        import matplotlib.pyplot as plt
+
+        _, target = plt.subplots()
+    elif backend == "plotly":
+        import plotly.graph_objects as go
+
+        target = go.Figure()
```
aloctavodia, line 76: Not sure this works for bokeh or plotly

OriolAbril: I would use `import_module` to import the corresponding plotting backend submodule, then use `create_plotting_grid(1)` (maybe we could also make `number=1` default so it can be called with `create_plotting_grid()` to generate a single plot. https://github.com/arviz-devs/arviz-plots/blob/main/src/arviz_plots/backend/none/__init__.py#L103. I would avoid backend dependent ifs as much as possible because then adding new backends won't automatically add support for that plot its source code needs to be modified. For example, this one would not work with "none" backend currently, it is not very important one, but it can be useful and we use it extensively for testing as it is much faster than the other ones given it doesn't plot anything.

For the rest I think it would also be easier to use `plot_bknd.blah` instead of using visuals as I don't think you actually need the extra logic in visuals which is mostly about dataarray processing. https://arviz-plots.readthedocs.io/en/latest/gallery/plot_forest_pp_obs.html uses the import and plot_backend.xlabel so it can serve as an example.

docs/source/gallery/model_comparison/plot_compare.py
```diff
@@ -0,0 +1,34 @@
...
+
+from arviz_base import load_arviz_data
+
+import arviz_plots as azp
+
+azp.style.use("arviz-clean")
+
+backend="none"  # change to preferred backend
+
+cmp_df = pd.DataFrame({"elpd_loo": [-4.575778, -14.309050, -16],
```
aloctavodia, line 25: Not sure about this

OriolAbril: I think it is ok, it will be faster and it also probably looks nicer with 3 than with 2 models

docs/source/gallery/model_comparison/plot_compare.py
```diff
@@ -0,0 +1,34 @@
...
+cmp_df = pd.DataFrame({"elpd_loo": [-4.575778, -14.309050, -16], 
+                       "p_loo": [2.646204, 2.399241, 2], 
+                       "elpd_diff": [0.000000, 9.733272, 11], 
+                       "weight": [1.000000e+00, 3.215206e-13, 0], 
+                       "se": [2.318739, 2.673219, 2], 
+                       "dse": [0.00000, 2.68794, 2], 
+                       "warning": [False, False, False], 
+                       "scale": ["log", "log", "log"]}, index=["modelo_p", "modelo_l", "modelo_d"])
+
+azp.plot_compare(cmp_df, backend=backend)
```
OriolAbril, line 34: I am just realizing the example gallery relies on the returned output being a plot collection and using `pc.show()`. I'll comment below, even if it is not used, we could still return a PlotCollection for consistency and to help users update visual elements if desired (like for other plots, but we could skip that part too)

docs/source/gallery/model_comparison/plot_compare.py
OriolAbril, line 1: General comment about this, I am still working on the example gallery and automating parts of the content that won't need our input. For example, instead of manually listing examples using the same function, I am trying to add a directive that generates a grid of elements (inspired from sphinx-gallery extension)

What parts do you think should not be part of your input?

src/arviz_plots/plots/compareplot.py
```diff
@@ -0,0 +1,137 @@
...
+
+    if relative_scale:
+        cmp_df = cmp_df.copy()
+        cmp_df[i_c] = cmp_df[i_c] - cmp_df[i_c].iloc[0]
+
+    if figsize is None:
+        figsize = (10, len(cmp_df))
+
+    p_be = import_module(f"arviz_plots.backend.{backend}")
+    _, target = p_be.create_plotting_grid(1, figsize=figsize)
```
OriolAbril, line 82:
```suggestion
    chart, target = p_be.create_plotting_grid(1, figsize=figsize)
    plot_collection = PlotCollection(
        viz=DataTree.from_dict({
            "/": Dataset({"chart": np.array(chart, dtype=object), "plot": np.array(target, dtype=object)}
        })
    )
```

src/arviz_plots/plots/compareplot.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+def plot_compare(
+    cmp_df,
+    color="black",
+    similar_band=True,
+    relative_scale=False,
+    figsize=None,
+    target=None,
+    backend=None,
```
OriolAbril, line 14: I would maintain `plot_kwargs` because it still makes sense, the rest aes_map, stats_kwargs make no sense so I agree they should not be here.

src/arviz_plots/plots/compareplot.py
```diff
@@ -0,0 +1,137 @@
...
+
+    if backend is None:
+        backend = rcParams["plot.backend"]
+
+    if relative_scale:
+        cmp_df = cmp_df.copy()
+        cmp_df[i_c] = cmp_df[i_c] - cmp_df[i_c].iloc[0]
+
+    if figsize is None:
+        figsize = (10, len(cmp_df))
```
OriolAbril, line 79: This should be set in dots, matplotlib can handle dots without much issue whereas the rest can't handle inches at all. figsize related functions take both `figsize` and `figsize_units` so both work (default is inches for matplotlib, dots for the rest), but things we want to be proper backend agnostic should use dots insead. Multiplying by 100 or so should make the trick

src/arviz_plots/plots/compareplot.py
```diff
@@ -0,0 +1,137 @@
...
+    if similar_band:
+        if scale == "log":
+            x_0, x_1 = cmp_df[i_c].iloc[0] - 4, cmp_df[i_c].iloc[0]
+        else:
+            x_0, x_1 = cmp_df[i_c].iloc[0], cmp_df[i_c].iloc[0] + 4
+
+        p_be.fill_between_y(
+            x=[x_0, x_1],
+            y_bottom=yticks_pos[-1],
+            y_top=yticks_pos[0],
```
OriolAbril, line 122: I would make this cover a bit more than the exact y range where data is plotted.

OriolAbril (commented):
looks great, my main comment is using `plot_kwargs` and `setdefault` so it is more similar to other plots and offers similar felxibility

docs/source/gallery/model_comparison/plot_compare.py
aloctavodia: Not sure I understand your question.

src/arviz_plots/plots/compareplot.py
aloctavodia: Do I need to set data?

docs/source/gallery/model_comparison/plot_compare.py
OriolAbril: I removed the `(gallery_xyz)=` for exmple, which is now automated, left only title, description and seealso box as user input now. But maybe we can improve that further

src/arviz_plots/plots/compareplot.py
OriolAbril: I don't understand the question. Do you mean storing the artists back into the plot collection? If so no, only the return needs to be plotcollection for the gallery to work (moreover, the target in plotly is kind of a hack, the important object is the chart/figure one so this way both will be returned)

aloctavodia: I mean that PlotCollection has a data argument. And that argument is mandatory, right?

OriolAbril: Ooohhh, is it also required via the constructor? Maybe give it an empty dataset if so

aloctavodia: OK. I will check

aloctavodia: I passed and empty dataset and I also needed to pass the backend, otherwise I got an error for plotly and bokeh. 

```
File arviz-plots/src/arviz_plots/plot_collection.py:216, in PlotCollection.__init__(self, data, viz_dt, aes_dt, aes, backend, **kwargs)
         self.backend = backend
     elif "chart" in viz_dt:
          self.backend = viz_dt["chart"].item().__module__.split(".")[0]
     if aes is None and aes_dt is not None:
         aes = _get_aes_dict_from_dt(aes_dt)

AttributeError: 'NoneType' object has no attribute '__module__'
```

docs/source/gallery/model_comparison/plot_compare.py
```diff
@@ -0,0 +1,30 @@
+"""
+(gallery_forest_pp_obs)=
```
OriolAbril, line 2:
```suggestion
```

docs/source/gallery/model_comparison/plot_compare.py
```diff
@@ -0,0 +1,30 @@
...
+# Posterior predictive and observations forest plot
+
+Overlay of forest plot for the posterior predictive samples and the actual observations
```
OriolAbril, line 5:
```suggestion
# Predicive model comparison plot

Compare multiple models using predictive accuracy estimates like  {abbr}`LOO-CV (leave one out cross validation)` or {abbr}`WAIC (widely applicable information criterion)`
```

docs/source/gallery/model_comparison/plot_compare.py
```diff
@@ -0,0 +1,30 @@
...
+
+---
+
+:::{seealso}
+API Documentation: {func}`~arviz_plots.plot_forest`
+
+Other gallery examples using `plot_forest`: {ref}`gallery_forest`, {ref}`gallery_forest_shade`
```
OriolAbril, line 12:
```suggestion
API Documentation: {func}`~arviz_plots.plot_compare`
```

docs/source/gallery/model_comparison/plot_compare.py
```diff
@@ -0,0 +1,30 @@
...
+:::
+"""
+import arviz_plots as azp
+
+azp.style.use("arviz-clean")
+
+backend="none"  # change to preferred backend
```
OriolAbril, line 19:
```suggestion
```

docs/source/gallery/model_comparison/plot_compare.py
```diff
@@ -0,0 +1,30 @@
...
+cmp_df = pd.DataFrame({"elpd_loo": [-4.5, -14.3, -16.2], 
+                       "p_loo": [2.6, 2.3, 2.1], 
+                       "elpd_diff": [0, 9.7, 11.3], 
+                       "weight": [0.9, 0.1, 0], 
+                       "se": [2.3, 2.7, 2.3], 
+                       "dse": [0, 2.7, 2.3], 
+                       "warning": [False, False, False], 
+                       "scale": ["log", "log", "log"]}, index=["Model B", "Model A", "Model C"])
+
+azp.plot_compare(cmp_df, backend=backend)
```
OriolAbril, line 30:
```suggestion
pc = azp.plot_compare(cmp_df, backend="none")  # change to preferred backend
pc.show()
```

docs/source/gallery/model_comparison/plot_compare.py
```diff
@@ -0,0 +1,27 @@
+"""
+# Predicive model comparison plot
+
+Compare multiple models using predictive accuracy estimates like  {abbr}`LOO-CV (leave one out cross validation)` or {abbr}`WAIC (widely applicable information criterion)`
+
+---
+
+:::{seealso}
+API Documentation: {func}`~arviz_plots.plot_compare`
+:::
```
OriolAbril, line 10: If there are any docs/url for the dashboard version of plot compare we could link there in here too

OriolAbril (commented):
tried fixing the example for now, skimmed the code. It looks good, and uses the walrus! I had actually forgotten it existed 🤣 but I love it

src/arviz_plots/plots/compareplot.py
OriolAbril: yes, backend is needed too, otherwise the pc.show() part won't work. That was an initial attempt at identifying backend from the provided object, but it doesn't really work as you saw, it does work sometimes, but the automatic backend identification is still pending

docs/source/gallery/model_comparison/plot_compare.py
```diff
@@ -0,0 +1,27 @@
...
+"""
+import arviz_plots as azp
```
OriolAbril, line 12:
```suggestion
import arviz_plots as azp
import pandas as pd
```

docs/source/gallery/model_comparison/plot_compare.py
aloctavodia: Not now, we need to refactor the entire dashboard project.

src/arviz_plots/plots/compareplot.py
```diff
@@ -0,0 +1,190 @@
...
+                    {"chart": np.array(chart, dtype=object), "plot": np.array(target, dtype=object)}
+                )
+            }
+        ),
+        backend=backend,
+        **pc_kwargs,
+    )
+
+    if isinstance(target, np.ndarray):
+        target = target.tolist()
```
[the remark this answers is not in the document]
OriolAbril: It generally is an array of empty lists, so then `.map` subsets the array and target ends up being the list that represents the desired subplot. It is very probably broken for (1,1) "grid" though so we get an empty array instead of an array with a single empty list 🤔

PR #78, opened 2024-08-04
Improve backend documentation and get plot_forest to follow best practices

OriolAbril:


docs/sphinxext/gallery_generator.py
```diff
@@ -49,6 +54,33 @@
...
+
+   .. image:: /gallery/_images/{basename}.png
+      :alt:
+
+   +++
+   {title}
+"""
+
+minigallery_in_example = """
+f"## Other examples with `{fun}`
```
OriolAbril, line 77:
```suggestion
## Other examples with `{fun}`
```

src/arviz_plots/plots/forestplot.py
```diff
@@ -131,40 +131,27 @@ def plot_forest(
 
     Examples
     --------
-    The following examples focus on behaviour specific to ``plot_forest``.
+    The following examples focus on behaviour specific to ``plot_forest`` and aim to illustrate
+    usage of specific arguments.
```
aloctavodia, line 135:
```suggestion
```

src/arviz_plots/plots/forestplot.py
```diff
@@ -131,40 +131,27 @@ def plot_forest(
...
-    Default forest plot for multiple models:
-
-    .. plot::
-        :context: close-figs
-
-        >>> pc = plot_forest({"centered": centered, "non centered": non_centered})
-        >>> pc.add_legend("model")
+    usage examples see :ref:`plots_intro`. Alternatively, to see common usage patterns
+    or how the plot looks for different combinations, links to all examples using
+    ``plot_forest`` in the example gallery are shown at the bottom of the docstring.
```
aloctavodia, line 139:
```suggestion
```

src/arviz_plots/plots/forestplot.py
```diff
@@ -176,30 +163,7 @@ def plot_forest(
...
-        >>> ).assign_coords(column=["labels", "forest", "ess"])
-        >>> pc = plot_forest(c_aux, combined=True)
-        >>> pc.map(
-        >>>     visuals.scatter_x, "ess", data=centered.azstats.ess().ds,
-        >>>     coords={"column": "ess"}, color="C0"
-        >>> )
-
-    Note that we are using the same :class:`~.PlotCollection`, so when using
-    ``map`` all the same aesthetic mappings used by ``plot_forest`` are used.
+    .. minigallery:: plot_forest
```
aloctavodia, line 166:
```suggestion
     For a general introduction to batteries-included functions like this one and common
    usage examples see :ref:`plots_intro`. 
    .. minigallery:: plot_forest
```

src/arviz_plots/plots/forestplot.py
aloctavodia: I am not sure this intro is that useful so I suggest remove most of them and move the first sentence after the example and before the mini-gallery

docs/source/api/backend/index.rst
```diff
@@ -22,12 +22,44 @@ Available backends
...
+
+           .. grid-item::
+
+              Data
+
+           .. grid-item::
+
+              .. raw:: html
+
+                 <s>Viz</s>
```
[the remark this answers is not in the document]
OriolAbril: I think the best way forward would be to combine everything into an image. But didn't want to get into that to prematurely. I am still not sure about the backend name for example, if we keep this Data~Viz~ idea we could rename it to "data" instead of "none".

docs/sphinxext/gallery_generator.py
```diff
@@ -170,23 +213,95 @@ def main(app):
...
+    with open(gallery_dir / "backreferences.json", "w", encoding="utf-8") as f:
+        json.dump(backreferences, f)
+
     with open(gallery_dir / "index.md", "w", encoding="utf-8") as fi:
         fi.write("\n".join(index_page))
 
     os.chdir(working_dir)
 
 
+class MiniGallery(Directive):
```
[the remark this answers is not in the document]
OriolAbril: it only works on rst for now though (same as the one from sphinx-gallery which I adapted)

src/arviz_plots/plots/forestplot.py
OriolAbril: maybe we could get rid of that completely and link to plots_intro in a See Also section (which we should add at some point, similarly to the ones in current arviz)

src/arviz_plots/plots/forestplot.py
OriolAbril: Left a comment above, but putting these two together might give users the idea they are related resources when they are completely different. One is quite general about functions inside `arviz_plots.plots` and covers the genral logic for plot_kwargs, aes_map, a bit on how to handle the returned object. Whereas the other are specific examples using the function. 

IMO, people who are beginners and want to learn to use arviz-plots should go to the first resource, people who already have an idea on how to use it but are searching for a specific plot or want to see what do some of the arguments actually do should go to the 2nd

src/arviz_plots/plots/forestplot.py
aloctavodia: that also makes sense

OriolAbril:
Let's try this and keep iterating. It would also be nice to get feedback on the automatic figsize (currently in use for plot_forest, plot_trace and plot_trace_dist)

PR #79, opened 2024-08-05
Adding MCSE Plot 

imperorrp:
First commit for MCSE plot. A lot of ESS plot conventions are followed, including the extra methods (mean and sd) and `scatter_xy` visual element portions developed in it. 

The computing of quantiles for plotting the errorbars as desired requires subsetting the distribution so the subsets can be flattened and then the `arviz-stats` quantile function called- the `xarray_sel_iter` function from `arviz_base` was imported and attempted to be used for this. I then tried to concatenate the computed values to the `mcse_dataset` along the plot_axis dim as coordinate 'z'. There are some issues with achieving this though since the `xr.concat()` and `xr.merge()` are not in-place functions so the logic/algorithm for extending `mcse_dataset` like this might need more thought. There might be a simpler way to do this that I haven't thought of though. This bit is marked as WIP in the code. If we pick the alternative (below) approach it could be removed altogether. 

Since `pc.map()` automatically subsets DataArray/Dataset type objects passed to it, moving the computation to the new `error_bar` visual element function makes it a lot more simple so this is the current implementation I've kept in this commit. This requires calling `arviz-stats` and doing some computation in the visual element function which none of the other visual element functions do, but it works already. If any other plots with different logic also require usage of this visual element function then modifications might be needed. 

Also added the backend function `errorbar` for backend=None and backend=Matplotlib so far. The Bokeh and the new Plotly backend's implementations are left.

The current visual outputs of the plot are like this:

image

image

image

image

image

----
📚 Documentation preview 📚: https://arviz-plots--79.org.readthedocs.build/en/79/


src/arviz_plots/plots/mcseplot.py
```diff
@@ -0,0 +1,491 @@
...
+                else:
+                    z_mcse_dataset[var_name] = quantile_da_expanded
+                    # print(f"\n adding var {var_name} to z_mcse_dataset: {z_mcse_dataset}")
+
+            print(f"\n z_mcse_dataset = {z_mcse_dataset}")"""
+            print(len(a))
+
+            # print(f"\n final z_mcse_dataset = {z_mcse_dataset}")
+
+            # for now the quantile_values can be computed in the visual element function itself
```
OriolAbril, line 317: you already have `xdata_dataset`, also `mcse_y_dataset` (which now should be the errorbars, not the y position). I think `quantiles_dataset = distribution.quantile(probs, dim=mcse_dims)` will mostly give you what you want, the only problem is it will have a `quantile` dimension instead of an `mcse_dim` one, you can rename and reindex it or use quantile as dimension name for the other objects.

afer that you should be able to concat them into a new object with `plot_axis` having `x` (xdata_dataset), `y` (quantiles_dataset) and `error` (mcse_y_dataset) so the visual function can process that and still accept x, y in case we want aesthetics mapped to them

imperorrp:
Added the Bokeh backend function for errorbars. The `dash` function used in the legacy MCSE plot is deprecated now so `scatter` was used instead, and the  `multi_line` function was retained:

image

imperorrp:
Rebased the PR

aloctavodia:
closing as superseded by #211

Issue #52, opened 2024-06-04
Add Rootograms


aloctavodia:
Rootograms are a visual model checking tool for count models. You can read more here and  https://fromthebottomoftheheap.net/2016/06/07/rootograms/ here https://arxiv.org/abs/1605.01311

![output](https://github.com/arviz-devs/arviz-plots/assets/1338958/8fcb0483-b816-449c-a8ce-c44684f10d3e)

The black dots (and dashed lines) represent the observations and the bars are the predictions from the model. The predictions "hang" from the observations.  A perfect model will be one where the hanging bars touch the dashed grey lines at zero. Instead of the frequency, we plot its square root. This makes it easier to compare observed and expected frequencies
even for low frequencies.

The previous figure is from an example in BAP3. We should create a more Bayesian version and represent the uncertainty for the bars.

```Python
def rotogram(idata, ax):
    max_ = 17
    bins = np.array(range(0, max_))
    dims = idata.posterior_predictive.dims
    n_samples = dims["chain"] * dims["draw"]
    pred =  (np.histogram(idata.posterior_predictive["satellite"].values.ravel(),  bins=bins)[0] / n_samples)**0.5
    observed = np.histogram(crab["satellite"].values, bins=bins)[0]**0.5

    ax.bar(bins[:-1], observed, 0.5, bottom=pred-observed, color="C2")
    ax.plot(bins[:-1], pred,  "k.--")
    ax.hlines(0, -0.5, max_-1.5, linestyle="--", color="C1")


fig, axes = plt.subplots(2, 2, sharey=True, figsize=(12, 6))
for ax, idata, model in zip(axes.ravel(), 
                            [idata_crab_p, idata_crab_hp, idata_crab_nb, idata_crab_hnb],
                            [model_crab_p, model_crab_hp, model_crab_nb, model_crab_hnb]):
    rotogram(idata, ax)
    ax.set_title(model.family.name)

axes[1,1].set_xlabel("Satellites")
axes[1,0].set_xlabel("Satellites")
fig.text(-0.03, 0.5, "$\sqrt{Frequency}$", va="center", rotation="vertical", fontsize=14)
```

PR #81, opened 2024-08-12
Adding New Rootogram Plot 

imperorrp:
Pushed draft of the new Rootogram implementation code (#52 ) . 

- The existing hist functions from the currently still-open plot_dist hist addition PR #47 were used here as well as some code from the plot_ppc PR #55. 

- The hist visual element backend interface docstring was modified with 'y' now referring to the 'top' y-coordinate explicitly and not height of the bars (as suggested by @OriolAbril). The matplotlib implementation of the hist visual element (which uses `bar` behind the scenes) was also updated internally with `y=y-bottom` to reflect this. (This has to be updated in #47 now as well). The bars were not being plotted in the expected positions otherwise.

- WIP work includes appropriate binning for each facetted subset of the data to be plotted. 

- The observed data is used for binning for each subset and the bin heights here are required for setting the 'top's of the predictive bars so this is computed always, regardless of whether `observed` is passed as True or False (The True/False condition only determines whether it is plotted or not)

Current plot output when the `rugby` default Arviz datatree is passed (should work with any datatree with posterior predictive and observed groups like plot_ppc though): 

`azp.plot_rootogram(data)`
image

`azp.plot_rootogram(data, plot_kwargs={"predictive": False})`
image

`azp.plot_rootogram(data, observed=False)`
image

```
pc = azp.plot_rootogram(data, backend="bokeh")
pc.show()
```
image

`azp.plot_rootogram(data, observed_rug=True)`
image

^If we want a rugplot for rootograms as well like plot_ppc.

----
📚 Documentation preview 📚: https://arviz-plots--81.org.readthedocs.build/en/81/

### DIFF
--- a/src/arviz_plots/plots/rootogramplot.py
+++ b/src/arviz_plots/plots/rootogramplot.py
@@ -288,6 +288,78 @@ def plot_rootogram(
 
     print(f"\n new_obs_hist = {new_obs_hist}")
 
+    # ---------(PPC data)-------------
+
+    min_bottom = xr.Dataset()  # minimum value of the histogram 'bottoms, for observed_rug 'y'
+
+    pp_kwargs = copy(plot_kwargs.get("predictive", {}))
+
+    if pp_kwargs is not False:
+        pp_hist_dims, pp_hist_aes, pp_hist_ignore = filter_aes(
+            plot_collection, aes_map, "predictive", reduce_dims
+        )
+        # getting first default color from color cycle and picking it
+        pp_default_color = plot_bknd.get_default_aes("color", 1, {})[0]
+        if "color" not in pp_hist_aes:
+            pp_kwargs.setdefault("color", pp_default_color)
+
+        if "alpha" not in pp_hist_aes:
+            pp_kwargs.setdefault("alpha", 0.2)
+
+        pp_stats_kwargs = copy(stats_kwargs.get("predictive", {}))
+        pp_stats_kwargs.setdefault("bins", bins)
+
+        pp_hist = pp_distribution.azstats.histogram(
+            dims=list(pp_hist_dims) + list(sample_dims), **pp_stats_kwargs
+        )
+
+        # print(f"\n pp_density histogram form pp_hist = {pp_hist}")
+
+        # the top of the predictive bars height = the observed height for that bin
+        # the bottom = difference between observed and predictive height for that bin
+
+        # getting mean of counts across all predictive samples and taking its square root
+        pp_hist.loc[{"plot_axis": "histogram"}] = (
+            pp_hist.sel(plot_axis="histogram") / total_pp_samples
+        ) ** 0.5
+
+        # new_pp_hist dataset
+        new_pp_hist = xr.Dataset()
+
+        for var_name in list(pp_hist.keys()):
+            left_edges = pp_hist[var_name].sel(plot_axis="left_edges").values
+            right_edges = pp_hist[var_name].sel(plot_axis="right_edges").values
+
+            # getting top of histogram (observed values dataset's 'y' coord)
+            new_histogram = new_obs_hist[var_name].sel(plot_axis="y").values
+            # print(f"\n new_pp_hist (new_histogram) observed heights= {new_histogram}")
+
+            histogram_bottom = new_histogram - pp_hist[var_name].sel(plot_axis="histogram").values
+            # print(f"\n new_pp_hist histogram_bottom= {histogram_bottom}")
+
+            stacked_data = np.stack(
+                (new_histogram, left_edges, right_edges, histogram_bottom), axis=-1
+            )
+            new_var = xr.DataArray(
+                stacked_data,
+                dims=["hist_dim", "plot_axis"],
+                coords={
+                    "plot_axis": ["histogram", "left_edges", "right_edges", "histogram_bottom"]
+                },
+            )
+
+            new_pp_hist[var_name] = new_var
+            min_histogram_bottom = min(histogram_bottom)
+            min_bottom[var_name] = min_histogram_bottom - (0.2 * (0 - min_histogram_bottom))
+
+        print(f"\n new_pp_hist = {new_pp_hist}")
+
+        plot_collection.map(
+            hist, "predictive", data=new_pp_hist, ignore_aes=pp_hist_ignore, **pp_kwargs
+        )
+
+    # ------------(Observed)--------------------
+
     if observed:  # all observed data group plotting logic happens here
         obs_kwargs = copy(plot_kwargs.get("observed", {}))
         if obs_kwargs is not False:
@@ -339,6 +411,9 @@ def plot_rootogram(
             if "size" not in rug_aes:
                 rug_kwargs.setdefault("size", 30)
 
+            # rug_kwargs.setdefault("y", min_bottom)
+            print(f"\n min_bottom = {min_bottom}")
+
             # print(f"\nobs_distribution = {obs_distribution}")
 
             plot_collection.map(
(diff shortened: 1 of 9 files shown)


src/arviz_plots/plots/rootogramplot.py
```diff
@@ -0,0 +1,470 @@
...
+    # print(f"\n obs_distribution = {obs_distribution}")
+
+    # ---------(observed data)-----------
+    # observed data calculations are made outside of and before 'if observed' since predictive also
+    # depends on this computed data (number of bins and top of predictive bars for rootograms)
+
+    # use get_bins func from arviz-stats on observed data and then use those bins for
+    # computing histograms for predictive data as well
+    # WIP: currently only the bins for one variable (without any facetting) is retrieved and used
+    bins = array_stats.get_bins(obs_distribution["home_points"].values)
```
imperorrp, line 241: Reminder for later: Use xarray aware `get_bins` function once available from `Arviz-Stats`

imperorrp:
Todo: Shift rug lower, below the bottom of the bars

imperorrp:
Rugplots now render below the bottom of the rootogram bars. 

To allow for some tolerance gap for visibility, I've currently set this algorithm: 
```
min_histogram_bottom = min(histogram_bottom)
min_bottom[var_name] = min_histogram_bottom - (0.2 * (0 - min_histogram_bottom))
```

(It may be that picking a specific number to subtract from min_histogram is better as that'd keep this gap a constant number of units, but then the chart size may vary and so would the perceivable gap) 

Output looks like this now when the rugplot is plotted: 

image

src/arviz_plots/plots/rootogramplot.py
```diff
@@ -0,0 +1,449 @@
...
+    # wrap plot collection with pp distribution
+    if plot_collection is None:
+        pc_kwargs.setdefault("col_wrap", 5)
+        pc_kwargs.setdefault(
+            "cols",
+            ["__variable__"]  # special variable to create one plot per variable
+            + list(facet_dims),  # making sure multiple plots are created for each facet dim
+        )
+        pc_kwargs["aes"] = pc_kwargs.get("aes", {}).copy()
+        pc_kwargs["aes"].setdefault("overlay", sample_dims)  # setting overlay dim
```
OriolAbril, line 192: This can be removed, we only want facetting in rootgram

src/arviz_plots/plots/rootogramplot.py
```diff
@@ -0,0 +1,449 @@
...
+from arviz_plots.visuals import hist, labelled_title, line_xy, scatter_xy, trace_rug
+
+
+def plot_rootogram(
+    dt,
+    var_names=None,
+    filter_vars=None,
+    group="posterior",
+    observed=None,
+    observed_rug=False,
```
OriolAbril, line 23: I would remove these two. Not including the observed data doesn't really make sense for the rootgram. We could skip the y shift in such cases but then the resulting plot would be a histogram directly, if users want a "regular" histogram they should use `plot_dist`.

Moreover, as you might have noticed, the rug plot doesn't really give much information because the data is discrete, so there are many lines overlapping at 1, many lines overlapping at 2... so the rugplot doesn't really show how many observations do we have at 1 or at 2, it seems each has only a single observation. And playing with alpha can only help so much, at most you'd get a qualitative information about the number of variables at each value which is less info than the observed line/scatter provide. So I would remove the argument and the rugplot.

src/arviz_plots/plots/rootogramplot.py
```diff
@@ -0,0 +1,449 @@
...
+            """plot_kwargs['observed'] can't be False, use observed=False to remove observed 
+            plot element"""
+        )
+
+    # making sure both posterior/prior predictive group and observed_data group exists in
+    # datatree provided
+    if observed:
+        for group_name in (predictive_data_group, "observed_data"):
+            if group_name not in dt.children:
+                raise TypeError(f'`data` argument must have the group "{group_name}" for ppcplot')
```
OriolAbril, line 148: After removing `observed` just keep this outside the if.

src/arviz_plots/plots/rootogramplot.py
```diff
@@ -0,0 +1,449 @@
...
+
+        x = (left_edges + right_edges) / 2
+        y = obs_hist[var_name].sel(plot_axis="histogram").values
+
+        stacked_data = np.stack((x, y), axis=-1)
+        new_var = xr.DataArray(
+            stacked_data, dims=["hist_dim", "plot_axis"], coords={"plot_axis": ["x", "y"]}
+        )
+
+        new_obs_hist[var_name] = new_var
```
OriolAbril, line 269:
```suggestion
new_obs_hist = xr.concat(
    (ds.sel(plot_axis=["right_edges", "left_edges"]).sum("plot_axis") / 2, obs_hist.sel(plot_axis="histogram", drop=True)),
    dim="plot_axis",
).assign_coords(plot_axis=["x", "y"])
```

src/arviz_plots/plots/rootogramplot.py
```diff
@@ -0,0 +1,449 @@
...
+
+        # getting mean of counts across all predictive samples and taking its square root
+        pp_hist.loc[{"plot_axis": "histogram"}] = (
+            pp_hist.sel(plot_axis="histogram") / total_pp_samples
+        ) ** 0.5
+
+        # new_pp_hist dataset
+        new_pp_hist = xr.Dataset()
+
+        for var_name in list(pp_hist.keys()):
```
OriolAbril, line 307: also update and get rid of the loop

src/arviz_plots/plots/rootogramplot.py
OriolAbril: Also note that the code in the PR would only have worked when all variables get the same number of bins (which doesn't always happen) because the dimension name is hardcoded to `hist_dim` in all variables of the dataset. As I mentioned in slack, the behaviour has changed in the `get_bins` PR to support this effect. Now each variable gets independent dimensions `hist_dim_mu`, `hist_dim_tau`...

imperorrp:
Tests are failing due to `plot_dist` modification requirement since `plot_rootogram` currently depends on the unmerged arviz-stats 'get_bins' branch. This modification will be required globally once this Arviz-Stats branch is merged too

imperorrp:
Added tests for `plot_rootogram`

aloctavodia:
An example of a "hanging" rootogram  and a "suspended" rootogram, including the uncertainty for the predictions
https://github.com/arviz-devs/Exploratory-Analysis-of-Bayesian-Models/pull/60

imperorrp:
> An example of a "hanging" rootogram and a "suspended" rootogram, including the uncertainty for the predictions [arviz-devs/Exploratory-Analysis-of-Bayesian-Models#60](https://github.com/arviz-devs/Exploratory-Analysis-of-Bayesian-Models/pull/60). We do not necessarily need to follow those examples, they are just suggestions. Actually, once we have a working rootograms that example will use whatever arviz implements.

If we follow those examples, the `line_y` visual element function from #85 can be used for the thin lines to represent the hanging and suspended lines. For the hanging case, we can use the existing logic and for suspended we could modify the logic in this portion of the code:

``` 
histogram_bottom = new_obs_hist.sel(plot_axis="y") - pp_hist.sel(plot_axis="histogram")
histogram_bottom = histogram_bottom.expand_dims(plot_axis=["histogram_bottom"])
# print(f" diff = {a}\n")

new_pp_hist = xr.concat(
    (
        new_obs_hist.sel(plot_axis="y"),  # getting tops of histogram (observed values)
        pp_hist.sel(plot_axis="left_edges"),
        pp_hist.sel(plot_axis="right_edges"),
        histogram_bottom,
    ),
    dim="plot_axis",
).assign_coords(plot_axis=["histogram", "left_edges", "right_edges", "histogram_bottom"])

```

And instead pass a `histogram_top` along the histogram dimension (root of predictive count minus observed count) and 0 for the histogram_bottom dimension: 

```
histogram_top = pp_hist.sel(plot_axis="histogram") - new_obs_hist.sel(plot_axis="y")
histogram_top = histogram_top.expand_dims(plot_axis=["histogram"])
```

aloctavodia:
Closing in favor of #142

Issue #14, opened 2023-10-05
Add plot_violin


aloctavodia:

imperorrp:
The violin plot in its current implementation in #85 overlays the densities of multiple models when multiple models are provided on both 'left' and 'right' sides (although these sides can be toggled on/off with `plot_kwargs`). Do we want to keep this behavior or @OriolAbril's suggestion to have the first two models be mapped to the 'left' and 'right' sides of the violin plot, and then perhaps overlay the additional models beyond 2, or something else entirely?

OriolAbril:
My proposal is to raise an error if there are more than two models. I think overlaying things on a violin plot is a terrible idea and should never happen.

imperorrp:
Okay, I'll add in a check and make the modifications to do so

aloctavodia:
We have decided not to implement violin plots.

PR #85, opened 2024-08-29
First commit for violinplot

imperorrp:
First commit for violinplot ( #14 )

`plot_violin`  computes kde marginal densities like `plot_dist` does and then flips the 'x' and 'y' values to generate the characteristic violin-shape. 

- Separate artists have been defined for 'left' and 'right' densities and thus can be customized aesthetically or switched on/off with relevant `plot_kwargs`.  In addition, I've also computed them separately so that this computation can also be customized through `stats_kwargs`, through things like bandwidth and others. 

- The `line_xy` visual element was modified with a new keyword arg 'negative' that makes either the `sel(plot_axis='x')` or `sel(plot_axis='y')` dataarrays negative

- New visual elements were added for `line_y` and `scatter_y` which are pretty much symmetrical to `line_x` and `scatter_x` but for the other axis. 

- The 'y' aesthetic mapping applied to the credible interval, point estimate and point estimate text in `plot_dist` was converted to an 'x' aesthetic in case of multiple models- this needs refinement though (Maybe a mapping akin to the one in the ESS Plot PR #58 ) 

- The `remove_axis` is now applied to the 'x' axis and a modification was required in the matplotlib backend for remove_axis for this

Possible improvement: having a separate 'face' and 'edge' artist for the left and right densities, like plot_ridge. This will require creating a symmetrical visual element for `fill_between_x` like `fill_between_y`.

----
📚 Documentation preview 📚: https://arviz-plots--85.org.readthedocs.build/en/85/

### DIFF
--- a/src/arviz_plots/backend/matplotlib/__init__.py
+++ b/src/arviz_plots/backend/matplotlib/__init__.py
@@ -234,6 +234,7 @@ def line(x, y, target, *, color=unset, alpha=unset, width=unset, linestyle=unset
     """Interface to matplotlib for a line plot."""
     artist_kws.setdefault("zorder", 2)
     kwargs = {"color": color, "alpha": alpha, "linewidth": width, "linestyle": linestyle}
+    print(f"\n kws = {kwargs}")
     return target.plot(x, y, **_filter_kwargs(kwargs, Line2D, artist_kws))[0]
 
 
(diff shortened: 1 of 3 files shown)


imperorrp:
Current plot outputs: 

`azp.plot_violin(data, var_names=["mu", "tau"])`
image

`azp.plot_violin({"centered": data, "non centered": data2}, var_names=["mu", "tau"])`
image

`azp.plot_violin(data)`
image

imperorrp:
Added an error check preventing more than 2 models from being passed. I've tried to subselect the two models for the left and right violin densities, but it seems like two colors are still being passed over to the backends for the line visual element by `pc.map`, one for each of the models if they were overlapped instead. This has to be fixed somehow.

PR #93, opened 2024-10-09
Add plot_psense_dist

### LINKED ISSUE #92: Add priorsense dens and ecdf plots
Depends on https://github.com/arviz-devs/arviz-stats/issues/23

aloctavodia:
closes  #92 

~quick early draft. Eventually,~ it should look similar to this https://n-kall.github.io/priorsense/reference/figures/README-dens_plot-1.png

KDE + pointintervals (default)

kde

ecdf + pointervals

ecdf

histogram + point intervals (not clear at all, we could try using a "step" histogram and the same number of bins)

hist

pointervals 

point_interval

----
📚 Documentation preview 📚: https://arviz-plots--93.org.readthedocs.build/en/93/

### DIFF
--- a/src/arviz_plots/plots/psensedistplot.py
+++ b/src/arviz_plots/plots/psensedistplot.py
@@ -97,7 +97,7 @@ def plot_psense_dist(
     PlotCollection
     """
     if sample_dims is None:
-        sample_dims = rcParams["data.sample_dims"]
+        sample_dims = "sample"
     if isinstance(sample_dims, str):
         sample_dims = [sample_dims]
     if kind is None:
@@ -110,10 +110,6 @@ def plot_psense_dist(
         pc_kwargs = {}
     else:
         pc_kwargs = pc_kwargs.copy()
-
-    # distribution = process_group_variables_coords(
-    #     dt, group=group, var_names=var_names, filter_vars=filter_vars, coords=coords
-    # )
 
     if alphas is None:
         alphas = (0.8, 1.25)
--- a/src/arviz_plots/plots/psensedistplot.py
+++ b/src/arviz_plots/plots/psensedistplot.py
@@ -138,8 +138,8 @@ def plot_psense_dist(
     # Instead we could have weighted KDEs/ecdfs/etc
     ds_prior = new_ds(dt, "prior", alphas, sample_dims=sample_dims)
     ds_likelihood = new_ds(dt, "likelihood", alphas, sample_dims=sample_dims)
-    distribution = concat([ds_prior, ds_likelihood], dim="__group__").assign_coords(
-        {"__group__": ["prior", "likelihood"]}
+    distribution = concat([ds_prior, ds_likelihood], dim="component_group").assign_coords(
+        {"component_group": ["prior", "likelihood"]}
     )
     distribution = process_group_variables_coords(
         distribution, group=None, var_names=var_names, filter_vars=filter_vars, coords=coords
@@ -173,11 +173,15 @@ def plot_psense_dist(
         pc_kwargs.setdefault("y", [-0.4, -0.225, -0.05])
         pc_kwargs["aes"].setdefault("color", ["alpha"])
         pc_kwargs["aes"].setdefault("y", ["alpha"])
-        pc_kwargs.setdefault("cols", ["__group__"])
+        pc_kwargs.setdefault("cols", ["component_group"])
         pc_kwargs.setdefault(
             "rows",
             ["__variable__"]
-            + [dim for dim in distribution.dims if dim not in sample_dims + ["__group__", "alpha"]],
+            + [
+                dim
+                for dim in distribution.dims
+                if dim not in sample_dims + ["component_group", "alpha"]
+            ],
         )
 
         figsize = pc_kwargs["plot_grid_kws"].get("figsize", None)
--- a/src/arviz_plots/plots/psensedistplot.py
+++ b/src/arviz_plots/plots/psensedistplot.py
@@ -8,6 +8,7 @@ from xarray import concat
 
 from arviz_plots.plot_collection import PlotCollection, process_facet_dims
 from arviz_plots.plots.distplot import plot_dist
+from arviz_plots.plots.utils import process_group_variables_coords
 
 
 def plot_psense_dist(
@@ -15,7 +16,7 @@ def plot_psense_dist(
     alphas=None,
     var_names=None,
     filter_vars=None,
-    group="posterior",
+    # group="posterior",
     coords=None,
     sample_dims=None,
     kind=None,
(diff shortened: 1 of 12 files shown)


src/arviz_plots/plots/psensedistplot.py
```diff
@@ -0,0 +1,292 @@
...
+
+    resampled = [extract(dt, group="posterior").drop("chain")]
+    for weights in (lower_w, upper_w):
+        resampled.append(
+            extract(
+                dt,
+                group="posterior",
+                num_samples=s_size,
+                weights=weights,
+                random_seed=42,
```
aloctavodia, line 286: I think there is no need to allow the user to change the random_seed.

src/arviz_plots/plots/psensedistplot.py
```diff
@@ -0,0 +1,292 @@
...
+        stats_kwargs.setdefault("density", {"density": True})
+
+    if alphas is None:
+        alphas = (0.8, 1.25)
+
+    # Here we are generating new datasets for the prior and likelihood
+    # by resampling the original dataset with the power scale weights
+    # Instead we could have weighted KDEs/ecdfs/etc
+    ds_prior = new_ds(dt, "log_prior", alphas)
+    ds_likelihood = new_ds(dt, "log_likelihood", alphas)
```
OriolAbril, line 210: I would actually merge these two datasets along a `component` dimension which is what I'd use as the facetting dimension for the columns. Then a single call to `plot_dist` should be enough plus we'll automatically get the correct labels in the titles, with also the changes in `new_ds` we'd have to hardcode `sample_dims="sample"` when calling `plot_dist` but that is all.

I'd also do that before initializing the plotcollection so we can use `cols=["component"]` and `aes={"color": ["alpha"]}` when defining mappings.

src/arviz_plots/plots/psensedistplot.py
```diff
@@ -0,0 +1,292 @@
...
+                group="posterior",
+                num_samples=s_size,
+                weights=weights,
+                random_seed=42,
+                resampling_method="stratified",
+            ).drop("chain")
+        )
+
+    # Middle chain is the reference chain (alpha == 1)
+    return concat(resampled, dim="chain").rename({"sample": "draw"}).sel(chain=[1, 0, 2])
```
OriolAbril, line 292:
```suggestion
    resampled = [
        extract(
            dt,
            group="posterior",
            num_samples=s_size,
            weights=weights,
            random_seed=42,
            resampling_method="stratified",
        ).drop_indexes(["sample", "chain", "draw"])
        for weights in (lower_w, upper_w)
    ]
    resampled.insert(1, extract(dt, group="posterior").drop_indexes(["sample", "chain", "draw"]))

    return concat(resampled, dim="alpha").assign_coords(alpha=[alphas[0], 1, alphas[1]])
```

This should generate a dataset with dimensions `alpha` (to overlay over) and `sample` (to reduce) plus whatever dimensions were there already. Consequently, when we generate the legend with `plot_collection.legend("alpha")` it will title the legend `alpha` and use the correct values in the legend entries automatically.

OriolAbril (commented):
We need some work specially with templates and docs on how to write new plots but I really like that we can have so many plots be 1) data wrangling 2) create plot_collection 3) call plot_dist

src/arviz_plots/plots/psensedistplot.py
```diff
@@ -0,0 +1,249 @@
...
+
+    pc_kwargs : mapping
+        Passed to :class:`arviz_plots.PlotCollection.wrap`
+
+    Returns
+    -------
+    PlotCollection
+    """
+    if sample_dims is None:
+        sample_dims = "sample"
```
aloctavodia, line 100: setting `sample_dims = "sample"`, because the result of `new_ds` is a stacked dataset. Not sure how to unstacked, given that the indexes are duplicated.

OriolAbril: We'll want to do this when calling plot_dist, because by then we have already stacked all provided sample_dims into `sample`. As of now however it isn't really an argument to the function, everything is hardcoded to chain and draws (already on the arviz-stats side).

aloctavodia:
This is ready to merge on my side.

OriolAbril:
we should add some more docs and tests (even if built on top of arviz-stats arviz-plots is not so lawless 😅 )

aloctavodia:
haha, OK

src/arviz_plots/plots/psensedistplot.py
```diff
@@ -0,0 +1,282 @@
...
+        alphas = (0.8, 1.25)
+
+    # Here we are generating new datasets for the prior and likelihood
+    # by resampling the original dataset with the power scale weights
+    # Instead we could have weighted KDEs/ecdfs/etc
+    ds_prior = new_ds(dt, "prior", alphas, sample_dims=sample_dims)
+    ds_likelihood = new_ds(dt, "likelihood", alphas, sample_dims=sample_dims)
+    distribution = concat([ds_prior, ds_likelihood], dim="__group__").assign_coords(
+        {"__group__": ["prior", "likelihood"]}
+    )
```
OriolAbril, line 143: I really dislike the `__group__` after having tried it with `__sample__` in current ArviZ for `psislw`. I'd only use things like that for dimensions that are either immediately reduced right after or never make their way to the user. The `__group__` here will be a dimension on the generated plot_collection plus a dimension with aesthetics mapped to it which the user might want to modify.

Might not be the best idea, but a more compact view in a single column is also possible for example:


```python
from arviz_plots import plot_psense_dist, style

style.use("arviz-clean")

idata = load_arviz_data("rugby")

pc = plot_psense_dist(
    idata,
    var_names=["sd_def", "defs"],
    coords={"team": ["Scotland", "Wales"]},
    plot_kwargs={"kde": False},
    pc_kwargs={
        "cols": None,
        "aes": {"y": ["alpha", "__group__"], "linestyle": ["__group__"]},
        "y": [-1.2, -0.8, 0, 0, 0.8, 1.2],
    },
    aes_map={"credible_interval": ["color", "y", "linestyle"]}
)
pc.add_legend("__group__")
pc.show()
```

imatge

The `__` needs to be added to the aesthetic definitions, and even makes it into the legend.

I can also see my choices so far being a bit too optimistic on name clashing (probably some tests had data with a `group` dimension and broke 🙈 ), but in that case I'd try to find a slightly different name not so prone to clashing with a variable or dimension. How about `psense_group`? I know it is actually longer but it doesn't give out so much of "this is a special thing" vibe

aloctavodia: Yeah, not happy either. I changed exactly because a clash with a test. What do you think about `group_component`, too long? And`group_c` or  `group_p`. Can we add a blank space? Too confusing?

OriolAbril: I like `component_group` feels a bit more natural. I generally don't care too much about things getting on the longer side, I think clear and intuitive labels are a feature worth the extra characters

OriolAbril (approved):

PR #102, opened 2024-11-06
move out new_ds to arviz-stats

aloctavodia:
Depends on arviz-devs/arviz-stats#38

----
📚 Documentation preview 📚: https://arviz-plots--102.org.readthedocs.build/en/102/


OriolAbril:
We can modify the dependencies to install arviz-stats from github: https://github.com/arviz-devs/arviz-plots/blob/2758b924e5fdb620b4ac3fc645ba29e3369fc930/pyproject.toml#L30 (which we have done multiple times). Otherwise we'd have to either release constantly or end up developing very slowly

PR #105, opened 2024-11-15
Add plot for distribution of convergence diagnostics

aloctavodia:
I don't have a good name for this plot so its code name is `plot_nose`. The motivation is to visually inspect the values of ESS and R-hat for many variables. 

In PyMC-BART we have a [plot_convergence](https://www.pymc.io/projects/bart/en/latest/examples/bart_introduction.html#convergence-diagnostics) function that generates a similar plot, which we used to check the "BART variables" which are always multidimensional with a size at least equal to the observations.  The interest is not on any individual variables. This type of plot may be of interest to others outside of PyMC-BART.

```python
azp.plot_convergence_dist(radon, var_names=["za_county"])
```

I feel tempted to call "rhat_rank" just "rhat". What do you think?

nose_default

specific diagnostics can be passed as strings. 

```python
azp.plot_convergence_dist(radon, var_names=["za_county"], diagnostics=["ess_tail", "ess_median"])
```

nose_method

some ess methods accept a prob argument, we can pass it also as part of the string, like a "tuple"

```python
azp.plot_convergence_dist(radon,
              var_names=["za_county"],
              diagnostics=["ess_tail(0.1, 0.9)", "ess_local(0.1, 0.9)", "ess_quantile(0.9)"],
              ref_line=False)
```

nose_probs

~We could also add reference lines for ESS and r-hat following standard recommendations~ (added). For PyMC-BART, we have an "inflated" R-hat reference value. We use a [heuristics](https://github.com/pymc-devs/pymc-bart/blob/main/pymc_bart/utils.py#L126C1-L128C84) to increase the 1.01 value based on the number of variables. Not saying we should use that heuristic, just saying we may want to have something similar.

----
📚 Documentation preview 📚: https://arviz-plots--105.org.readthedocs.build/en/105/

### DIFF
--- a/docs/source/gallery/inference_diagnostics/plot_convergence_dist.py
+++ b/docs/source/gallery/inference_diagnostics/plot_convergence_dist.py
@@ -6,7 +6,7 @@ Plot the distribution of ESS and R-hat.
 ---
 
 :::{seealso}
-API Documentation: {func}`~arviz_plots.plot_ess`
+API Documentation: {func}`~arviz_plots.plot_convergence_dist`
 :::
 """
 
@@ -17,9 +17,10 @@ import arviz_plots as azp
 azp.style.use("arviz-clean")
 
 data = load_arviz_data("radon")
-pc = azp.plot_convergence_dist(data,
-                               var_names=["za_county"],
-                               backend="none",  # change to preferred backend
+pc = azp.plot_convergence_dist(
+    data,
+    var_names=["za_county"],
+    backend="none",  # change to preferred backend
 )
 
 pc.show()
--- a/docs/source/gallery/inference_diagnostics/plot_convergence_dist.py
+++ b/docs/source/gallery/inference_diagnostics/plot_convergence_dist.py
@@ -1,0 +1,25 @@
+"""
+# Convergence diagnostics distribution plot
+
+Plot the distribution of ESS and R-hat.
+
+---
+
+:::{seealso}
+API Documentation: {func}`~arviz_plots.plot_ess`
+:::
+"""
+
+from arviz_base import load_arviz_data
+
+import arviz_plots as azp
+
+azp.style.use("arviz-clean")
+
+data = load_arviz_data("radon")
+pc = azp.plot_ess(
+    data,
+    var_names=["za_county"],
+    backend="none",  # change to preferred backend
+)
+pc.show()
--- a/docs/source/gallery/inference_diagnostics/plot_convergence_dist.py
+++ b/docs/source/gallery/inference_diagnostics/plot_convergence_dist.py
@@ -17,9 +17,9 @@ import arviz_plots as azp
 azp.style.use("arviz-clean")
 
 data = load_arviz_data("radon")
-pc = azp.plot_ess(
-    data,
-    var_names=["za_county"],
-    backend="none",  # change to preferred backend
+pc = azp.plot_convergence_dist(data,
+                               var_names=["za_county"],
+                               backend="none",  # change to preferred backend
 )
+
 pc.show()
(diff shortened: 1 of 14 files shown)


AlexAndorra:
Anything I can help with to merge this one @aloctavodia ?

docs/source/gallery/inference_diagnostics/plot_convergence_dist.py
```diff
@@ -0,0 +1,25 @@
...
+"""
+
+from arviz_base import load_arviz_data
+
+import arviz_plots as azp
+
+azp.style.use("arviz-clean")
+
+data = load_arviz_data("radon")
+pc = azp.plot_ess(
```
AlexAndorra, line 20: Shouldn't this be `azp.plot_convergence_dist`?

AlexAndorra (approved):
Looks all good @aloctavodia , except for a typo I think. After that we can merge

docs/source/gallery/inference_diagnostics/plot_convergence_dist.py
aloctavodia: Good catch!

AlexAndorra (approved):
LGTM now

src/arviz_plots/plots/convergencedistplot.py
```diff
@@ -0,0 +1,202 @@
...
+
+        ess_ref = dt.sizes["chain"] * 100
+        # is this valid for all r_hat methods? Do we want to correct for multiple comparisons?
+        r_hat_ref = 1.01
+        for diagnostic in diagnostics:
+            if "ess" in diagnostic:
+                ref = ess_ref
+            if "rhat" in diagnostic:
+                ref = r_hat_ref
+            plot_collection.viz[diagnostic]["plot"].item().axvline(
```
aloctavodia, line 177: This only works with Matplotlib as other backends not necessarily have the `axvline` method

OriolAbril: yep, plotting backends should not be called directly but via thin wrappers that act as translators so our plotting functions see a common api independently of the backend being used. It also needs tests before it can be merged, I'll try to push it forward a bit more now

OriolAbril:
> For PyMC-BART, we have an "inflated" R-hat reference value. We use a [heuristics](https://github.com/pymc-devs/pymc-bart/blob/main/pymc_bart/utils.py#L126C1-L128C84) to increase the 1.01 value based on the number of variables. Not saying we should use that heuristic, just saying we may want to have something similar.

We probably want these two to be arguments to the function.

> I feel tempted to call "rhat_rank" just "rhat". What do you think?

I would actually use both (might need a bit extra code to parse diagnostic names though). My thinking/proposed logic. ArviZ defaults are subject to change as new better algorightms/variations of them are developed. `rhat` would be calling `rhat` without specifiying any method so the default is used whereas `rhat_rank` would be `rhat(..., method="rank")` which currently matches the default but might not match the default in the future. Consequently, the default diagnostics should probably be `ess_bulk, ess_tail, rhat`. We could do the same for ess, but not sure it is as relevant and in that case I'd leave the default diagnostics to `ess_bulk, ess_tail, rhat`

OriolAbril:
Already working example with all backends shown: https://arviz-plots--105.org.readthedocs.build/en/105/gallery/plot_convergence_dist.html (bokeh also has the vertical line but it doesn't extend the axis limits up to the vertical line so ends up hidden, a bit weird but hopefully not super critical)

aloctavodia:
LGTM, not sure what's the source of the new error.

OriolAbril:
The old error was pre-commit. I have started adding tests and realized we need to work a bit on the API for the ess methods (test is currently failing but pushed anyway so it is here already)

aloctavodia:
sorry for the mess!

PR #108, opened 2024-11-18
Add energy plot

aloctavodia:
```python
schools = azb.load_arviz_data('centered_eight')
azp.plot_energy(schools)
```

energy_kde

```python
azp.plot_energy(schools, kind="ecdf")
```

energy_ecdf

This is not working, because we have a `np.nan` value
```python
azp.plot_energy(schools, kind="hist")
```

----
📚 Documentation preview 📚: https://arviz-plots--108.org.readthedocs.build/en/108/

### DIFF
--- a/src/arviz_plots/plots/energyplot.py
+++ b/src/arviz_plots/plots/energyplot.py
@@ -94,6 +94,11 @@ def plot_energy(
         pc_kwargs = pc_kwargs.copy()
 
     new_ds = _get_energy_ds(dt)
+
+    sample_dims = ["chain", "draw"]
+    if not all(dim in new_ds.dims for dim in sample_dims):
+        raise ValueError("Both 'chain' and 'draw' dimensions must be present in the dataset")
+
     pc_kwargs.setdefault("cols", None)
     pc_kwargs["aes"] = pc_kwargs.get("aes", {}).copy()
     pc_kwargs["aes"].setdefault("color", ["energy"])
@@ -108,7 +113,7 @@ def plot_energy(
         filter_vars=None,
         group=None,
         coords=None,
-        sample_dims=None,
+        sample_dims=sample_dims,
         kind=kind,
         point_estimate=None,
         ci_kind=None,
--- a/src/arviz_plots/plots/energyplot.py
+++ b/src/arviz_plots/plots/energyplot.py
@@ -1,0 +1,138 @@
+"""Energy plot code."""
+import numpy as np
+from arviz_base import convert_to_dataset, rcParams
+
+from arviz_plots.plots.distplot import plot_dist
+
+
+def plot_energy(
+    dt,
+    bfmi=False,
+    kind=None,
+    plot_collection=None,
+    backend=None,
+    labeller=None,
+    aes_map=None,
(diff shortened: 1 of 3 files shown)


src/arviz_plots/plots/energyplot.py
```diff
@@ -0,0 +1,138 @@
...
+        plot_kwargs = {}
+    else:
+        plot_kwargs = plot_kwargs.copy()
+    if pc_kwargs is None:
+        pc_kwargs = {}
+    else:
+        pc_kwargs = pc_kwargs.copy()
+
+    new_ds = _get_energy_ds(dt)
+    pc_kwargs.setdefault("cols", ["__variable__"])
```
OriolAbril, line 97:
```suggestion
    pc_kwargs.setdefault("cols", None)
```

Does it work like this? There will be only a variable in `new_ds` so the result should be the same, but using `__variable__` feels a bit confusing

src/arviz_plots/plots/energyplot.py
```diff
@@ -0,0 +1,138 @@
...
+    plot_kwargs.setdefault("point_estimate_text", False)
+    plot_kwargs.setdefault("title", False)
+
+    plot_collection = plot_dist(
+        new_ds,
+        var_names=None,
+        filter_vars=None,
+        group=None,
+        coords=None,
+        sample_dims=None,
```
OriolAbril, line 111: I don't think we want this. Sample dims should either be an input and passed downstream to plot_dist or if having flattened to `sample` is not to be supported here then we should hardcode it and check both `chain` and `draw` are present.

src/arviz_plots/plots/energyplot.py
```diff
@@ -0,0 +1,138 @@
...
+        plot_collection=plot_collection,
+        backend=backend,
+        labeller=labeller,
+        aes_map=aes_map,
+        plot_kwargs=plot_kwargs,
+        stats_kwargs=stats_kwargs,
+        pc_kwargs=pc_kwargs,
+    )
+
+    plot_collection.add_legend("energy", loc="outside right upper")
```
OriolAbril, line 125: I opened an issue for this, it won't work across backends yet: https://github.com/arviz-devs/arviz-plots/issues/109

src/arviz_plots/plots/energyplot.py
```diff
@@ -0,0 +1,138 @@
...
+
+    if bfmi:
+        raise NotImplementedError("BFMI is not implemented yet")
+
+    return plot_collection
+
+
+def _get_energy_ds(dt):
```
OriolAbril, line 133: I'll try to look at https://docs.xarray.dev/en/stable/generated/xarray.DataArray.diff.html and see if we can make all this a bit more direct. I think it will also break when getting transposed inputs

aloctavodia: I have to admit, that I spent more than I expected to write this part (and try alternatives) :-)

src/arviz_plots/plots/energyplot.py
aloctavodia: I added a temporary patch until we have a proer solution.

Issue #109, opened 2024-11-19
Legend support


OriolAbril:
At the time of writing, we have some legend support for bokeh, relatively good legend support for matplotlib and no legend support for plotly.

We should explore better APIs and extend legend support. At a minimum, `pc.add_legend("dim_name")` should work on all backends and generate acceptable legends.

Once `add_legend` is working better, it should be called automatically for several plots, including (but might not be limited to):

* plot_psense_dist -> for alpha
* plot_energy
* plot_xyz when model is mapped to color

Advaitgaur004:
Added a legend for Plotly and it is working fine. I will plot `psense_dict` for alpha. Plot energy and `plot_anything` mapped to color is yet to be completed if the initial implementation is correct. @OriolAbril

OriolAbril:
Basic line legends should already work for all backends. We should activate it for all plots that need a default legend and then open sub-issues of this one with all the issues we encounter while doing that. I have already started with bokeh legend placing

PR #115, opened 2025-01-22
Add consistent legend support across backends

Advaitgaur004:
## Description
This PR implements consistent legend support across all plotting backends (matplotlib, bokeh, and plotly) and adds automatic legend generation for plot_psense_dist.

### Key Changes
- Added full legend support for Plotly backend
- Enhanced Bokeh legend with automatic placement logic
- Added automatic legend generation for:
  - `plot_psense_dist` (for alpha parameter)

### Implementation Details
- Implemented `legend()` function for Plotly backend
- Enhanced Bokeh's legend placement with auto-positioning
- Added automatic legend generation in `plot_psense_dist` for power scaling factor
- Ensured consistent behavior of `pc.add_legend("dim_name")` across all backends

## Checklist
- [x] Added/updated tests
- [x] Updated documentation
- [ ] Added example notebooks (if applicable)
- [ ] Passes all CI checks

## Issue Targeted 
#109 

----
📚 Documentation preview 📚: https://arviz-plots--115.org.readthedocs.build/en/115/

### DIFF
--- a/src/arviz_plots/backend/bokeh/legend.py
+++ b/src/arviz_plots/backend/bokeh/legend.py
@@ -20,7 +20,7 @@ def legend(
     artist_kwargs=None,
     legend_target=None,
     side="auto",
-    legend_placement_threshold=600, #Magic number
+    legend_placement_threshold=600,  # Magic number
     **kwargs,
 ):
     """Generate a legend on a figure given lists of labels and property kwargs.
@@ -69,7 +69,7 @@ def legend(
             side = "right"
         else:
             side = "center"
-    
+
     leg = Legend(
         items=[(str(label), [glyph]) for label, glyph in zip(label_list, glyph_list)],
         title=title,
--- a/src/arviz_plots/backend/bokeh/legend.py
+++ b/src/arviz_plots/backend/bokeh/legend.py
@@ -19,7 +19,8 @@ def legend(
     artist_type="line",
     artist_kwargs=None,
     legend_target=None,
-    side="right",
+    side="auto",
+    legend_placement_threshold=600, #Magic number
     **kwargs,
 ):
     """Generate a legend on a figure given lists of labels and property kwargs.
@@ -62,11 +63,9 @@ def legend(
         glyph = artist_fun(**{**artist_kwargs, **kws})
         glyph_list.append(glyph)
 
-    # Add automatic legend placement logic
     if side == "auto":
-        # Place legend outside if there's enough space, inside otherwise
         plot_width = target_plot.width
-        if plot_width >= 600:  # arbitrary threshold
+        if plot_width >= legend_placement_threshold:
             side = "right"
         else:
             side = "center"
--- a/src/arviz_plots/backend/bokeh/legend.py
+++ b/src/arviz_plots/backend/bokeh/legend.py
@@ -61,6 +61,16 @@ def legend(
     for kws in kwarg_list:
         glyph = artist_fun(**{**artist_kwargs, **kws})
         glyph_list.append(glyph)
+
+    # Add automatic legend placement logic
+    if side == "auto":
+        # Place legend outside if there's enough space, inside otherwise
+        plot_width = target_plot.width
+        if plot_width >= 600:  # arbitrary threshold
+            side = "right"
+        else:
+            side = "center"
+    
     leg = Legend(
         items=[(str(label), [glyph]) for label, glyph in zip(label_list, glyph_list)],
         title=title,
--- a/src/arviz_plots/backend/plotly/legend.py
+++ b/src/arviz_plots/backend/plotly/legend.py
@@ -1,15 +1,17 @@
 """Plotly legend generation."""
 
+
 def dealiase_line_kwargs(kwargs):
     """Convert arviz common interface properties to plotly ones."""
     prop_map = {"linewidth": "width", "linestyle": "dash"}
     return {prop_map.get(key, key): value for key, value in kwargs.items()}
 
+
 def legend(
     target, kwarg_list, label_list, title=None, artist_type="line", artist_kwargs=None, **kwargs
 ):
     """Generate a legend with plotly.
-    
+
     Parameters
     ----------
     target : plotly.graph_objects.Figure
@@ -34,7 +36,7 @@ def legend(
     """
     if artist_kwargs is None:
         artist_kwargs = {}
-    
+
     if artist_type == "line":
         artist_fun = target.add_scatter
         kwarg_list = [dealiase_line_kwargs(kws) for kws in kwarg_list]
@@ -50,11 +52,7 @@ def legend(
             mode=mode,
             line=kws,
             showlegend=True,
-            **artist_kwargs
+            **artist_kwargs,
         )
 
-    target.update_layout(
-        showlegend=True,
-        legend_title_text=title,
-        **kwargs
-    )
+    target.update_layout(showlegend=True, legend_title_text=title, **kwargs)
--- a/src/arviz_plots/backend/plotly/legend.py
+++ b/src/arviz_plots/backend/plotly/legend.py
@@ -2,7 +2,7 @@
 
 def dealiase_line_kwargs(kwargs):
     """Convert arviz common interface properties to plotly ones."""
-    prop_map = {"width": "width", "linestyle": "dash"}
+    prop_map = {"linewidth": "width", "linestyle": "dash"}
     return {prop_map.get(key, key): value for key, value in kwargs.items()}
 
 def legend(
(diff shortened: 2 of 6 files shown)


src/arviz_plots/backend/bokeh/legend.py
```diff
@@ -61,6 +61,16 @@ def legend(
     for kws in kwarg_list:
         glyph = artist_fun(**{**artist_kwargs, **kws})
         glyph_list.append(glyph)
+
```
Advaitgaur004, line 64: Add logic for automatic legend placement: to the right if the total plot size is greater than 600; otherwise, it should be centered. The placement is adjustable based on the use case.

src/arviz_plots/backend/plotly/legend.py
```diff
@@ -1,8 +1,51 @@
...
+    """Convert arviz common interface properties to plotly ones."""
+    prop_map = {"width": "width", "linestyle": "dash"}
+    return {prop_map.get(key, key): value for key, value in kwargs.items()}
 
 def legend(
     target, kwarg_list, label_list, title=None, artist_type="line", artist_kwargs=None, **kwargs
 ):
-    """Generate a legend with plotly."""
-    raise NotImplementedError("Still in progress")
+    """Generate a legend with plotly.
```
Advaitgaur004, line 11: implement the Plotly legend support

src/arviz_plots/plots/psensedistplot.py
```diff
@@ -249,4 +249,10 @@ def plot_psense_dist(
         stats_kwargs=stats_kwargs,
     )
 
+    # Add legend for alpha parameter automatically
```
Advaitgaur004, line 252: Automatically add the legend for the alpha parameter without manual effort.

src/arviz_plots/backend/plotly/legend.py
```diff
@@ -1,8 +1,51 @@
...
+    Parameters
+    ----------
+    target : plotly.graph_objects.Figure
+        The figure to add the legend to
+    label_list : list
+        List of labels for each legend entry
+    artist_kwargs : dict, optional
+        Additional kwargs passed to all artists
+    **kwargs : dict
+        Additional kwargs passed to legend configuration
```
amaloney, line 22: thank you for writing a docstring, it is greatly appreciated. Unfortunately the docstring does not match the function signature, which you will need to update.

src/arviz_plots/backend/plotly/legend.py
```diff
@@ -1,8 +1,51 @@
...
+    """
+    if artist_kwargs is None:
+        artist_kwargs = {}
+    
+    if artist_type == "line":
+        artist_fun = target.add_scatter
+        kwarg_list = [dealiase_line_kwargs(kws) for kws in kwarg_list]
```
amaloney, line 29: the `kwarg_list` is an input argument, yet it is being overwritten here. Which do you want to use, the user supplied one, or the one being created here?

src/arviz_plots/backend/plotly/legend.py
```diff
@@ -1,8 +1,51 @@
...
+            showlegend=True,
+            **artist_kwargs
+        )
+
+    # Configure legend
+    target.update_layout(
+        showlegend=True,
+        legend_title_text=title,
+        **kwargs
+    )
```
amaloney, line 51: There is no return object, and you should mention that in the docstring

src/arviz_plots/plots/psensedistplot.py
amaloney: In general, you do not need to add comments like this to your PR. The reviewer will know what has been changed and will be able to understand what you are doing...hopefully. It's added noise when reviewing, because I have to resolve your comments in the GH interface.

amaloney (changes requested):
Add more context in the linking issue, and why these features are needed. Also add visuals that explicitly show how this new code is adding to the user experience.

src/arviz_plots/backend/plotly/legend.py
Advaitgaur004: The one that is created at line number 29 is a standard implementation. 
Reason: if I send 
**kwarg_list = [
    {"color": "red", "width": 3, "linestyle": "--"},
    {"color": "blue", "width": 2, "linestyle": "-."}
]**
as kwarg_list in the legend function and run without using dealiase_line_kwargs, it shows 
`AttributeError: Line2D.set() got an unexpected keyword argument 'width'`
The dealiase_line_kwargs will change the name from `width` to `linewidth,` which is recognizable by matplotlib's Line2D.

If the parameter is passed correctly, then this function `dealiase_line_kwargs`  is unnecessary. In that case, I need to change the legend file created by @OriolAbril in matplotlib. 
This is for safety in case someone mistakenly writes "width" instead of "linewidth." It's better to keep this here; I hope that addresses your curiosity.

src/arviz_plots/backend/plotly/legend.py
```diff
@@ -1,8 +1,51 @@
 """Plotly legend generation."""
 
+def dealiase_line_kwargs(kwargs):
+    """Convert arviz common interface properties to plotly ones."""
+    prop_map = {"width": "width", "linestyle": "dash"}
+    return {prop_map.get(key, key): value for key, value in kwargs.items()}
```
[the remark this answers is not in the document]
Advaitgaur004: I think we should keep it there. If, by mistake, someone sends "width" as an argument instead of "lineswidth," it should handle the issue. For example:
```
kwargs_list= [
    {"color": "red", "width": 3, "linestyle": "--"},
    {"color": "blue", "width": 2, "linestyle": "-."}
]
```
If they use "lineswidth," then we are good again; it's just for a better user experience.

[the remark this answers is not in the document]
Advaitgaur004: I found this function to be intuitive. The reason is that in some plotting libraries, one may need to use "linewidth," while in others, "width" is required. As a user or developer, it can be challenging to remember which argument is valid for which library. This function elegantly addresses that issue. It's also worth noting that it's not the same as in `backend.bokeh.legend`; if you look closely, only "line_width" is a valid argument there.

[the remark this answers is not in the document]
Advaitgaur004: The moral of the story is that it is best to create a standard utility module. Different plotting backends require different naming conventions, so we should consistently use `width` (which is valid in Plotly but not in Matplotlib or Bokeh). Additionally, any other arguments, such as 'linewidth', 'line_width', or similar variations, should be handled gracefully.

Any Thought @amaloney

src/arviz_plots/backend/bokeh/legend.py
```diff
@@ -61,6 +61,16 @@ def legend(
...
+    # Add automatic legend placement logic
+    if side == "auto":
+        # Place legend outside if there's enough space, inside otherwise
+        plot_width = target_plot.width
+        if plot_width >= 600:  # arbitrary threshold
```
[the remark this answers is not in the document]
Advaitgaur004: Okay I have linked MRE For better user experience

src/arviz_plots/backend/plotly/legend.py
Advaitgaur004: Done

Advaitgaur004:
I have added a legend placement threshold in the Plotly backend legend, which can now be altered, eliminating the need for hard coding.
And for MRE that links to `bokeh.backend.legend` 
```
if side == "auto":
        plot_width = target_plot.width
        if plot_width >= legend_placement_threshold:
            side = "right"
        else:
            side = "center"
```

can be reproduced by following the code.

```python
import numpy as np
from bokeh.plotting import figure, show
from bokeh.layouts import gridplot

def create_example_plot(width):
    p = figure(width=width, height=300)
    x = np.linspace(0, 10, 100)
    grid = gridplot([[p]], merge_tools=False)
    p.line(x, np.sin(x), line_color="red")
    p.line(x, np.cos(x), line_color="blue")
    legend(grid, 
           [{'color': 'red'}, {'color': 'blue'}], 
           ['sin(x)', 'cos(x)'], 
           side='auto')
    
    return grid
narrow_plot = create_example_plot(300)
wide_plot = create_example_plot(800)
show(gridplot([[narrow_plot, wide_plot]]))
```

300x300

300x800

Advaitgaur004:
Any Thoughts on aliasing in various backend plotting modules, one has to make thing easier for developer too @aloctavodia

Advaitgaur004:
gentle reminder @amaloney

amaloney:
As a user I would expect the following.

- If I request an image size of 300x300 pixels then I should get an image of 300x300 pixels.

I get this with the first example

image

but I do not get this with the second example.

image

The second plot has a total dimension of 300x800 pixels where a portion of that image is where the plot exists plus the size of the legend. This is not what I would have expected the output to be, since I requested an plot of 300x800 pixels in size.

There unfortunately is no easy solution to this.

We could instead have an argument where the user can request the legend be outside the plot area. Then in the docstring of the method we indicate that if this argument is selected, the resulting plot area will be smaller in size to accomodate the legend, given the requested figure size. How about you try implementing this idea.

Advaitgaur004:
First of all, there is a slight error; the dimensions should be 800x300, as the height is predefined to be 300 in my test case. Therefore, it is not 300x800 but rather 800x300.

If the `legend_in_place` argument is set to False (indicating that the user wants the legend to be outside the plot), I will reduce the original plot width to 80% and adjust the legend to occupy the remaining 20%. Should I go this way or should we automate this thing, if the width is too large then place the legend inside the plot(reasonable) and if small then legend should be outside

Advaitgaur004:
@amaloney Review

OriolAbril (approved):
This is a great starting point, thanks!

I have rebased it and taken care of tests and docs so it can be merged, sorry we haven't been able to do proper continued reviews at a reasonable pace so far.

Advaitgaur004:
> This is a great starting point, thanks!
> 
> I have rebased it and taken care of tests and docs so it can be merged, sorry we haven't been able to do proper continued reviews at a reasonable pace so far.

No worries sénior 👍

PR #122, opened 2025-02-05
add arviz-style for bokeh

aloctavodia:
We offer 3 custom styles for Matplotlib and Plotly, but not for Bokeh. This is my hacky attempt to fix that. @amaloney, maybe you can suggest a better approach.  Also, the style itself needs further work to make it closer to the one for matplotib and plotly. 

----
📚 Documentation preview 📚: https://arviz-plots--122.org.readthedocs.build/en/122/


amaloney:
quick glance, but I think we can use the `yaml` files as the records of truth, even for a default style

aloctavodia:
Yes, we have a yaml file, we should at least two more. I added an attribute/property that as far as I understand is not part of the standard bokeh configuration "Cycler: colors :" I use this to provide the color palette for each style (see https://github.com/arviz-devs/arviz-plots/pull/122/files#diff-50c474257b57636a34efff7b446a6669a00a2501e47b2a0d23d8dd752f54577eR39).

Issue #96, opened 2024-10-17
Add step type to histograms


aloctavodia:
Stepped histograms, instead of bars histograms, make it slightly easier to overlap and compare multiple distributions.

OriolAbril:
I agree we should have this option. Not really sure about defaults though, personally I find easier to read overlaid bar histograms with transparency. Easier which is not easy though, both break multiple gestalt principles of visual perception and require a lot of concentration to see the continuity of each distribution.

The-Broken-Keyboard:
Heyy

If this issue is still open then I would like to work on this.

Please assign me this issue 

Thank you

aloctavodia:
assigned

The-Broken-Keyboard:
heyy @aloctavodia 

I was implementing the step histogram for various backends and I have reached the conclusion that step histogram functionality is only provided by matplotlib and neither bokeh nor plotly provide any functionality like step histograms. 

Do you want me to implement for matplotlib only and leave it for other two backends?

aloctavodia:
It should be implemented for all or none of the backends. There is no direct option but you can do something like this https://docs.bokeh.org/en/latest/docs/user_guide/basic/lines.html#step-lines and then add two lines.

For plotly not sure. What about this https://stackoverflow.com/questions/64453989/plotly-how-to-plot-histogram-in-root-style-showing-only-the-contours-of-the-his

The-Broken-Keyboard:
Okay, these workarounds look feasible.

Will do these for bokeh and plotly

The-Broken-Keyboard:
@aloctavodia I think we can close this now, since the functionality has been implemented now

PR #147, opened 2025-02-21
added functionality of step histogram for all three backends

The-Broken-Keyboard:
This implements the new functionality of step histograms discussed in #96 

- To implement step_histogram functionality, it uses `step` function available in matplotlib backend 
- In Bokeh it uses `step` function available in bokeh backend
- In plotly there is no direct function available to achieve this goal, so it uses `go.scatter` functionality to implement step histogram functionality

**step_hist** flag parameter has been added at appropriate places, with default value as False, to maintain compatiblity with old version of functionalities related  to histogram. This new parameter enables users to switch to step histograms just by passing value True to it.

Here are examples using plot_dist function in arviz_plots with flag step_hist as True :

Backend: Plotly
image

Backend: Matplotlib
image

Backend: Bokeh
image

### DIFF
--- a/src/arviz_plots/backend/bokeh/__init__.py
+++ b/src/arviz_plots/backend/bokeh/__init__.py
@@ -275,14 +275,37 @@ def hist(
     facecolor=unset,
     edgecolor=unset,
     alpha=unset,
+    step_hist=False,
+    step_mode="center",
     **artist_kws,
 ):
-    """Interface to Bokeh for a histogram bar plot."""
+    """Interface to Bokeh for a histogram bar or step plot."""
     if color is not unset:
         if facecolor is unset:
             facecolor = color
         if edgecolor is unset:
             edgecolor = color
+
+    # Ensure alpha and colors have default values
+    alpha = 1.0 if alpha is unset else alpha
+    edgecolor = "black" if edgecolor is unset else edgecolor
+
+    if step_hist:
+        x = [l_e[0], l_e[0]]
+        y_step = [0, y[0]]
+        for i, y_i in enumerate(y):
+            x.append(r_e[i])
+            y_step.append(y_i)
+        x.append(r_e[-1])
+        y_step.append(bottom)
+
+        p = target.step(x, y_step, line_color=edgecolor, alpha=alpha, mode=step_mode, **artist_kws)
+
+        target.x_range = Range1d(float(l_e[0]), float(r_e[-1]))
+        target.y_range = Range1d(float(bottom), float(max(y)) * 1.2)  # Add padding to y-axis
+
+        return p
+
     kwargs = {"bottom": bottom, "fill_color": facecolor, "line_color": edgecolor, "alpha": alpha}
     return target.quad(top=y, left=l_e, right=r_e, **_filter_kwargs(kwargs, artist_kws))
 
--- a/src/arviz_plots/backend/matplotlib/__init__.py
+++ b/src/arviz_plots/backend/matplotlib/__init__.py
@@ -211,6 +211,7 @@ def hist(
     alpha=unset,
     facecolor=unset,
     edgecolor=unset,
+    step_hist=False,
     **artist_kws,
 ):
     """Interface to matplotlib for a histogram bar plot."""
@@ -225,9 +226,23 @@ def hist(
             facecolor = color
         if edgecolor is unset:
             edgecolor = color
-    kwargs = {"bottom": bottom, "color": facecolor, "edgecolor": edgecolor, "alpha": alpha}
+    kwargs = {"color": facecolor, "alpha": alpha}
+    if step_hist:
+        return target.step(
+            np.r_[l_e, r_e[-1]],
+            np.r_[y, y[-1]],
+            where="post",
+            **_filter_kwargs(kwargs, None, artist_kws),
+        )
+
     return target.bar(
-        l_e, height, width=widths, align="edge", **_filter_kwargs(kwargs, None, artist_kws)
+        l_e,
+        height,
+        width=widths,
+        align="edge",
+        bottom=bottom,
+        edgecolor=edgecolor,
+        **_filter_kwargs(kwargs, None, artist_kws),
     )
 
 
--- a/src/arviz_plots/backend/plotly/__init__.py
+++ b/src/arviz_plots/backend/plotly/__init__.py
@@ -4,6 +4,7 @@ Notes
 -----
 :term:`artists` are returned, but it seems modifying them won't modify the figure.
 """
+
 import re
 import warnings
 
@@ -286,9 +287,10 @@ def hist(
     alpha=unset,
     facecolor=unset,
     edgecolor=unset,
+    step_hist=False,
     **artist_kws,
 ):
-    """Interface to matplotlib for a histogram bar plot."""
+    """Interface to Plotly for a histogram plot."""
     artist_kws.setdefault("showlegend", False)
     widths = np.asarray(r_e) - np.asarray(l_e)
     if np.any(bottom != 0):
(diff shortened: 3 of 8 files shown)


src/arviz_plots/backend/bokeh/__init__.py
```diff
@@ -275,14 +275,37 @@ def hist(
...
+    """Interface to Bokeh for a histogram bar or step plot."""
     if color is not unset:
         if facecolor is unset:
             facecolor = color
         if edgecolor is unset:
             edgecolor = color
+
+    # Ensure alpha and colors have default values
+    alpha = 1.0 if alpha is unset else alpha
+    edgecolor = "black" if edgecolor is unset else edgecolor
```
aloctavodia, line 291:
```suggestion
kwargs = {"bottom": bottom, "fill_color": facecolor, "line_color": edgecolor, "alpha": alpha}
```

src/arviz_plots/backend/matplotlib/__init__.py
```diff
@@ -225,9 +226,23 @@ def hist(
...
+
     return target.bar(
-        l_e, height, width=widths, align="edge", **_filter_kwargs(kwargs, None, artist_kws)
+        l_e,
+        height,
+        width=widths,
+        align="edge",
+        bottom=bottom,
+        edgecolor=edgecolor,
+        **_filter_kwargs(kwargs, None, artist_kws),
```
aloctavodia, line 245:
```suggestion
 l_e, height, width=widths, align="edge", **_filter_kwargs(kwargs, None, artist_kws)
```

not clear why are you changing

src/arviz_plots/backend/plotly/__init__.py
```diff
@@ -300,19 +302,48 @@ def hist(
...
-    hist_object = go.Bar(
-        x=(l_e + r_e) / 2,
-        y=height,
-        base=bottom,
-        width=widths,
-        marker=marker_kwargs,
-        **_filter_kwargs({"opacity": alpha}, artist_kws),
-    )
+
+    if step_hist is False:
```
aloctavodia, line 306: if `step_hist` is just True or False ask "if step_hist"

src/arviz_plots/plot_collection.py
```diff
@@ -935,6 +935,7 @@ def map(
         subset_info=False,
         store_artist=True,
         artist_dims=None,
+        step_hist=False,
```
aloctavodia, line 938: map is a generic function it should not have arguments specific arguments of specific visuals

src/arviz_plots/plot_collection.py
```diff
@@ -974,6 +975,10 @@ def map(
         artist_dims : mapping of {hashable : int}, optional
             Dictionary of sizes for proper allocation and storage when using
             ``map`` with functions that return an array of :term:`artist`.
+        step_hist : boolean, default False
+            Flag to indicate that the `hist` function should be called with
+            the keyword argument `step_hist=True` if step histogram is required
+            instead of bar histogram.
```
aloctavodia, line 981:
```suggestion
```

src/arviz_plots/plots/distplot.py
```diff
@@ -39,6 +39,7 @@ def plot_dist(
     aes_map=None,
     plot_kwargs=None,
     stats_kwargs=None,
+    step_hist=False,
```
aloctavodia, line 42: This should not be here.

src/arviz_plots/plots/distplot.py
```diff
@@ -114,6 +115,11 @@ def plot_dist(
         * credible_interval -> passed to eti or hdi
         * point_estimate -> passed to mean, median or mode
 
+    step_hist : boolean, default False
+            Flag to indicate that the `hist` function should be called with
+            the keyword argument `step_hist=True` if step histogram is required
+            instead of bar histogram.
+
```
aloctavodia, line 122:
```suggestion
```

src/arviz_plots/plots/distplot.py
```diff
@@ -281,7 +287,12 @@ def plot_dist(
             )
 
             plot_collection.map(
-                hist, "hist", data=density, ignore_aes=density_ignore, **density_kwargs
+                hist,
+                "hist",
+                data=density,
+                ignore_aes=density_ignore,
+                step_hist=step_hist,
+                **density_kwargs,
```
aloctavodia, line 295:
```suggestion
                **density_kwargs,
```

step_hist defaults to "False" if the user wants a step histogram, then it should passed to `density`

src/arviz_plots/plots/ppcdistplot.py
```diff
@@ -28,6 +28,7 @@ def plot_ppc_dist(
     aes_map=None,
     plot_kwargs=None,
     stats_kwargs=None,
+    step_hist=False,
```
aloctavodia, line 31:
```suggestion
```

src/arviz_plots/plots/ppcdistplot.py
```diff
@@ -91,6 +92,11 @@ def plot_ppc_dist(
 
         * density -> passed to kde, ecdf, ...
 
+    step_hist : boolean, default False
+            Flag to indicate that the `hist` function should be called with
+            the keyword argument `step_hist=True` if step histogram is required
+            instead of bar histogram.
+
```
aloctavodia, line 99:
```suggestion
```

src/arviz_plots/plots/ppcdistplot.py
```diff
@@ -285,6 +291,7 @@ def plot_ppc_dist(
                 "observe_density",
                 data=dt_observed,
                 ignore_aes=observed_ignore,
+                step_hist=step_hist,
```
aloctavodia, line 294:
```suggestion
```

src/arviz_plots/visuals/__init__.py
```diff
@@ -13,7 +13,7 @@
 from arviz_base.labels import BaseLabeller
 
 
-def hist(da, target, backend, **kwargs):
+def hist(da, target, backend, step_hist=False, **kwargs):
```
aloctavodia, line 16:
```suggestion
def hist(da, target, backend,, **kwargs):
```

src/arviz_plots/visuals/__init__.py
```diff
@@ -24,6 +24,7 @@ def hist(da, target, backend, **kwargs):
         da.sel(plot_axis="left_edges"),
         da.sel(plot_axis="right_edges"),
         target,
+        step_hist=step_hist,
```
aloctavodia, line 27:
```suggestion
```

aloctavodia (changes requested):

src/arviz_plots/backend/matplotlib/__init__.py
The-Broken-Keyboard: @aloctavodia 

Actually the bottom and edgecolor attributes are not there in step histogram, so if we feel kwargs (containing bottom and edgecolor values) in to step function then it throws error. 
So I removed bottom and edgecolor attributes from kwargs dictionary and I did feed these attributes separately into bar histogram function.

I can create two separate kwargs too for step and bar histograms. Let me know if you want that

aloctavodia: We should then ignore them

src/arviz_plots/plots/distplot.py
The-Broken-Keyboard: heyy @aloctavodia ,

so do you want step_hist key with True value to be passed in plot_kwargs instead of passing step_hist value directly?

The-Broken-Keyboard: image
like this?

The-Broken-Keyboard: Or do you want me to pass step_hist in parameter kind, instead of sending it in plot_kwargs? This is also a possibility and it looks more sensible

aloctavodia: Currently, if we want to change the color of a histogram we do `plot_kwargs={"hist": {"color": "gray"}`, so if we would like to change the "style" of the histogram we should do `plot_kwargs={"hist": {"step": True}`. If instead we want to change the way we compute the histogram (or kde, ecdf...) we pass  `stats_kwargs={"density": {"bins": 100}});` see https://arviz-plots.readthedocs.io/en/latest/tutorials/plots_intro.html#plot-kwargs and https://arviz-plots.readthedocs.io/en/latest/tutorials/plots_intro.html#stats-kwargs

The-Broken-Keyboard:
@aloctavodia all the requested changes have been addressed. Please review.

aloctavodia (commented):
I pushed a few changes. Please make sure everything works as expected.

The-Broken-Keyboard:
I checked it and everything is working as expected. There were few modifications required to adjust cleaned changes. I have done that.

@aloctavodia

aloctavodia:
great! Can you add a test?

The-Broken-Keyboard:
Should we write a test for step histogram itself, or should we write test for the plots which use kind="hist" ( with step: True in plot_kwargs  ) ? @aloctavodia

aloctavodia:
the second one

The-Broken-Keyboard:
@aloctavodia I have added the tests for dist_plot ( since that was only using "hist" kind )

Also the docs build is failing on 'pavacalibrationplot.py' file i guess, which i have not touched. Can you check that please?

Also review the tests and let me know if more changes are required.

aloctavodia:
I made changes to ppcdistplot that are not here, but they should.

The-Broken-Keyboard:
@aloctavodia fixed that. Please review the tests which I added and let me know if anything else needs to be done.

aloctavodia (approved):

aloctavodia:
thanks @The-Broken-Keyboard!

PR #158, opened 2025-03-01
Added plot_bf() function for bayes_factor in arviz-plots

PiyushPanwarFST:
Implemented the `plot_bf()` function in `arviz-plots`. 
Added a test case to validate the functionality of `plot_bf()`.
Included `plot_bf()` in the documentation.
Incorporated feedback received so far.
Resolved all the linter(pylint) errors.
Screenshot 2025-03-01 at 2 31 11 PM

----
📚 Documentation preview 📚: https://arviz-plots--158.org.readthedocs.build/en/158/


PiyushPanwarFST:
@aloctavodia, I have created a new PR with all your suggested changes. Due to certain constraints, I had to close the previous PR. Please review the new one at your convenience and let me know if any further improvements are needed. Thank you

src/arviz_plots/plots/bfplot.py
```diff
@@ -0,0 +1,218 @@
...
+        ci_prob=None,
+        plot_collection=plot_collection,
+        backend=backend,
+        labeller=labeller,
+        plot_kwargs=plot_kwargs,
+        stats_kwargs=stats_kwargs,
+        pc_kwargs=pc_kwargs,
+    )
+
+    if ref_val == 0:
```
aloctavodia, line 182: `ref_val` could take any real value.

aloctavodia, line 182:
```suggestion
    if ref_val is not False:
```

src/arviz_plots/plots/bfplot.py
```diff
@@ -0,0 +1,218 @@
...
+            plot_collection, aes_map, "bf_text", sample_dims
+        )
+
+    bf_dt = xr.Dataset({var_name: xr.DataArray([0])})
+    plot_collection.map(
+        annotate_xy,
+        "bf_text",
+        text=f"BF10: {bf['BF10']:.2f}\nBF01: {bf['BF01']:.2f}",
+        x=ref_val + 13,
+        y=max(ref_vals["prior"], ref_vals["posterior"]) * 1.7,
```
aloctavodia, line 209: Not sure these magic numbers will generally work. See for instance https://arviz-plots--158.org.readthedocs.build/en/158/gallery/plot_bf.html for plotly and bokeh backends

PiyushPanwarFST: Hey, As the annotation was coming in between the line while plotting. I added this value to adjust the position of the annotation.
Can you suggest if any better way exists?

aloctavodia: I understand that, but as you can see, this solution does not work for the example.  
Other solution could be: 
* Find a way to position the text using a relative frame-reference
* plot it at max(value_at_posterior, value_at_prior), but just one number (the BF10 or BF01), this is the same as before but because we have just one number it could be less crowed. Probably this is the best not only in this case but in general 
* Use the legend
* Use the title.

PiyushPanwarFST:
Screenshot 2025-03-07 at 6 02 50 PM

Heyy @aloctavodia,  I have added a new func called `add_text_legend` in `plot_collection.py` to add the annotation values of bayes_factor. This function will help us to add text-based legends for other plots also. I think there might be better way exist but this one I found satisfactory. Let me know your thoughts, always ready to improve.

aloctavodia:
Most of the code is duplicated from add legend. A better approach may be to add an  argument to the existing function instead of duplicating it. Also better to reduce. 

Also a simpler approach will be to use anotate text, and just reduce the amount of information, just adding the value of one of the bayes factors should be enough if the docs explain wich value is the one plotted

PiyushPanwarFST:
Screenshot 2025-02-26 at 12 13 05 PM

The main issue with the annotated text approach is that the BF info appears in the middle of the plot. Since the legend approach works only with Matplotlib, I can:

1. Use an if statement to apply annotation only when using Matplotlib, avoiding issues in Bokeh or Plotly.
2. Otherwise, I can go ahead with adding another argument to the existing legend function.

PiyushPanwarFST:
Screenshot 2025-03-08 at 7 25 15 PM

Hey, as per your suggestion, I have added a new argument to the existing add_legend() function to adjust the Bayes factor values. Let me know if any further modifications are needed.

docs/source/gallery/inference_diagnostics/plot_bf.py
```diff
@@ -0,0 +1,23 @@
+"""
```
aloctavodia, line 1: Bayes factors are not a method for inference diagnostics. They are a tool for model comparison.

src/arviz_plots/plots/bfplot.py
```diff
@@ -0,0 +1,210 @@
+"""Contain functions for Bayesian Factor plotting."""
```
aloctavodia, line 1:
```suggestion
"""Contain functions for Bayes Factor plotting."""
```

docs/source/gallery/inference_diagnostics/plot_bf.py
aloctavodia: Please fix this

PR #182, opened 2025-03-17
plot_converge_dist: Add grouped argument

aloctavodia:
All variables together

```python
azp.plot_convergence_dist(data)
```
all_grouped

Only `"za_county"` and `"a"` variables

```python
azp.plot_convergence_dist(data),  var_names=["za_county", "a"])
```
some_grouped

Only `"za_county"` and `"a"` variables, not together

```python
azp.plot_convergence_dist(data),  var_names=["za_county", "a"], grouped=False)
```
some_separated

This will fail for `kind="ecdf"` but work for the rest. The reason is that `"g"` has a different shape than (`"za_county"` or `"a"`). This should be fixed in arviz-stats.

```python
azp.plot_convergence_dist(data),  kind="hist", var_names=["za_county", "a", "g"], grouped=False)
```
some_separated_dim

----
📚 Documentation preview 📚: https://arviz-plots--182.org.readthedocs.build/en/182/

### DIFF
--- a/src/arviz_plots/plots/convergencedistplot.py
+++ b/src/arviz_plots/plots/convergencedistplot.py
@@ -15,6 +15,7 @@ from arviz_plots.visuals import vline
 def plot_convergence_dist(
     dt,
     diagnostics=None,
+    grouped=True,
     ref_line=True,
     var_names=None,
     filter_vars=None,
@@ -35,6 +36,10 @@ def plot_convergence_dist(
 ):
     """Plot the distribution of convergence diagnostics (ESS and/or R-hat).
 
+    By default all variables are grouped togehter and one plot per diagnostic is created.
+    If you are interested in representing individual (multidimnesional variables) pass them
+    in `var_names`. Additionally, you will need to specify the
+
     Information on how the diagnostics are computed can be found in [1]_.
 
     Parameters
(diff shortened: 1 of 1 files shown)


src/arviz_plots/plots/convergencedistplot.py
```diff
@@ -35,6 +36,10 @@ def plot_convergence_dist(
 ):
     """Plot the distribution of convergence diagnostics (ESS and/or R-hat).
 
+    By default all variables are grouped togehter and one plot per diagnostic is created.
+    If you are interested in representing individual (multidimnesional variables) pass them
+    in `var_names`. Additionally, you will need to specify the
```
AlexAndorra, line 41: End of sentence is missing, isn't it?

src/arviz_plots/plots/convergencedistplot.py
```diff
@@ -128,6 +137,17 @@ def plot_convergence_dist(
...
 
+    Select two variable and plot them separately
+
+    .. plot::
+        :context: close-figs
+
+        >>> plot_convergence_dist(
+        >>>     radon,
+        >>>     var_names=["za_county", "a"],
+        >>>     group_by_dimensions=["County"]
```
AlexAndorra, line 148: Isn't it `grouped` ?

AlexAndorra (changes requested):
Great addition @aloctavodia ! Left two comments

aloctavodia:
Thanks @AlexAndorra. Fixed

Issue #156, opened 2025-02-27
Add square-root scale for plotly and bokeh


aloctavodia:
For the rootogram we need to transform the y-axis to the square-root scale. This is already implemented for matplotlib backend but not for the rest.

The-Broken-Keyboard:
@aloctavodia assign it to me, I will do it

The-Broken-Keyboard:
@aloctavodia to implement square root scale in Bokeh, it requires to add a typescript or javascript file because that's the only possible way to add custom scale in Bokeh. Is it fine to add typescript file for scaling in Bokeh backend folder.

aloctavodia:
Not sure, probably OK.

The-Broken-Keyboard:
@aloctavodia we can close this i guess, since it is already complete now.

PR #183, opened 2025-03-18
adds sqrt scale for yaxis in plotly

### LINKED ISSUE #179: Add square-root scale for plotly

The-Broken-Keyboard:
Closes #179 and #156

Plotly does not natively support custom scales, so implementing a square root (√y) scale requires a transformation of both the plot data and the y-axis. The `apply_square_root_scale` function achieves this by iterating through all traces in the chart, applying the square root transformation to the y-values, and updating the y-axis tick values and labels accordingly. This ensures that the visual representation maintains interpretability while reflecting the transformed scale.

Example:

**Linear scale plot:**

image

**Square root scale plot:**

image

----
📚 Documentation preview 📚: https://arviz-plots--183.org.readthedocs.build/en/183/

### DIFF
--- a/src/arviz_plots/plots/ppcrootogramplot.py
+++ b/src/arviz_plots/plots/ppcrootogramplot.py
@@ -314,7 +314,7 @@ def plot_ppc_rootogram(
             **title_kwargs,
         )
 
-    if backend in ("matplotlib", "bokeh"):
+    if backend in ("matplotlib", "bokeh", "plotly"):
         plot_collection.map(
             set_y_scale,
             store_artist=False,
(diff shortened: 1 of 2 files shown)


src/arviz_plots/plots/ppcrootogramplot.py
```diff
@@ -314,7 +314,7 @@ def plot_ppc_rootogram(
             **title_kwargs,
         )
 
-    if backend in ("matplotlib", "bokeh"):
+    if backend in ("matplotlib", "bokeh", "plotly"):
```
aloctavodia, line 317:
```suggestion
```

aloctavodia:
No need to fix it now (we can open a separated PR). But for plotly and bokeh we need a rule to select fewer yticks values, like what matplotlib does automatically.

PR #189, opened 2025-03-22
adds rule to select fewer ticks in sqrt scale for bokeh and plotly

### LINKED ISSUE #184: Select fewer y-ticks values in sqrt scale for Plotly & Bokeh

The-Broken-Keyboard:
closes #184 

Instead of adding all integer ticks within range [0,max_y_value], new rule adds fewer ticks in same range. 

Bokeh:

image

Plotly:

image

### DIFF
--- a/src/arviz_plots/backend/bokeh/__init__.py
+++ b/src/arviz_plots/backend/bokeh/__init__.py
@@ -44,8 +44,8 @@ def set_sqrt_yscale(target):
 
     ticks = []
     if max_y > 0:
-        for i in range(math.ceil(max_y) + 1):
-            ticks.append(np.sqrt(i))
+        for i in range(math.ceil(np.sqrt(max_y)) + 1):
+            ticks.append(i)
 
     target.yaxis.formatter = CustomJSTickFormatter(
         code="""
--- a/src/arviz_plots/backend/plotly/__init__.py
+++ b/src/arviz_plots/backend/plotly/__init__.py
@@ -5,6 +5,7 @@ Notes
 :term:`artists` are returned, but it seems modifying them won't modify the figure.
 """
 
+import math
 import re
 import warnings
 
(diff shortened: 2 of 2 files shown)


aloctavodia:
This will still be problematic if frequencies have large values, for instance

dt = azb.from_dict({
    'observed_data': {"obs":pz.Poisson(2).rvs(1000)},
    'posterior_predictive': {"obs":pz.Poisson(2).rvs((2, 1000, 1000))}
})

The-Broken-Keyboard:
Yeah, true. I just realised that it will be problematic if maximum y value is high. 

To handle this, Can we have a default fix number of ticks which we can determine dynamically between minimum y value and maximum y value? So even for high values, the number of ticks can be fixed.

Or else, we can have a dynamic rule which will determine the good number of ticks for a given range of values. Then determining those ticks won't be a difficult task to do....

What do you suggest @aloctavodia ?

aloctavodia:
It is probably a good idea to check what matplotlib does by default and implement something similar.

The-Broken-Keyboard:
Heyy @aloctavodia 
I just went through the implementation in Matplotlib ( reference link : https://matplotlib.org/stable/api/ticker_api.html#matplotlib.ticker.AutoLocator ), they have a upper limit for no. of ticks which is 9 and lower limit for no. of ticks which is 1. They compute no. of integral intervals in the given range of y values, if they fall between 2-9 then they place those many ticks. If it goes above or below limits then they clip those values on 1 or 9 ( as you can see in below given screenshot ) . This is similar approach to what I suggested, either to fix no. of ticks or to find optimal no. of ticks based upon the given y value range. Let me know what would you prefer to do.....

image

aloctavodia:
Fixing the number of ticks like Matplotlib seems reasonable.

The-Broken-Keyboard:
> One more details. There is a lot of blank space below 0 for plotly. The plot should start at 0 like it does fro bokeh and matplotlib
> 
> imagen

Yeah, I kept that knowingly because the circular dots in plotly ( which are little bigger in size than a normal dot ) would get cut if they fall on value 0. I can remove it or I can reduce the gap as well. Whatever you suggest @aloctavodia .

aloctavodia:
could you please update the images in the PR description?

aloctavodia (approved):

The-Broken-Keyboard:
@aloctavodia actually I am currently in a class so couldn't update image when you asked.
I will update it once I get out of this class. Sorry for delay.

aloctavodia:
Sure, no rush. Thanks

PR #194, opened 2025-03-31
Improve docstring of `plot_ppc_pit`

AlexAndorra:
- Just a small improvements to make `plot_ppc_pit` easier to interpret
- Also added pixi support

----
📚 Documentation preview 📚: https://arviz-plots--194.org.readthedocs.build/en/194/

### DIFF
--- a/src/arviz_plots/plots/ppcpitplot.py
+++ b/src/arviz_plots/plots/ppcpitplot.py
@@ -1,4 +1,5 @@
 """Plot ppc pit."""
+
 import warnings
 from copy import copy
 from importlib import import_module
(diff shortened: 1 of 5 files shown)


src/arviz_plots/plots/ppcpitplot.py
```diff
@@ -47,12 +48,17 @@ def plot_ppc_pit(
...
 
     Alternatively, we can visualize the coverage of the central posterior credible intervals by
     setting ``coverage=True``. This allows us to assess whether the credible intervals includes
-    the observed values. We can obtain the coverage of the central intervals from the PIT by
+    the observed values.
+
+    If the difference is positive, the model is under-confident: the predictions have a wider spread than the data -- they are too uncertain.
+    If the difference is negative, the model is over-confident: the predictions have a narrower spread than the data -- they are too certain.
+
+    We can obtain the coverage of the central intervals from the PIT by
```
aloctavodia, line 61: Better to link here https://arviz-devs.github.io/EABM/Chapters/Prior_posterior_predictive_checks.html#pit-ecdfs. The link should also be added to `plot_ecdf_pit` and `plot_loo_pit`

aloctavodia (commented):
@amaloney Is it fine to add pixi, or will this not survive the migration to the monorepo?

amaloney:
> @amaloney Is it fine to add pixi, or will this not survive the migration to the monorepo?

`pixi` is working on the concepts of a workspace https://github.com/prefix-dev/pixi/discussions/387, but right now, no it will not survive the move to the monorepo setup without a refactor. I did spend time trying to make it work, but ended up abandoning it for now.

AlexAndorra:
Ah shoot 🤦‍♂️ Then will close that one and reopen a clean PR

AlexAndorra:
BTW, what do you use for virtual envs in this repo @aloctavodia and @amaloney, since you're not using pixi? Because I don't see conda envs either

amaloney:
> BTW, what do you use for virtual envs in this repo @aloctavodia and @amaloney, since you're not using pixi? Because I don't see conda envs either

I don't think we currently have a recommended method for creating virtual environments with the new ArviZ packages. I think using pixi is a perfectly reasonable way to do it, as well as any of the other systems like conda or uv, or even poetry. We should remedy not having a preferred method in the documentation somewhere.

aloctavodia:
I think it is fine to wait until we have the monorepo to provide that preferred method.

PR #199, opened 2025-04-05
handles title spacing during creation of plot

### LINKED ISSUE #94: Better handling of titles in plotly
AFAIK, plotly doesn't have a constrained_layout=True that can easily fix the issue on the user side, so we should look into handling titles a bit better so they don't overlap so much with the upper subplot (if there is one). A quick look at the example gallery shows this quite clearly.

I think plotly reserves this space at plot creation time, so a potential workaround would be to add titles to the plots when they are created, then update them so they show an empty string yet hopefully keeping the grid spacing with space for the title to potentially be filled later by the plotting functions.

The-Broken-Keyboard:
Closes #94 

Since the titles for subplots are added after plots are created in plotly, that is why there occurs an overlap between titles and upper subplots.

This PR solves this issue by adding an empty string while creating plots to reserve places for titles, which can be populated later.

examples:

Before 1:

image

After 1:

image

Before 2:

image

After 2:

image

----
📚 Documentation preview 📚: https://arviz-plots--199.org.readthedocs.build/en/199/

### DIFF
--- a/src/arviz_plots/backend/plotly/__init__.py
+++ b/src/arviz_plots/backend/plotly/__init__.py
@@ -309,6 +309,7 @@ def create_plotting_grid(
         shared_yaxes=sharey,
         start_cell="top-left",
         horizontal_spacing=plot_hspace,
+        subplot_titles=[" " for i in range(int(rows) * int(cols))],
         column_widths=width_ratios if width_ratios is None else list(width_ratios),
         **kwargs,
     )


aloctavodia:
This fix many instances. But still overlap for some examples, see https://arviz-plots--199.org.readthedocs.build/en/199/gallery/00_plot_rank.html and https://arviz-plots--199.org.readthedocs.build/en/199/gallery/02_plot_ess_evolution.html

The-Broken-Keyboard:
> This fix many instances. But still overlap for some examples, see https://arviz-plots--199.org.readthedocs.build/en/199/gallery/00_plot_rank.html and https://arviz-plots--199.org.readthedocs.build/en/199/gallery/02_plot_ess_evolution.html

Yeah, there are conflicts in these plots which I have noticed and I am already in try to figure out solution for this. But these issues are not because of Titles of subplots....these are because of X-axis labels of each sub plot (they are also being set later after creation of plot) ..... unfortunately there is no direct way to put a blank x label while creating subplot like titles....so I am looking for some work around.

You can see @aloctavodia , both examples which you shared has x label conflicts...not title conflicts....


So titles are handled now, handling x labels is a different task imo....what do you say?

The-Broken-Keyboard:
@aloctavodia  one work around is to keep x labels only for bottom most row .....since it will be evident that same label applies to upper plots too... ( That's what I have seen in many past cases )

aloctavodia:
Tangential to this PR,  for these types of plots, we may want to default to sharing the x-axis.  That may help to make the plots less crowded.

The-Broken-Keyboard:
> Tangential to this PR, for these types of plots, we may want to default to sharing the x-axis. That may help to make the plots less crowded.

Yeah, sharing the x-axis would work...then we can just put 1 x label at bottom of whole plot which will be shared x label

aloctavodia:
We can merge this, as this will improve many of the plots. But even when the original issue mentioned "titles", it was clearly about a broader problem of overlapping elements.

The-Broken-Keyboard:
@aloctavodia yeah I understand that, I am working on it and I will soon create another PR with solution for x labels too. 

I am already looking for that solution

PR #207, opened 2025-04-08
Add combine_plots, allowing to arrange a set of existing plots into a single chart

aloctavodia:
A common task for presenting results is combining different plots in a single figure. This can be done with image editing tools. But that can be overkill for simple tasks like adding a couple of plots side by side.

Here, I am sharing a possible approach to that, but not sure if this is the best route. Before keep working on this solution I want to get feedback to improve this or move in a completely different directions. 

 In this approach, we have a function `combine_plots` that accepts a list of tuple. The first element of the tuple are ArviZ plotting functions, and the second elementis  optional a dictionary of kwargs

So for instance we can do 

```python
dt = azb.load_arviz_data('centered_eight')

combine_plots(dt, plots=[(azp.plot_dist, {"kind": "kde",}),
                         (azp.plot_dist, {"kind": "ecdf",}),
                         (azp.plot_dist, {"kind": "hist"}),
                        ],
                expanded_dim="column",
                group="posterior",
                var_names=["mu", "tau"],                         
                         )
```

or along rows

```python
combine_plots(dt, plots=[(azp.plot_dist, {"kind": "kde",}),
                         (azp.plot_dist, {"kind": "ecdf",}),
                         (azp.plot_dist, {"kind": "hist"}),
                        ],
                expanded_dim="rows",
```

or 

```python
combine_plots(dt, plots=[(azp.plot_ppc_pit, {}),
                         (azp.plot_ppc_rootogram, {}),
                        ],
                var_names="home_points",
                group="posterior_predictive",         
                         )          
``` 

                 
And it does not currently work for some other plots

```python
dt = azb.load_arviz_data("rugby")
combine_plots(dt, plots=[
                         (azp.plot_ess_evolution, {}),
                         (azp.plot_dist, {}),
                         (azp.plot_rank, {}),
],
                var_names="mu",
                         )
```
            

----
📚 Documentation preview 📚: https://arviz-plots--207.org.readthedocs.build/en/207/

### DIFF
--- a/docs/source/gallery/mixed/combine_plots.py
+++ b/docs/source/gallery/mixed/combine_plots.py
@@ -1,0 +1,30 @@
+"""
+# Custom diagnostic plots combination 
+
+Column layout with user requested diagnostic plots: ESS evolution plot, rank plot and autocorrelation plot.
+
+---
+
+:::{seealso}
+API Documentation: {func}`~arviz_plots.combine_plots`
+:::
+"""
+from arviz_base import load_arviz_data
+
+import arviz_plots as azp
+
+azp.style.use("arviz-variat")
+
+data = load_arviz_data("non_centered_eight")
+pc = azp.combine_plots(
+    data,
+    [
+        (azp.plot_ess_evolution, {}),
+        (azp.plot_rank, {}),
+        (azp.plot_autocorr, {}),
+    ],
+    var_names=["theta", "mu", "tau"],
+    coords={"school": ["Hotchkiss", "St. Paul's"]},
+    backend="none"  # change to preferred backend
+)
+pc.show()
--- a/docs/source/gallery/mixed/combine_plots.py
+++ b/docs/source/gallery/mixed/combine_plots.py
@@ -26,6 +26,7 @@ pc = azp.combine_plots(
     ],
     var_names=["theta", "mu", "tau"],
     coords={"school": ["Hotchkiss", "St. Paul's"]},
-    backend="none"  # change to preferred backend
+    backend="none",  # change to preferred backend
+    pc_kwargs={"plot_grid_kws": {"figsize": (10, 7)}}
 )
 pc.show()
--- a/docs/source/gallery/mixed/combine_plots.py
+++ b/docs/source/gallery/mixed/combine_plots.py
@@ -2,7 +2,7 @@
 # Custom diagnostic plots combination 
 
 Arrange three diagnostic plots (ESS evolution plot, rank plot and autocorrelation plot)
-in a custom row layout.
+in a custom column layout.
 
 ---
 
@@ -26,7 +26,6 @@ pc = azp.combine_plots(
     ],
     var_names=["theta", "mu", "tau"],
     coords={"school": ["Hotchkiss", "St. Paul's"]},
-    expand="row",
     backend="none"  # change to preferred backend
 )
 pc.show()
--- a/docs/source/gallery/mixed/combine_plots.py
+++ b/docs/source/gallery/mixed/combine_plots.py
@@ -1,7 +1,8 @@
 """
 # Custom diagnostic plots combination 
 
-Column layout with user requested diagnostic plots: ESS evolution plot, rank plot and autocorrelation plot.
+Arrange three diagnostic plots (ESS evolution plot, rank plot and autocorrelation plot)
+in a custom row layout.
 
 ---
 
@@ -25,6 +26,7 @@ pc = azp.combine_plots(
     ],
     var_names=["theta", "mu", "tau"],
     coords={"school": ["Hotchkiss", "St. Paul's"]},
+    expand="row",
     backend="none"  # change to preferred backend
 )
 pc.show()
--- a/src/arviz_plots/backend/none/__init__.py
+++ b/src/arviz_plots/backend/none/__init__.py
@@ -196,7 +196,7 @@ def hist(
     r_e,
     target,
     *,
-    bottom=None,
+    bottom=0,
     color=unset,
     facecolor=unset,
     edgecolor=unset,
@@ -287,7 +287,7 @@ def step(x, y, target, *, color=unset, alpha=unset, width=unset, linestyle=unset
     if not ALLOW_KWARGS and artist_kws:
         raise ValueError("artist_kws not empty")
     artist_element = {
-        "function": "line",
+        "function": "step",
         "x": np.atleast_1d(x),
         "y": np.atleast_1d(y),
         **_filter_kwargs(kwargs, artist_kws),
@@ -398,7 +398,7 @@ def ciliney(
     if not ALLOW_KWARGS and artist_kws:
         raise ValueError("artist_kws not empty")
     artist_element = {
-        "function": "line",
+        "function": "ciliney",
         "x": np.atleast_1d(x),
         "y_bottom": np.atleast_1d(y_bottom),
         "y_top": np.atleast_1d(y_top),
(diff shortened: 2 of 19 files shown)


src/arviz_plots/backend/none/__init__.py
```diff
@@ -398,7 +398,7 @@ def ciliney(
     if not ALLOW_KWARGS and artist_kws:
         raise ValueError("artist_kws not empty")
     artist_element = {
-        "function": "line",
+        "function": "ciliney",
```
OriolAbril, line 401: quite off scope but I did realize here. This funcion ends up calling `plot` which is the same artist that is called by `line`, the difference is it calls `plot([x, x], [y_bottom, y_top]...)` but this [x, x] transform/cleanup should be done at the `visuals` level to keep the backend level as minimal as possible.

Extra note: if using [`vlines`](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.vlines.html) here for whatever reason (performance, better fit for non matplotlib backends...) then it would make sense to have it as its own backend function

src/arviz_plots/plot_collection.py
```diff
@@ -689,8 +689,8 @@ def wrap(
                 {
                     "chart": np.array(fig, dtype=object),
                     "plot": (dims, flat_ax_ary.reshape(plots_raw_shape)),
-                    "row": (dims, flat_row_id.reshape(plots_raw_shape)),
-                    "col": (dims, flat_col_id.reshape(plots_raw_shape)),
+                    "row_index": (dims, flat_row_id.reshape(plots_raw_shape)),
+                    "col_index": (dims, flat_col_id.reshape(plots_raw_shape)),
```
OriolAbril, line 693: I had to rename these variables because this function adds a `column` or `row` coordinate, and coordinates and variables can't have the same name. I like these names better I think, maybe even `column_index`, but this means we need to update this page of the docs too: https://arviz-plots.readthedocs.io/en/latest/tutorials/intro_to_plotcollection.html#customizing-your-plotcollection. Commenting so I don't forget

src/arviz_plots/plot_collection.py
```diff
@@ -1019,8 +1019,11 @@ def map(
 
         for var_name, sel, isel in plotters:
             da = data[var_name].sel(sel)
-            if np.all(np.isnan(da)):
-                continue
+            try:
+                if np.all(np.isnan(da)):
+                    continue
+            except TypeError:
+                pass
```
OriolAbril, line 1026: with none backend the elements in `da` can be lists or dicts, so isnan errors out. This allows the new `render` function to work through `.map` and I think the behaviour is still exactly what we want: if everything is exactly nan do nothing, otherwise go ahead and call the function which should work with whatever it gets

src/arviz_plots/plots/combine.py
```diff
@@ -0,0 +1,95 @@
...
+
+def backendize_kwargs(kwargs, backend):
+    return {key: get_valid_arg(key, value, backend) for key, value in kwargs.items() if key != "function"}
+
+def render(da, target, backend, **kwargs):
+    plot_backend = import_module(f"arviz_plots.backend.{backend}")
+    plot_kwargs = da.item()
+    plot_fun_name = plot_kwargs["function"]
+    plot_kwargs = backendize_kwargs(plot_kwargs, backend)
+    return getattr(plot_backend, plot_fun_name)(target=target, **{**plot_kwargs, **kwargs})
```
OriolAbril, line 28: This probably needs a better place to live in. Also happy to get better naming suggestions.

src/arviz_plots/plots/combine.py
```diff
@@ -0,0 +1,95 @@
...
+        expanded_coord_names = [
+            getattr(elem[0], "__name__") + f"_{idx:02d}" for idx, elem in enumerate(plots)
+        ]
+    sample_dims = rcParams["data.sample_dims"]
+    column_length = len(expanded_coord_names)
+
+    distribution = process_group_variables_coords(
+        dt, group=group, var_names=var_names, filter_vars=filter_vars, coords=coords
+    )
+    facet_dims = ["__variable__"] + ([] if "predictive" in group else [dim for dim in distribution.dims if dim not in sample_dims])
```
OriolAbril, line 51: This is quite the assumption, we might want to make this an argument of the function with that as default

src/arviz_plots/plots/combine.py
```diff
@@ -0,0 +1,95 @@
...
+        pc.coords = {expanded_dim: name}
+        inverted_dt = {}
+        for viz_group, ds in pc_i.viz.children.items():
+            for viz_var_name, da in ds.data_vars.items():
+                if viz_var_name in {"plot", "row_index", "col_index"}:
+                    continue
+                if viz_var_name not in inverted_dt:
+                    inverted_dt[viz_var_name] = {}
+                inverted_dt[viz_var_name].update({viz_group: da})
+        inverted_dt = xr.DataTree.from_dict({key: xr.Dataset(values) for key, values in inverted_dt.items()})
```
OriolAbril, line 91: The `.viz` datatree has the following structure:

```
⊦ var_name1
|        ⊦ artist1
|        ⊦ artist2
|        L ...
⊦ var_name2
|        ⊦ artist1
|        ⊦ artist2
|        L ...
L ...
```

However, in this case the alternative:

```
⊦ artist1
|        ⊦ var_name1
|        ⊦ var_name2
|        L ...
⊦ artist2
|        ⊦ var_name1
|        ⊦ var_name2
|        L ...
L ...
```

works better as it allows us to loop through groups only then have `.map` take care of all the looping across variables and matching plot elements to be rendered with their target.

src/arviz_plots/plots/combine.py
```diff
@@ -0,0 +1,101 @@
...
+    sample_dims = rcParams["data.sample_dims"]
+    column_length = len(expanded_coord_names)
+
+    distribution = process_group_variables_coords(
+        dt, group=group, var_names=var_names, filter_vars=filter_vars, coords=coords
+    )
+    facet_dims = ["__variable__"] + ([] if "predictive" in group else [dim for dim in distribution.dims if dim not in sample_dims])
+
+    pc_kwargs = {}
+    pc_kwargs["plot_grid_kws"] = {}
```
aloctavodia, line 56:
```suggestion
    if pc_kwargs is None:
        pc_kwargs = {}
    pc_kwargs["plot_grid_kws"] = pc_kwargs.get("plot_grid_kws", {}).copy()
```

docs/source/gallery/mixed/combine_plots.py
```diff
@@ -0,0 +1,30 @@
+"""
+# Custom diagnostic plots combination 
+
+Column layout with user requested diagnostic plots: ESS evolution plot, rank plot and autocorrelation plot.
```
aloctavodia, line 4:
```suggestion
# Combination of plots

Arrange multiple plots in a customizable column or row layout. For example, combine ESS evolution, rank, and autocorrelation plots into a single view.

```

aloctavodia: or something like this

src/arviz_plots/plots/__init__.py
```diff
@@ -2,6 +2,7 @@
 
 from .autocorr_plot import plot_autocorr
 from .bf_plot import plot_bf
+from .combine import combine_plots
```
aloctavodia, line 5: Is the name `combine_plots` intentional instead of `plot_*`? It makes sense to make the name really different, jsut checking. 
Are we happy with `combine` or do we prefer alternatives like `layout`, `grid`, `arrange`?

aloctavodia (commented):
a few small comments

src/arviz_plots/plots/__init__.py
OriolAbril: I am ok with `combine_plots` but I wouldn't oppose something like `plot_layout`, `plot_grid` (this one might be confusing with the .grid method of plotcollection though) or `plot_combined` to keep the `verb_name` pattern

docs/source/gallery/mixed/combine_plots.py
OriolAbril: I really like the description but I think it is better for the docstring instead of the example as it describes the capabilities of the function instead of the specific example. I'll update the docstring if you agree, I can update the example description to use language closer to that while being specific to the example at hand too

aloctavodia: OK, I have been adding more descriptive and general text to some of the gallery elements. My rationally is that people don't read docs that much, so a little bit of redundancy may be good. But that's something to discuss.

src/arviz_plots/plots/__init__.py
aloctavodia: No strong preference either.

Issue #204, opened 2025-04-07
Fix  x label conflicts


aloctavodia:
See #199

The-Broken-Keyboard:
I will solve it

aloctavodia:
Thanks!

The-Broken-Keyboard:
@aloctavodia , I tried few things. Issue is that plotly doesn't allow to give x axis label for subplots while creating them.  Axis labels are added later only. so there are 2 possible solutions for this in our context

1. We can add some vertical spacing while creating subplots so that if x labels are given later then they won't  cause any conflict. But drawback with this solution is that setting vertical spacing causes compression of plots in other cases where x labels are not required and there exists good amount of empty space between plots. 

for example:
for these plots, issue is solved
![Image](https://github.com/user-attachments/assets/6a5efe3a-3318-4538-b583-61e509b841b7)

for these plots, plots get compressed and empty space exists

![Image](https://github.com/user-attachments/assets/51c5b376-8bdc-4cc2-bdac-f23ced44eded)

![Image](https://github.com/user-attachments/assets/b69cc035-b48b-4ee7-8caa-827005dacc39)

2. Second solution is what you suggested. Since this issue occurs only for specific plots we can set their shared x as true and we can just give a common x label at bottom. For this we will require to keep a flag which will indicate that for this plot the shared x has been set true so that we won't set x labels for each subplot

what do you suggest?

aloctavodia:
My suggestion to set sharex True by default is restricted to plots where the x-axis values are "naturally" shared. This includes ESS plots, rank plot, trace plot etc, and excludes like plot_dist. I think that's a good default independent of this overlapping issue.

Adding vertical space can be problematic unless the user has a way to tweak it, or we can provide good defaults. Maybe this is something we should move upstream to the plotly people.

@OriolAbril?

The-Broken-Keyboard:
I tried for few default values with vertical spacing. But good for one always makes some other plot weird. Since there is different vertical spacing value required for different no. of rows and also it depends on if we would be adding x labels for subplots later or not. Giving users a way to tweak this value can be a solution but trying many different vertical spacing values can be cumbersome for users, but this will definitely enable them to rectify the overlap of their plots.

I would set share x value as true for the plots like ESS, Rank etc. while for others we can escalate this to plotly people. If they fix this then best for us, otherwise we can give users the power to change vertical spacing value. @aloctavodia

PR #208, opened 2025-04-13
fixes shared x property for appropriate plots ( like rank plot, trace plot etc )

The-Broken-Keyboard:
This PR sets default value of `sharex` as True for plots like rank plot, trace plot, ess plot etc. ( the plots which share x axis ). 
It first sets default value as True then it sets x label accordingly. If sharex has been set as True then it sets x axis label for the subplots in last row only. 

Also no direct way was available to check if Shared x is true or not, so wrote a work around function `is_shared_x` which takes chart and checks for shared x value.

Few Examples of fixes:

Before 1:

image

After 1:

image

Before 2:

image

After 2:

image

Before 3:

image

After 3:

image

Also this PR solves conflicts due to x labels ( #204 ) of plots which share their x axes.

----
📚 Documentation preview 📚: https://arviz-plots--208.org.readthedocs.build/en/208/

### DIFF
--- a/src/arviz_plots/backend/plotly/__init__.py
+++ b/src/arviz_plots/backend/plotly/__init__.py
@@ -26,6 +26,28 @@ unset = UnsetDefault()
 pat = re.compile(r"^(row|col)\s?:")
 
 
+def is_shared_x(fig):
+    """Check if all x-axes are shared in the given figure."""
+    x_axes = [fig.layout[key] for key in fig.layout if key.startswith("xaxis")]
+
+    if len(x_axes) <= 1:
+        return True
+
+    master_axis = None
+    for axis in x_axes:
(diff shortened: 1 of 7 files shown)


aloctavodia:
We should include `plot_psense_quantities`, `plot_ppc_pit`, `plot_loo_pit`

The-Broken-Keyboard:
@aloctavodia , for plot_psense_quantities sharex was already set as true by default, so I didn't had to change anything there. That's why there were no changes in that file. For plot_ppc_pit , I have made the change and for plot_loo_pit we don't need to make any change because it calls plot_ecdf_pit internally first and plot_ecdf_pit itself sets sharex as True. So that is handled in ecdf_pit itself.

aloctavodia:
Perfect, thanks for checking!

PR #221, opened 2025-04-17
Add PlotMatrix plotting manager class

### LINKED ISSUE #48: Add PlotMatrix
Add a 2nd plot manager class, similar to PlotCollection but that generates a matrix (with a diagonal too optionally) with all the pairwise subset combinations. It will be needed for:

* plot_pair
* plot_elpd

My idea would be to use the same method names as PlotCollection whenever possible, adding new ones for the other features.
 
For example:

* `.map` will also exist, and will be mapping over the diagonal which is the only mapping operation that can maintain the method signature. Thus, the diagonal of `plot_pair` could be populated by calling `plot_dist` using the `PlotMatrix` class instead of a `PlotCollection`.
* `.map_lower` and `.map_upper` to map functions to the plot below the diagonal (lower triangular matrix) and above. These would have signature `.map_xyz(da_x, da_y, <same as .map>)` with da_x being the data to be mapped to the x axis (thus selected with the column index) and da_y the data mapped to the y axis.

OriolAbril:
Still very early draft, but already working for the base cases. Here are some preliminary examples:

```python
import arviz_base as azb
import arviz_plots as azp
from arviz_plots.plot_matrix import PlotMatrix

azp.style.use("arviz-variat")
post = azb.load_arviz_data("centered_eight").posterior.dataset.isel(school=slice(0, 3))

def label_diag(da, target, backend, var_name, sel, isel, **kwargs):
    target.axis("off")
    return target.text(0.2, 0.5, labeller.make_label_vert(var_name, sel, isel), size=8)

def label_couple(da_x, da_y, target, backend, var_name_x, sel_x, isel_x, var_name_y, sel_y, isel_y,**kwargs):
    target.axis("off")
    return target.text(
        0.2,
        0.5,
        f"x data: {labeller.make_label_vert(var_name_x, sel_x, isel_x)}\n---\ny data: {labeller.make_label_vert(var_name_y, sel_y, isel_y)}",
        size=6,
    )

pc = PlotMatrix(post, ["__variable__", "school"], plot_grid_kws={"figsize": (5, 4)})
labeller = azb.labels.BaseLabeller()
pc.map(label_diag, subset_info=True)
pc.map_triangle(label_couple, subset_info=True)
```

imatge

```python
def scatter_pairplot(da_x, da_y, target, backend, **kwargs):
    target.axis("off")
    return target.plot(da_x, da_y, ".", **kwargs)[0]

pc = PlotMatrix(post, ["__variable__", "school"], aes={"color": ["chain"]}, plot_grid_kws={"figsize": (5, 4)})
labeller = azb.labels.BaseLabeller()
pc.map(label_diag, subset_info=True, ignore_aes={"color"})
pc.map_triangle(scatter_pairplot, triangle="both", alpha=.2)
```

imatge

----
📚 Documentation preview 📚: https://arviz-plots--221.org.readthedocs.build/en/221/


OriolAbril:
I added some tests and I think it is usable, but will probably need some extra work once we start working on `plot_pair`. As of now, the main known issue is `__variable__` must be one of the faceting dimensions. It might be possible to not force this but I am not sure what the behaviour should be in such cases for the off-diagonal plots. If we decide to keep the faceting variable behaviour we should improve docs and error messages and/or defaults too.

PR #223, opened 2025-04-18
Add references to ppc plots

### LINKED ISSUE #200: Add reference values to plot_ppc_dist

rohanbabbar04:
Closes #200 
- Add reference values to plot_ppc_dist

### DIFF
--- a/src/arviz_plots/plots/ppc_dist_plot.py
+++ b/src/arviz_plots/plots/ppc_dist_plot.py
@@ -5,13 +5,14 @@ from copy import copy
 from importlib import import_module
 
 import numpy as np
+import xarray as xr
 from arviz_base import rcParams
 from arviz_base.labels import BaseLabeller
 
 from arviz_plots.plot_collection import PlotCollection
 from arviz_plots.plots.dist_plot import plot_dist
 from arviz_plots.plots.utils import filter_aes, process_group_variables_coords, set_wrap_layout
-from arviz_plots.visuals import ecdf_line, hist, line_xy
+from arviz_plots.visuals import ecdf_line, hist, line_xy, vline
 
 
 def plot_ppc_dist(
@@ -24,6 +25,7 @@ def plot_ppc_dist(
     sample_dims=None,
     kind=None,
     num_samples=50,
+    references=None,
     plot_collection=None,
     backend=None,
     labeller=None,
@@ -62,6 +64,8 @@ def plot_ppc_dist(
         Defaults to ``rcParams["plot.density_kind"]``
     num_samples : int, optional
         Number of samples to plot. Defaults to 100.
+    references : int, float, list, tuple
+        Value(s) used as reference points representing prior knowledge.
     plot_collection : PlotCollection, optional
     backend : {"matplotlib", "bokeh"}, optional
     labeller : labeller, optional
@@ -294,6 +298,19 @@ def plot_ppc_dist(
                 data=dt_observed,
                 ignore_aes=observed_ignore,
                 **observed_density_kwargs,
+            )
+    if references is not None:
+        _, ref_aes, ref_ignore = filter_aes(plot_collection, aes_map, "references", "sample")
+        ref_kwargs = {}
+        if "color" not in ref_aes:
+            ref_kwargs.setdefault("color", "black")
+        if "linestyle" not in ref_aes:
+            ref_kwargs.setdefault("linestyle", plot_bknd.get_default_aes("linestyle", 2, {})[1])
+        references = [references] if np.isscalar(references) else references
+        for value in references:
+            ref_dt = xr.Dataset({list(observed_dist.data_vars)[0]: xr.DataArray(value)})
+            plot_collection.map(
+                vline, "reference", data=ref_dt, ignore_aes=ref_ignore, **ref_kwargs
             )
 
     return plot_collection


rohanbabbar04:
For dictionaries, if we want to retain the default behavior of rotating the text to 90 degrees, would it make sense to pass a text argument to vline for each of the backends — Matplotlib, Bokeh, and Plotly — since the rotation parameters differ between them?

Additionally, we’d prefer to place the text vertically somewhere around the middle of the line (e.g., at max(y_range) * 0.3). Should we go ahead with this approach, or is there a better alternative?

An code change I could think of

```python
def vline(x, target, *, text_label=None, color=unset, alpha=unset, width=unset, linestyle=unset, **artist_kws):
    """Interface to matplotlib for a vertical line spanning the whole axes."""
    artist_kws.setdefault("zorder", 0)
    if text_label is not None:
        text_kwargs = {"rotation": 90, "alpha": 0.5, "color": "black"}
        text(x, target.get_ylim()[1] * 0.3, text_label, target, **text_kwargs)
    kwargs = {"color": color, "alpha": alpha, "linewidth": width, "linestyle": linestyle}
    return target.axvline(x, **_filter_kwargs(kwargs, Line2D, artist_kws))

```

rohanbabbar04:
- Have updated the `text` method in plotly to use Annotate instead of Scatter which supports `textangle` and performs as it was.
- For the plot, I would want the annotate label to be around the middle of the vline, we can easily do it for matplotlib using `get_ylim()`, will it be possible to do it in bokeh or plotly.

rohanbabbar04:
Would there be a requirement to have different references for different data variables?

aloctavodia:
yes, references could be different for different variables.

OriolAbril:
It will be more work initially but I think it will be better overall. As we'll most probably want to add reference lines to other plots too, I think it would be best to do so via a specific helper function. My initial idea would be to have a function that takes a plotcollection as a required argument and the references as int, lists, dict or xarray object and then adds the line and optionally some text too.

As the plotcollection would be required, the int/list/dict references can be internally converted to a compatible dataset so that the function can call .map for the line and then if necessary for the text too.

rohanbabbar04:
Yup, I think making a utils function for plot_references makes the best sense..
Keeping the text aside, this is how it works

```python
from arviz_plots import style, plot_ppc_dist, plot_mcse
style.use("arviz-variat")
radon = load_arviz_data('rugby')
print(radon.data_vars)
pc = plot_ppc_dist(
        radon,
        backend="matplotlib", references={"home_points": (10, 30), "away_points": (0, 40)})
pc.show()
````

image

**Comment by aloctavodia** [comment]
LGTM as this solves the original issue.

I wonder if we should have a similar function to plot "bands", filled colour between two points defined either on the y-axis or x-axis. It should change the name from `plot_reference` to `plot_reference_lines` or `add_reference_line` or `add_lines`. And also have one argument to switch between vlines and hlines.

**Comment by aloctavodia** [comment]
ohh one more think, probably we want to default to gray instead of black

**Comment by rohanbabbar04** [comment]
> LGTM as this solves the original issue.
> 
> I wonder if we should have a similar function to plot "bands", filled colour between two points defined either on the y-axis or x-axis. If so, then we should change the name from `plot_reference` to `plot_reference_lines` or `add_reference_line` or `add_lines`. And also have one argument to switch between vlines and hlines.

I like the idea, we could have something like plot_bandsin the utils itself

**Comment by aloctavodia** [comment]
I think is ok to have the "reference" argument for `plot_ppc_dist` (and other plot_ppc) plots. But I made `add_reference_lines` easier to use outside of those function. So we can do stuff like this.

```python
dt = load_arviz_data('centered_eight')
pc = azp.plot_dist(
        dt,
        kind='ecdf',
        var_names=["mu", "tau"],
        )
azp.add_reference_lines(pc, references=0)
```

src/arviz_plots/plots/utils.py
```diff
@@ -161,3 +166,70 @@ def set_grid_layout(pc_kwargs, plot_bknd, ds, num_rows=None, num_cols=None):
...
+            if key in data_vars:
+                ref_dt = Dataset({key: DataArray(np.asarray(value))})
+                plot_collection.map(
+                    plot_func, "ref_lines", data=ref_dt, ignore_aes=ref_ignore, **ref_kwargs
+                )
+    else:
+        ref_dt = Dataset({var: DataArray(np.asarray(references)) for var in data_vars})
+        plot_collection.map(
+            plot_func, "ref_lines", data=ref_dt, ignore_aes=ref_ignore, **ref_kwargs
+        )
```
OriolAbril, line 235: Given both if branches generate datasets then call .map, I would move the call to .map outside the if/else block. In fact, I might even "outsource" this logic to a function in arviz-plots that given a dataset and some values in variable format like the references here it returns a new dataset filled with the provided values and similar/compatible dims+coords.

Rough idea:

```suggestion
    ref_dict = {}
    if isinstance(references, dict):
        for key, value in references.items():
            if key in data_vars:
                ref_dict[key] = DataArray(np.asarray(value))
    else:
        ref_dict = {var: DataArray(np.asarray(references)) for var in data_vars}
    plot_collection.map(
        plot_func, "ref_lines", data=Dataset(ref_dict), ignore_aes=ref_ignore, **ref_kwargs
    )
```

Main caveat not being taken into account. The fact that each plot can get multiple reference lines, which might end up as dataarrays with an extra `var_name_0` dimension instead of a shared one (even if it has nans in some positions).

PR #230, opened 2025-04-25
Update add_reference_lines as a helper function

### LINKED ISSUE #229: Add reference lines function follow up
See https://github.com/arviz-devs/arviz-plots/pull/223 comments after merged

rohanbabbar04:
Closes #229 

----
📚 Documentation preview 📚: https://arviz-plots--230.org.readthedocs.build/en/230/

### DIFF
--- a/src/arviz_plots/plots/utils.py
+++ b/src/arviz_plots/plots/utils.py
@@ -1,11 +1,12 @@
 """Utilities for batteries included plots."""
-
+import warnings
 from copy import copy
 from importlib import import_module
 
 import numpy as np
+import xarray as xr
+from arviz_base import rcParams
 from arviz_base.utils import _var_names
-from xarray import DataArray, Dataset
 
 from arviz_plots.plot_collection import concat_model_dict, process_facet_dims
 from arviz_plots.visuals import hlines, vlines
@@ -35,7 +36,7 @@ def get_group(data, group, allow_missing=False):
     KeyError
         If unable to access `group` from `data` and ``allow_missing=False``.
     """
-    if isinstance(data, Dataset):
+    if isinstance(data, xr.Dataset):
         return data
     if hasattr(data, "name") and data.name == group:
         return data.ds
@@ -45,7 +46,7 @@ def get_group(data, group, allow_missing=False):
         if not allow_missing:
             raise
         return None
-    if isinstance(data, Dataset):
+    if isinstance(data, xr.Dataset):
         return data
     return data.ds
 
@@ -168,6 +169,60 @@ def set_grid_layout(pc_kwargs, plot_bknd, ds, num_rows=None, num_cols=None):
     return pc_kwargs
 
 
+def _references_to_dataset(references, ds, sample_dims):
+    """Generate an :class:`~xarray.Dataset` compatible with `ds` from `references`.
+
+    Parameters
+    ----------
+    references : scalar, array-like, dict, DataArray
+        References to cast into a compatible dataset.
+    ds : Dataset
+    sample_dims : iterable of hashable, optional
+
+    Returns
+    -------
+    Dataset
+        A Dataset with a subset of the variables, dimensions and coordinate names in `ds`,
+        with only an extra dimension "ref_line_dim" added when multiple references are requested
+        for one or some of the variables.
+    """
+    if isinstance(references, xr.Dataset):
+        return references
+    if isinstance(references, xr.DataArray):
+        name = references.name
+        if name is not None:
+            if name not in ds.data_vars:
+                warnings.warn(f"{name} not available in {ds.data_vars}", UserWarning)
+            return references.to_dataset()
+        references = references.values
+    if np.isscalar(references):
+        aux_ds = (
+            ds if not sample_dims else ds.isel({dim: 0 for dim in sample_dims if dim in ds.dims})
+        )
+        return xr.full_like(aux_ds, references, dtype=np.array(references).dtype)
+    if isinstance(references, list | tuple | np.ndarray):
+        references = {var_name: references for var_name in ds.data_vars}
+    if isinstance(references, dict):
+        ref_dict = {}
+        for var_name, da in ds.items():
+            if var_name not in references:
+                continue
+            ref_values = references[var_name]
+            sizes = {dim: length for dim, length in da.sizes.items() if dim not in sample_dims}
+            ref_dict[var_name] = xr.DataArray(
+                np.full(list(sizes.values()) + [np.size(ref_values)], ref_values),
+                dims=list(sizes) + ["ref_line_dim"],
+                coords={"ref_line_dim": np.arange(np.size(ref_values))}
+                | {
+                    coord_name: coord_da
+                    for coord_name, coord_da in da.coords.items()
+                    if coord_da.dims[0] not in sample_dims
+                },
+            )
+        return xr.Dataset(ref_dict)
+    raise TypeError("Unrecognized input type for `references`")
+
+
 def add_reference_lines(
     plot_collection,
     references,
@@ -175,7 +230,7 @@ def add_reference_lines(
     aes_map=None,
     plot_kwargs=None,
     backend=None,
-    data_vars=None,
+    sample_dims=None,
 ):
     """Add reference lines.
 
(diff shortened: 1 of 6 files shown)


OriolAbril:
Not sure where is the right place for this, but I think I need an in detail description on how the `references` argument behaves. Say I have a plotcollection from calling `plot_dist` on the centered_eight example data.

* int or float: a single line is added to every [plot](https://arviz-plots.readthedocs.io/en/latest/glossary.html#term-plot)
* array-like: IIUC, it can have any length and as many lines as elements in this array-like will be plotted in each plot.
* dict: IIUC, it must have be `{var_name: array-like}`. We then loop over each variable (which is not necessarily plot) and plot as many lines as elements in the corresponding array-like to every plot of that variable. For example, `{"tau": [1, 3], "theta": [-1, 0, 1]}` would add two lines to the plot for tau and three lines to all 8 plots of theta subsets?

I haven't had time to look at it in depth, but it is probably worth also checking https://github.com/arviz-devs/arviz/pull/2438 for cases to support and behaviour of reference values.

I am not sure how reference values behave in current arviz and I like adding multiple lines at once and not necessarily the same number of lines per plot, but I think we are still far from desirable behaviour with `references`. cc @aloctavodia

src/arviz_plots/plots/utils.py
```diff
@@ -238,16 +243,21 @@ def add_reference_lines(
...
+    ref_dt, artist_dims = {}, {}
     if isinstance(references, dict):
         for key, value in references.items():
             if key in data_vars:
-                ref_dt = Dataset({key: DataArray(np.asarray(value))})
-                plot_collection.map(
-                    plot_func, "ref_lines", data=ref_dt, ignore_aes=ref_ignore, **ref_kwargs
-                )
+                ref_dt[key] = DataArray(np.asarray(value))
+                artist_dims[key] = np.size(value)
```
OriolAbril, line 251: I have now seen it is not very clear in the docs, but `artist_dims` should be a mapping of dimension name to dimension length. Therefore I don't think it should be named the same as the variable (not sure it will work consistently either repeating variable names and dimension names in xarray, I am a bit surprised it works at all in fact). I would hardcode the dimension name to something like `ref_line_dim`

src/arviz_plots/plots/utils.py
```diff
@@ -238,16 +243,21 @@ def add_reference_lines(
...
     else:
-        ref_dt = Dataset({var: DataArray(np.asarray(references)) for var in data_vars})
-        plot_collection.map(
-            plot_func, "ref_lines", data=ref_dt, ignore_aes=ref_ignore, **ref_kwargs
-        )
+        ref_dt = {var: DataArray(np.asarray(references)) for var in data_vars}
+        artist_dims = {var: np.size(references) for var in data_vars}
```
OriolAbril, line 254: similar comment to above, with the mapping being {dim_name : dim_length} it should be a single item dict instead of having one key per variable.

src/arviz_plots/plots/utils.py
```diff
@@ -238,16 +243,21 @@ def add_reference_lines(
         ref_kwargs.setdefault("color", "gray")
     if "linestyle" not in ref_aes:
         ref_kwargs.setdefault("linestyle", plot_bknd.get_default_aes("linestyle", 2, {})[1])
-
+    ref_dt, artist_dims = {}, {}
```
OriolAbril, line 246: very minor comment but can this variable be renamed to `ref_dict`? We generally use `dt` for DataTree, `ds` for Dataset, `da` DataArray and following always the same convention helps read and understand the code faster. I can add a page to the contributing section about this if it helps too

rohanbabbar04:
I have made the changes, along with some minor adjustments to the code in `_references_to_dataset`. I also carried out some tests with different plots. I think `plot_ppc_dist` has its own `sample_dims = pp_dims`, which we would need to provide. Other than that, everything worked fine.
Should we work on adding references to all plots as a parameter, or do we expect users to provide `sample_dims`?

OriolAbril:
I will open a PR to add the `references_to_dataset` function to arviz-base then we can use it here. For now let's leave the current behaviour, if people ask for more things we can consider them later and add them if needed.

rohanbabbar04:
Sounds good, you can review this PR and if everything looks good. We can merge it.

rohanbabbar04:
Sounds good, I just saw the PR, should we keep `add_reference_lines` as a utils function or would it be preferable to keep references as a parameter like we have right now for `plot_ppc_dist`.

OriolAbril:
IIRC @aloctavodia proposal was to not have `references` argument anywhere but have the `add_reference_lines` helper function plus examples on the gallery using it. Internally, the `add_reference_lines` should use the function in arviz_base, the dependency is on the github version for now so once the PR is merged it can be used here.

src/arviz_plots/plots/utils.py
```diff
@@ -168,14 +169,68 @@ def set_grid_layout(pc_kwargs, plot_bknd, ds, num_rows=None, num_cols=None):
...
+    raise TypeError("Unrecognized input type for `references`")
+
+
 def add_reference_lines(
     plot_collection,
     references,
     orientation="vertical",
     aes_map=None,
     plot_kwargs=None,
     backend=None,
```
OriolAbril, line 232: Given `plot_collection` is a required argument, backend should not be an argument at all, the backend should be retrieved from the plot_collection to make sure both are aligned, otherwise plotting won't work.

src/arviz_plots/plots/utils.py
```diff
@@ -224,30 +285,32 @@ def add_reference_lines(
...
+        sample_dims = rcParams["data.sample_dims"]
+    if isinstance(sample_dims, str):
+        sample_dims = [sample_dims]
 
     plot_bknd = import_module(f".backend.{backend}", package="arviz_plots")
 
     plot_func = vlines if orientation == "vertical" else hlines
 
-    _, ref_aes, ref_ignore = filter_aes(plot_collection, aes_map, "reference", "sample")
+    _, ref_aes, ref_ignore = filter_aes(plot_collection, aes_map, "reference", sample_dims)
```
OriolAbril, line 297: I would use `ref_line` and then also allow a 2nd key in aes_map and plot_kwargs `ref_text` (with default for text to now show it and also can be added in a follow-up PR).

src/arviz_plots/plots/utils.py
```diff
@@ -224,30 +285,32 @@ def add_reference_lines(
...
-                ref_dt = Dataset({key: DataArray(np.asarray(value))})
-                plot_collection.map(
-                    plot_func, "ref_lines", data=ref_dt, ignore_aes=ref_ignore, **ref_kwargs
-                )
-    else:
-        ref_dt = Dataset({var: DataArray(np.asarray(references)) for var in data_vars})
-        plot_collection.map(
-            plot_func, "ref_lines", data=ref_dt, ignore_aes=ref_ignore, **ref_kwargs
-        )
+    ref_dt = _references_to_dataset(references, plot_collection.data, sample_dims=sample_dims)
```
OriolAbril, line 304: To be consistent this should be named `ref_ds` as it is a dataset, not a datatree.

Once we have `ref_ds` I think ideally we'd check if there are any aesthetics for `ref_line_dim` and if there aren't but there is a `ref_line_dim` dimension in `ref_ds` add some. I would default to `overlay` so no properties change between the lines. Doing so will allow users to request different properties for the multiple lines if desired. Pseudocode example:

```
# manually create ref_ds
pc = plot_something...
ref_ds = xr.concat([post.mean(), post.median()], dim="ref_dim_line")
# add aes to ref_dim_line dimension so mean is solid line and median is dashed
add_reference_lines(pc, ref_ds, ref_aes={"linestyle"})
```

This would also mean that there would be no need for the `vlines` and `hlines` functions in the `visuals` module as using `vline` or `hline` directly will work.

Implementation wise, something like:

```python
# add ref_aes and kwargs to function signature
for aes_key in ref_aes:
    pc.update_aes_from_dataset(
        aes_key,
        xr.Dataset(
            {
                var_name: plot_bknd.get_default_aes(aes_key, ref_ds.sizes["ref_line_dim"], kwargs)
                for var_name in ref_ds.data_vars
            }
        )
    )
```

OriolAbril (commented):
Combining the function in arviz-base with `.map` and a vline plotting function allows plotting reference lines quite easily. However, if having the function take care of references in a more general approach it still makes sense to keep it. IMO, this would mean handling line and/or text in one go and also checking for aesthetics mapping for reference lines and adding them if needed so people can request if desired that each reference line has some varying properties.

I have added a few comments along these lines below, let me know if you have any doubts here or on slack.

OriolAbril:
@rohanbabbar04 as we discussed I tried it out and managed to remove the loop over `ref_line_dim`. I also added some docs and kwargs for flexibility in aesthetic mappings. I have to work on something else now but I'll try to add some tests later and merge.

rohanbabbar04:
It would be good if we could merge it. We can add some tests later in another PR.

OriolAbril (approved):

PR #234, opened 2025-05-03
Make example less cluncky

aloctavodia:
I think these lines do not reflect patterns we want to encourage.  So I am removing them.

```python
plot_bknd = import_module(f".backend.{backend}", package="arviz_plots")
color = plot_bknd.get_default_aes("color", 1, {})[0]
```

```python
target = pc.viz["plot"].sel(column="forest").item()
plot_bknd.xlabel("Observations", target)
```

----
📚 Documentation preview 📚: https://arviz-plots--234.org.readthedocs.build/en/234/

### DIFF
--- a/docs/source/gallery/predictive_checks/99_plot_forest_pp_obs.py
+++ b/docs/source/gallery/predictive_checks/99_plot_forest_pp_obs.py
@@ -21,9 +21,9 @@ pc = azp.plot_forest(
     group="posterior_predictive",
     combined=True,
     labels=["obs_dim_0"],
-    backend = "none",  # change to preferred backend
-
+    backend="bokeh",  # change to preferred backend
 )
+
 pc.map(
     azp.visuals.scatter_x,
     "observations",
--- a/docs/source/gallery/predictive_checks/99_plot_forest_pp_obs.py
+++ b/docs/source/gallery/predictive_checks/99_plot_forest_pp_obs.py
@@ -9,17 +9,11 @@ Overlay of forest plot for the posterior predictive samples and the actual obser
 API Documentation: {func}`~arviz_plots.plot_forest`
 :::
 """
-from importlib import import_module
-
 from arviz_base import load_arviz_data
 
 import arviz_plots as azp
 
 azp.style.use("arviz-variat")
-
-backend="none"  # change to preferred backend
-plot_bknd = import_module(f".backend.{backend}", package="arviz_plots")
-color = plot_bknd.get_default_aes("color", 3, {})[-1]
 
 idata = load_arviz_data("non_centered_eight")
 pc = azp.plot_forest(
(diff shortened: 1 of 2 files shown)


docs/source/gallery/predictive_checks/99_plot_forest_pp_obs.py
```diff
@@ -9,33 +9,34 @@
...
-    color=color,
+    color="gray",
+)
+
+pc.map(
+    azp.visuals.labelled_x,
+    "xlabel",
+    coords={"column": "forest"},
+    text="Observations",
+    ignore_aes="y",  # we can omit this in matplotlib, but not bokeh, plotly
```
aloctavodia, line 40: Feature or bug?

OriolAbril:
I think I'd prefer the idea from #235 to replace the import backend-get aes values-get position from it instead of hardcoding the hex code for the color as it would be robust to theme updates or the example using a different theme.

I do agree these patterns are generally not something we want to encourage, especially the first one, things similar to the 2nd will be necessary to handle very specific requests of a plot, and part of the storing everything in the `pc.viz` is to allow this. But the example gallery is not the place for this, maybe we could have dedicated notebooks per backend doing the same very specific plots as a way to show how to get elements from a plot collection and customize them further than what is shown in https://arviz-plots.readthedocs.io/en/latest/tutorials/intro_to_plotcollection.html (or maybe extend that to have the 3 backend versions)

docs/source/gallery/predictive_checks/99_plot_forest_pp_obs.py
OriolAbril: matplotlib being too permissive with kwargs. Not ignoring the `y` aes means a `y=...` kwarg gets passed downstream. As the function sets the xlabel `y=...` is generally not allowed, but matplotlib does allow any kwarg that is valid for [text](https://matplotlib.org/stable/api/text_api.html#matplotlib.text.Text) even if it will be ignored as the position is set in a different way for xlabels.

OriolAbril: Should we allow `ignore_aes="all"` for cases like this? I think for titles, x/ylabels, grids we'll generally want to ignore all aesthetics. The defaults in `aes_map` handle this "natively" for batteries included plots, but using `.map` outside them would probably benefit from that option, I can make it an issue

aloctavodia: Sounds good.

aloctavodia:
I agree the hexcode was just a placeholder, in the meantime, we may use a name like "black" or "gray". 

We should have examples of those especific tweaks in the docs. We could also use EABM to show those kind of examples, currently I think there is only one example (setting xlim) .

OriolAbril (approved):

docs/source/gallery/predictive_checks/99_plot_forest_pp_obs.py
OriolAbril: I removed the comment because it really should not be omitted even if it works in matplotlib. In addition to calling `set_xlabel(..., y=...)` it is being called by every value of `y` available which we really don't need nor want to happen.

Issue #149, opened 2025-02-23
plot_ppc: groupby behaviour


aloctavodia:
Until explicit indexes are supported it won’t be possible to have the groupby-like behaviour

aloctavodia:
@OriolAbril do you know the status of this?

OriolAbril:
We need some changes in arviz-stats (or maybe on how we call it from arviz-plots) and plot_collection but it is implementable on our end only

PR #241, opened 2025-05-05
Explicit index support in PlotCollection.map

### LINKED ISSUE #237: Add `ignore_aes="all"` option in `.map`
See https://github.com/arviz-devs/arviz-plots/pull/234#discussion_r2072407031

### LINKED ISSUE #257: Bayes factor plot is not showing the value of the BF
There is a problem with the legend, most likely a silly mistake I made in #256

OriolAbril:
See the compose your own plot page for proof of concept.

Closes #237 and should unblock #149 (yet not fix it though as we'll still need to handle the kde/ecdf... computation through groupby for that to work completely).

Also closes #257

----
📚 Documentation preview 📚: https://arviz-plots--241.org.readthedocs.build/en/241/


OriolAbril:
I ended up extending the behaviour to coordinate values and we internally call `set_xindex` and use `dim_to_idx` if needed. That being said, I haven't checked how expensive it is but it is definitely better to call set_xindex once, not every time .map is called or a subset of a kwarg requested so we'll need to document that somehow.

Faceting and aes now work from coordinate values. https://arviz-plots--241.org.readthedocs.build/en/241/tutorials/compose_own_plot.html has a demonstration of the behaviour (or will as the rtd build is still in progress as I write this). I have also solved a couple smaller issues in the process, mainly the `fn(da, target, backend, ...)` to `fn(da, target, ...)` and renamed all instances of "chart" to "figure".

Needs more testing and documentation (existing ones have been fixed/updated as necessary), but if there are incoming PRs it might be better to merge and add them in a follow up PR to avoid git conflicts.

OriolAbril:
Legends also need to be fixed to adapt to the new aesthetics logic

PR #242, opened 2025-05-06
Add more variation to distribution examples

aloctavodia:
----
📚 Documentation preview 📚: https://arviz-plots--242.org.readthedocs.build/en/242/


OriolAbril (approved):
This also makes me think the see also section of the example gallery should include a link to the "batteries-included plots intro" page.

PR #246, opened 2025-05-07
Add savefig method to PlotCollection

aloctavodia:
----
📚 Documentation preview 📚: https://arviz-plots--246.org.readthedocs.build/en/246/

### DIFF
--- a/src/arviz_plots/plot_collection.py
+++ b/src/arviz_plots/plot_collection.py
@@ -2,6 +2,7 @@
 """Plot collection class."""
 import warnings
 from importlib import import_module
+from pathlib import Path
 
 import numpy as np
 import xarray as xr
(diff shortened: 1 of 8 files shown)


OriolAbril (commented):
Can you also add something like 

https://github.com/arviz-devs/arviz-plots/blob/b57aa39971e6318d99b5b908482403d13acd2100/src/arviz_plots/backend/none/__init__.py#L103-L110

to get a more informative message if someone attempts to save a "none" backend object?

It will also make things easier when documenting as I can have a general template with the backend api and render one page per backend substituting the backend only

src/arviz_plots/plot_collection.py
```diff
@@ -337,6 +337,13 @@ def show(self):
         else:
             self.viz["plot"].item()
 
+    def savefig(self, *args, **kwargs):
```
OriolAbril, line 340: given the backend versions have signature `savefig(chart, filename, **kwargs)` the *args part can only ever be `filename` so I would match that here.

Also, seeing the tests with a `str(path)`, would it be possible to support both string and path objects as `filename`? I think if we convert the input to `pathlib.Path` we can use `filename.stem == ".html/png"` instead of `.endswith` for checking the extension

OriolAbril (approved):
I'll add some docs and merge

PR #248, opened 2025-05-08
Add style to docs

aloctavodia:
Not sure where the best place to place it

----
📚 Documentation preview 📚: https://arviz-plots--248.org.readthedocs.build/en/248/


OriolAbril (approved):
It's ok for now, we can restructure at any time if we want

PR #254, opened 2025-05-14
Update ref_dim in add_reference_lines

rohanbabbar04:
- Pass ref_dim to references_to_dataset

With the references_to_dataset now updated in this [PR](https://github.com/arviz-devs/arviz-base/pull/53), making the fix to support it here.

----
📚 Documentation preview 📚: https://arviz-plots--254.org.readthedocs.build/en/254/

### DIFF
--- a/src/arviz_plots/plots/utils.py
+++ b/src/arviz_plots/plots/utils.py
@@ -255,7 +255,9 @@ def add_reference_lines(
 
     plot_func = vline if orientation == "vertical" else hline
 
-    ref_ds = references_to_dataset(references, plot_collection.data, sample_dims=sample_dims)
+    ref_ds = references_to_dataset(
+        references, plot_collection.data, sample_dims=sample_dims, ref_dim=ref_dim
+    )
     requested_aes = (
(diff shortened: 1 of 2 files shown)


OriolAbril (commented):
Given the goal is for this to eventually add text too if requested, I would update the default value for `ref_dim` in this function too to match the one in arviz-base. It also looks like I forgot to add it to the docstring

aloctavodia:
While here, what do you think about renaming `add_reference_lines` to `add_lines`?

rohanbabbar04:
> While here, what do you think about renaming `add_reference_lines` to `add_lines`?

I'm fine with that. We actually kept the word 'reference' in mind because we have `ref_dim`, and the function's initial purpose was to plot reference lines. However, ideally, it could also be used to plot other vlines/hlines.

rohanbabbar04:
Should we merge this? We can rename `add_reference_lines` to `add_lines` (if we wish to change it) in another PR, since the tests will fail for other PRs until this goes in

PR #256, opened 2025-05-15
bf_plot support multiple variables

aloctavodia:
This also fixes the failing tests due to the changes in `arviz_stats.bayes_factor`. There is an issue with the legend not showing the actual value of the BF, but I will merge this as soon as the tests pass and open an issue for that to unblock CI.


aloctavodia:
merging as test are passing locally, and error looks unrelated to this PR

PR #258, opened 2025-05-15
Add function for plotting reference bands

### LINKED ISSUE #227: Add function to plot bands
Similarly to `add_reference_lines`, we could have a `add_reference_bands` function. See #223.

rohanbabbar04:
Initial draft for plotting reference bands

Closes #227 

----
📚 Documentation preview 📚: https://arviz-plots--258.org.readthedocs.build/en/258/

### DIFF
--- a/src/arviz_plots/backend/none/__init__.py
+++ b/src/arviz_plots/backend/none/__init__.py
@@ -398,7 +398,7 @@ def hline(y, target, *, color=unset, alpha=unset, width=unset, linestyle=unset, 
     return artist_element
 
 
-def vspan(xmin, xmax, target, color=unset, alpha=unset, **artist_kws):
+def vspan(xmin, xmax, target, *, color=unset, alpha=unset, **artist_kws):
     """Interface to a vertical shaded region spanning the whole axes."""
     kwargs = {"color": color, "alpha": alpha}
     if not ALLOW_KWARGS and artist_kws:
@@ -413,7 +413,7 @@ def vspan(xmin, xmax, target, color=unset, alpha=unset, **artist_kws):
     return artist_element
 
 
-def hspan(ymin, ymax, target, color=unset, alpha=unset, **artist_kws):
+def hspan(ymin, ymax, target, *, color=unset, alpha=unset, **artist_kws):
     """Interface to a horizontal shaded region spanning the whole axes."""
     kwargs = {"color": color, "alpha": alpha}
     if not ALLOW_KWARGS and artist_kws:
--- a/src/arviz_plots/backend/none/__init__.py
+++ b/src/arviz_plots/backend/none/__init__.py
@@ -398,6 +398,36 @@ def hline(y, target, *, color=unset, alpha=unset, width=unset, linestyle=unset, 
     return artist_element
 
 
+def vspan(xmin, xmax, target, color=unset, alpha=unset, **artist_kws):
+    """Interface to a vertical shaded region spanning the whole axes."""
+    kwargs = {"color": color, "alpha": alpha}
+    if not ALLOW_KWARGS and artist_kws:
+        raise ValueError("artist_kws not empty")
+    artist_element = {
+        "function": "vspan",
+        "xmin": xmin,
+        "xmax": xmax,
+        **_filter_kwargs(kwargs, artist_kws),
+    }
+    target.append(artist_element)
+    return artist_element
+
+
+def hspan(ymin, ymax, target, color=unset, alpha=unset, **artist_kws):
+    """Interface to a horizontal shaded region spanning the whole axes."""
+    kwargs = {"color": color, "alpha": alpha}
+    if not ALLOW_KWARGS and artist_kws:
+        raise ValueError("artist_kws not empty")
+    artist_element = {
+        "function": "hspan",
+        "ymin": ymin,
+        "ymax": ymax,
+        **_filter_kwargs(kwargs, artist_kws),
+    }
+    target.append(artist_element)
+    return artist_element
+
+
 def ciliney(
     x,
     y_bottom,
--- a/src/arviz_plots/plots/utils.py
+++ b/src/arviz_plots/plots/utils.py
@@ -373,7 +373,7 @@ def add_reference_bands(
         aes_map = {}
     else:
         aes_map = aes_map.copy()
-    aes_map["ref_band"] = aes_map.get("ref_band", ["overlay"])
+    aes_map["ref_band"] = aes_map.get("ref_band", ["overlay_ref"])
     if sample_dims is None:
         sample_dims = list(set(plot_collection.data.dims).difference(plot_collection.facet_dims))
     if isinstance(sample_dims, str):
--- a/src/arviz_plots/plots/utils.py
+++ b/src/arviz_plots/plots/utils.py
@@ -373,14 +373,13 @@ def add_reference_bands(
         aes_map = {}
     else:
         aes_map = aes_map.copy()
-    aes_map["ref_band"] = aes_map.get("ref_band", ["overlay_ref"])
+    aes_map["ref_band"] = aes_map.get("ref_band", ["overlay_band"])
     if sample_dims is None:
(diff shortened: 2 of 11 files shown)


src/arviz_plots/backend/none/__init__.py
```diff
@@ -398,6 +398,36 @@ def hline(y, target, *, color=unset, alpha=unset, width=unset, linestyle=unset,
     return artist_element
 
 
+def vspan(xmin, xmax, target, color=unset, alpha=unset, **artist_kws):
```
OriolAbril, line 401:
```suggestion
def vspan(xmin, xmax, target, *, color=unset, alpha=unset, **artist_kws):
```

it shouldn't really change anything but this will match the other functions in the module and the equivalent functions in the other backends

src/arviz_plots/backend/none/__init__.py
```diff
@@ -398,6 +398,36 @@ def hline(y, target, *, color=unset, alpha=unset, width=unset, linestyle=unset,
...
+        "function": "vspan",
+        "xmin": xmin,
+        "xmax": xmax,
+        **_filter_kwargs(kwargs, artist_kws),
+    }
+    target.append(artist_element)
+    return artist_element
+
+
+def hspan(ymin, ymax, target, color=unset, alpha=unset, **artist_kws):
```
OriolAbril, line 416:
```suggestion
def hspan(ymin, ymax, target, *, color=unset, alpha=unset, **artist_kws):
```

src/arviz_plots/plots/utils.py
```diff
@@ -295,3 +295,124 @@ def add_reference_lines(
...
+        aes_map = {}
+    else:
+        aes_map = aes_map.copy()
+    aes_map["ref_band"] = aes_map.get("ref_band", ["overlay_ref"])
+    if sample_dims is None:
+        sample_dims = list(set(plot_collection.data.dims).difference(plot_collection.facet_dims))
+    if isinstance(sample_dims, str):
+        sample_dims = [sample_dims]
+    if ref_dim is None:
+        ref_dim = ["ref_dim_0", "ref_dim_1"]
```
OriolAbril, line 382: As this will later on be passed as input to `references_to_dataset` and some of them will also show in the artist/aesthetic mappings generated, I think it would be better to use more meaningful names (even if they were longer). Here for example, if we consider each band to be a "single reference" these would be `ref_dim, band_dim`.

OriolAbril (commented):
I think we could also add the `add_xyz` functions to be considered functions worthy of the minigallery. If you are interested in playing with this on the [gallery_generator](https://github.com/arviz-devs/arviz-plots/blob/main/docs/sphinxext/gallery_generator.py) let me know and I can help, otherwise we'll open an issue to handle it later on.

rohanbabbar04:
> I think we could also add the `add_xyz` functions to be considered functions worthy of the minigallery. If you are interested in playing with this on the [gallery_generator](https://github.com/arviz-devs/arviz-plots/blob/main/docs/sphinxext/gallery_generator.py) let me know and I can help, otherwise we'll open an issue to handle it later on.

Hmm...
Sounds good. Should we do it in a new PR? I think it would be better to keep the `add_reference_bands` in this PR and get it merged. I can work on the minigallery in a separate PR.

src/arviz_plots/plots/utils.py
```diff
@@ -295,3 +295,113 @@ def add_reference_lines(
...
+        The orientation of the reference lines, either "vertical" or "horizontal".
+    aes_map : mapping of {str : sequence of str}, optional
+        Mapping of artists to aesthetics that should use their mapping in `plot_collection`
+        when plotted. Valid keys are the same as for `plot_kwargs`.
+
+        The default is to use an "overlay_band" aesthetic for all elements.
+
+        It is possible to request aesthetics without mappings defined in the
+        provided `plot_collection`. In those cases, a mapping of "ref_dim" to the requested
+        aesthetic will be automatically added.
```
OriolAbril, line 332:
```suggestion
        It is possible to request aesthetics without mappings defined in the
        provided `plot_collection`. In those cases, a mapping of the dimensions in
        `ref_dim` minus its last element to the requested aesthetic will be
        automatically added.
```

src/arviz_plots/plots/utils.py
```diff
@@ -295,3 +295,113 @@ def add_reference_lines(
...
+    plot_kwargs : mapping of {str : mapping or False}, optional
+        Valid keys are:
+
+        * "ref_band" -> passed to :func:`~arviz_plots.visuals.vspan` for vertical `orientation`
+          and to :func:`~arviz_plots.visuals.hspan` for horizontal `orientation`
```
OriolAbril, line 337: can you also add `hspan` and `vspan` to https://github.com/arviz-devs/arviz-plots/blob/main/docs/source/api/visuals.rst. This will make these cross-references work and be rendered as links

src/arviz_plots/plots/utils.py
```diff
@@ -295,3 +295,113 @@ def add_reference_lines(
...
+
+    sample_dims : list, optional
+        Dimensions to reduce unless mapped to an aesthetic.
+        Defaults to ``rcParams["data.sample_dims"]``
```
OriolAbril, line 341:
```suggestion
        Dimensions that should not be added to the Dataset generated from
        `refereces` via :func:`arviz_base.references_to_dataset`.
        Defaults to all dimensions in ``plot_collection.data`` that are not ``facet_dims``
```

That description is true for `plot_xyz` functions but not here. You can use this description instead (same as add_reference_lines)

src/arviz_plots/plots/utils.py
```diff
@@ -295,3 +295,113 @@ def add_reference_lines(
...
+    ref_dim : list, optional
+        Specifies the name of the reference dimension for reference values.
+    **kwargs : mapping of {str : sequence}, optional
```
OriolAbril, line 344:
```suggestion
    **kwargs : sequence, optional
```

This one I recently learned I was doing wrong. The fact it is `**kwargs` and not `kwargs` means it must be a mapping with strings as keys, so in such cases only the type of the values should be documented.

src/arviz_plots/plots/utils.py
```diff
@@ -295,3 +295,113 @@ def add_reference_lines(
...
+        Valid keys are:
+
+        * "ref_band" -> passed to :func:`~arviz_plots.visuals.vspan` for vertical `orientation`
+          and to :func:`~arviz_plots.visuals.hspan` for horizontal `orientation`
+
+    sample_dims : list, optional
+        Dimensions to reduce unless mapped to an aesthetic.
+        Defaults to ``rcParams["data.sample_dims"]``
+    ref_dim : list, optional
+        Specifies the name of the reference dimension for reference values.
```
OriolAbril, line 343: I would also update this to match the behaviour of this function.

src/arviz_plots/plots/utils.py
```diff
@@ -295,3 +295,113 @@ def add_reference_lines(
...
+        ref_dim = ["ref_dim", "band_dim"]
+
+    plot_func = vspan if orientation == "vertical" else hspan
+
+    ref_ds = references_to_dataset(
+        references, plot_collection.data, sample_dims=sample_dims, ref_dim=ref_dim
+    )
+
+    requested_aes = set(aes_map["ref_band"]).difference(plot_collection.aes_set)
+    *ref_dim, _band_dim = ref_dim
```
OriolAbril, line 391: Not sure why pylint is not complaining, maybe the leading underscore. But I think we should do one of the following.

* use `*ref_dim, _` to not store the last dimension as anything
* use `*ref_dim, band_dim` and then check that `band_dim` has length 2 and raise an informative error otherwise (as I think the visual would break or behave weirdly in such case).

rohanbabbar04: Have added a check for this

src/arviz_plots/plots/utils.py
rohanbabbar04: Done

PR #264, opened 2025-05-20
rename add_reference_lines and reference_bands

aloctavodia:
This also tweaks the docstrings and examples a little bit. 

If you see the example with `plot_forest`, you will see that the band is incorrectly added to the two columns, not just the `forest` one. Not sure how to fix this.

----
📚 Documentation preview 📚: https://arviz-plots--264.org.readthedocs.build/en/264/

### DIFF
--- a/docs/source/gallery/utils/01_add_reference_bands.py
+++ b/docs/source/gallery/utils/01_add_reference_bands.py
@@ -26,8 +26,7 @@ pc.coords = {"column": "forest"}
 pc = azp.add_bands(
     pc,
     values=rope,
-    aes_map={"ref_band": ["color"]},
-    color=["#f66d7f"]
+    plot_kwargs={"ref_band":{"color": "#f66d7f"}},
 )
 
 pc.show()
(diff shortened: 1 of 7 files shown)


aloctavodia:
@rohanbabbar04

aloctavodia:
Yeah, I also think is weird, but I was not able to think of a better name

docs/source/gallery/utils/01_add_reference_bands.py
```diff
@@ -17,18 +17,17 @@
...
-
-pc = azp.add_reference_bands(
+pc.coords = {"column": "forest"}
+pc = azp.add_bands(
     pc,
-    references=references,
+    values=rope,
     aes_map={"ref_band": ["color"]},
-    color=["black", "gray"]
+    color=["#f66d7f"]
```
OriolAbril, line 30: doesn't using `plot_kwargs` work? It would be important to know why not if it doesn't, and conceptually I think it is what we want here (even if 1 element aes mapping achieves the same result)

aloctavodia: yes, it does.

rohanbabbar04 (approved):
Looks good!!
I think the same question as you guys have already discussed with the names `ref_line` and `ref_band` but otherwise it looks fine.

PR #267, opened 2025-05-25
fixes column name from elpd to elpd_loo in cmp_df dataframe

The-Broken-Keyboard:
In `plot_compare` it expects a dataframe ( result of compare function of arviz ), then it checks for `elpd` column. 

But the latest version of returning dataframe contains `elpd_loo` column instead of `elpd`. So this causes issue while using `plot_compare`

### DIFF
--- a/src/arviz_plots/plots/compare_plot.py
+++ b/src/arviz_plots/plots/compare_plot.py
@@ -68,7 +68,7 @@ def plot_compare(
     """
     # Check if cmp_df contains the required information
     column_index = [c.lower() for c in cmp_df.columns]
-    if "elpd" not in column_index:
+    if "elpd_loo" not in column_index:
         raise ValueError("cmp_df must have been created using the `compare` function from ArviZ.")
 
     # Set default backend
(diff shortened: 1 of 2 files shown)


src/arviz_plots/plots/compare_plot.py
```diff
@@ -67,8 +67,16 @@ def plot_compare(
         arXiv preprint https://arxiv.org/abs/1507.02646
     """
     # Check if cmp_df contains the required information
+
     column_index = [c.lower() for c in cmp_df.columns]
-    if "elpd_loo" not in column_index:
+
+    elpd_method = "elpd" + "_" + rcParams["stats.information_criterion"].lower()
```
aloctavodia, line 73: We are dropping support for waic. We are only going to support loo.

aloctavodia:
This works


import arviz_base as azb
import arviz_stats as azs
import arviz_plots as azp
model_compare = azs.compare({'Centered': azb.load_arviz_data('centered_eight'),
                 'Non-centered': azb.load_arviz_data('non_centered_eight')})

azp.plot_compare(model_compare)

The-Broken-Keyboard:
I have  updated the docstring and error statement. So that users will know that they have to use compare function from arviz-stats instead of arviz

PR #276, opened 2025-06-04
pair focus plot

The-Broken-Keyboard:
Adds pair focus plot


src/arviz_plots/plots/pair_focus_plot.py
```diff
@@ -0,0 +1,236 @@
...
+            "cols", ["__variable__"] + [dim for dim in distribution.dims if dim not in sample_dims]
+        )
+        n_plots, _ = process_facet_dims(distribution, pc_kwargs["cols"])
+        col_wrap = pc_kwargs["col_wrap"]
+        if n_plots <= col_wrap:
+            n_rows, n_cols = 1, n_plots
+        else:
+            div_mod = divmod(n_plots, col_wrap)
+            n_rows = div_mod[0] + (div_mod[1] != 0)
+            n_cols = col_wrap
```
aloctavodia, line 142: here we can use set_wrap_layout, right?

src/arviz_plots/plots/pair_focus_plot.py
```diff
@@ -0,0 +1,236 @@
...
+        pc_kwargs["figure_kwargs"] = pc_kwargs.get("figure_kwargs", {}).copy()
+        if "figsize" not in pc_kwargs["figure_kwargs"]:
+            pc_kwargs["figure_kwargs"]["figsize"] = figsize
+            pc_kwargs["figure_kwargs"]["figsize_units"] = figsize_units
+        pc_kwargs["figure_kwargs"].setdefault("sharex", True)
+        plot_collection = PlotCollection.wrap(
+            distribution,
+            backend=backend,
+            **pc_kwargs,
+        )
```
aloctavodia, line 170: can we group the two "if plot_collection is None"?

aloctavodia (changes requested):

aloctavodia:
I added a few changes, including 

* Making the two examples more different, so we get the chance to show more features in the gallery. 
* If var_names is not explicit, then we negate `focus_var`, as the comparison of a variable with itself is trivial. 
* Share the y-axis by default, not the x-axis. As the y-axis is the constant one. Sharing the x-axis produces nicer figures when the scales are similar, but when the scales are very different, it may generate default figures that are difficult to read
* I removed the border of the dots, as having a border becomes visually challenging when combined with a value of alpha < 1 for the face of the marker.

Something still pending is to remove the title and instead have y_labels and x_labels with the names of the variables. Otherwise is not clear from the figure alone what we are seeing. 

After that, we can merge. Unless someone else has more suggestions.

The-Broken-Keyboard:
removed sharex: True from `Scatterplot one variable against all others` in gallery, because sharex: true will not allow us to put different xlabels to different plots in plotly. Hence to put xlabels in all plots we would need to remove sharex

aloctavodia:
The last change is hiding an issue. It should be possible to set `sharex=True`

aloctavodia:
> @aloctavodia just for my clarification, why did you convert `tau` to `log_tau` while plotting for inference diagnostics?

When checking for divergences, it's common to plot the variables on the scale on which they were sampled. See here for example https://www.pymc.io/projects/examples/en/latest/diagnostics_and_criticism/Diagnosing_biased_Inference_with_Divergences.html internally packages like PyMC/Stan when using samplers like "HMC/NUTS" transform bounded variables (like a Beta distribution that is restricted to be in [0, 1]) to be unbounded (R). This makes sampling easier. I thought that the example could reflect that common practice.

The-Broken-Keyboard:
@aloctavodia but logically here the x axis is not shared right? because on x axis we have different variables for each plot.

In #204 as we discussed, for the plots with xaxis as True we are not putting x labels for upper row plots ( we are just putting x labels for the plots in last row ). This we did to resolve x label conflicts with title of plots in plotly. This is inherent issue because plotly doesn't allow us to reserve space for x labels while creating subplots. So when we put x labels after subplots are created then it overlaps the titles of lower row plots.

Now if we would allow sharex as True here  ( even when x axes are different variables ) , then the discussion in #204 and PR  #208 are not valid anymore. What should we do then?

The-Broken-Keyboard:
in PR #208 we assumed that when x axis will be shared then x labels also will be same. So we removed xlabels for plots in upper rows and just kept xlabel for last row. This assumption won't align with current situation in this plot. Here the xlabels can't be same because variables on x axis in each plot is different

aloctavodia:
There are cases when we want to share the x-axis and the xlabels, and cases when we just want to share the x-axis.

The-Broken-Keyboard:
> It already works for matplotlib and bokeh, the only issue is with plotly. We should fix that, in a separated PR.

OKay cool, I am setting sharex as true again. We will solve the share axis and share label issue in separate PR

aloctavodia (approved):

src/arviz_plots/plots/pairs_focus_plot.py
```diff
@@ -0,0 +1,222 @@
...
+    """Plot a fixed variable against other variables in the dataset.
+
+    Parameters
+    ----------
+    dt : DataTree
+        Input data
+    var_names: str or list of str, optional
+        One or more variables to be plotted.
+        Prefix the variables by ~ when you want to exclude them from the plot.
+    filter_vars: {None, “like”, “regex”}, optional, default=None
```
OriolAbril, line 44: `, optional` and `, default None` should not be used together. If the default value in the function signature is used it should be `default`. If the default value in the function signature is just a placeholder (i.e. `var_name=None` means all variables, `coords=None` means empty dictionary/no subset) it should be `optional`.

For extra context, we use numpydoc: https://numpydoc.readthedocs.io/en/latest/format.html which says:

> If it is not necessary to specify a keyword argument, use `optional`:
> Optional keyword parameters have default values, which are displayed as part of the function signature. [...]
> or as part of the type, instead of optional. If the default value would not be used as a value, optional is preferred. These are all equivalent:
>
>     copy : bool, default True
>     copy : bool, default=True
>     copy : bool, default: True

We always use the first one to keep the docstrings consistent

src/arviz_plots/plots/pairs_focus_plot.py
```diff
@@ -0,0 +1,222 @@
...
+        If None (default), interpret var_names as the real variables names.
+        If “like”, interpret var_names as substrings of the real variables names.
+        If “regex”, interpret var_names as regular expressions on the real variables names.
+    group : str, optional
```
OriolAbril, line 48: consequently, `default "posterior"`. Also remember that you don't have to think much for all these arguments that are common accross most functions. https://arviz-plots.readthedocs.io/en/latest/contributing/new_plot.html#call-signature has a dropdown "Function signature and docstring template" that you can copy to get started and adapt to your function

src/arviz_plots/plots/pairs_focus_plot.py
```diff
@@ -0,0 +1,222 @@
...
+        Group to use for plotting. Defaults to "posterior".
+    coords : mapping, optional
+        Coordinates to use for plotting. Defaults to None.
+    sample_dims : iterable, optional
+        Dimensions to reduce unless mapped to an aesthetic.
+        Defaults to ``rcParams["data.sample_dims"]``
+    focus_var: str
```
OriolAbril, line 55: As of now, this is a required argument, but it is treated as optional for everything except lacking the `, optional` part. I would either move `focus_var` and `focus_var_coords` to the top, with `focus_var` being a proper required argument or define a default for it (e.g. the 1d slice in the dataset). This diverges a bit from https://arviz-plots.readthedocs.io/en/latest/contributing/new_plot.html but the only other function with 2 required arguments is [`combine_plots`](https://arviz-plots.readthedocs.io/en/latest/api/generated/arviz_plots.combine_plots.html) where it comes before `var_names`, `filter_vars`...

Moreover, I would like to allow both string or DataArray as input. It is quite easy to add a new variable to the input data then use that so I am not very opinionated on that but I think it would be nice and feel natural.

src/arviz_plots/plots/pairs_focus_plot.py
```diff
@@ -0,0 +1,222 @@
...
+        Name of the variable to be plotted against all other variables.
+    focus_var_coords : mapping, optional
+        Coordinates to use for the target variable. Defaults to None.
+    plot_collection : PlotCollection, optional
+    backend : {"matplotlib", "bokeh","plotly"}, optional
```
OriolAbril, line 60:
```
backend : {"matplotlib", "bokeh", "plotly", "none"}, optional
        Plotting backend to use. Defaults to ``rcParams["plot.backend"]``
```

Also available in the template I mentioned before.

src/arviz_plots/plots/pairs_focus_plot.py
```diff
@@ -0,0 +1,222 @@
...
+    labeller : labeller, optional
+    aes_by_visuals : mapping, optional
+        Mapping of visuals to aesthetics that should use their mapping in `plot_collection`
+        when plotted. Valid keys are the same as for `visuals`.
```
OriolAbril, line 64: I would add a 2nd paragraph saying by default there are no aesthetic mappings at all. Could be something similar to https://arviz-plots.readthedocs.io/en/latest/api/generated/arviz_plots.plot_dist.html without the `model` related part but it doesn't need to be as extensive

src/arviz_plots/plots/pairs_focus_plot.py
```diff
@@ -0,0 +1,222 @@
...
+    visuals : mapping of {str : mapping or False}, optional
+        Valid keys are:
+
+        * scatter -> passed to :func:`~.visuals.scatter_x`
+        * divergence -> passed to :func:`~.visuals.divergence_scatter`. Defaults to False.
+        * title -> :func:`~.visuals.labelled_title`
+
+    pc_kwargs : mapping
```
OriolAbril, line 72: This one we have to adapt in the template now that we have changed the signature but it should be `**pc_kwargs` with the asterisks and no type info

src/arviz_plots/plots/pairs_focus_plot.py
```diff
@@ -0,0 +1,222 @@
...
+        Passed to :class:`arviz_plots.PlotCollection`
```
OriolAbril, line 73: `` Passed to :meth:`arviz_plots.PlotCollection.wrap` ``

src/arviz_plots/plots/pairs_focus_plot.py
```diff
@@ -0,0 +1,222 @@
...
+
+        >>> from arviz_plots import plot_pairs_focus, style
+        >>> style.use("arviz-variat")
+        >>> from arviz_base import load_arviz_data
+        >>> dt = load_arviz_data('centered_eight')
+        >>> plot_pairs_focus(
+        >>>     dt,
+        >>>     var_names=["theta", "tau"],
+        >>>     focus_var="mu",
+        >>> )
```
OriolAbril, line 94: we have the exact same example in the example gallery, so it should not be repeated here too. We could for example show how to use `focus_coords`

src/arviz_plots/plots/pairs_focus_plot.py
```diff
@@ -0,0 +1,222 @@
...
+        aes_by_visuals = aes_by_visuals.copy()
+
+    if backend is None:
+        if plot_collection is None:
+            backend = rcParams["plot.backend"]
+        else:
+            backend = plot_collection.backend
+
+    if var_names is None:
+        var_names = "~" + focus_var
```
OriolAbril, line 120: This doesn't really work. If I set `focus_var="theta", focus_coords={"school": "Choate"}` we would end up with only the comparisons against `mu` and `theta`. This also means that if I provide both var_names and focus_var I will end up with the comparison of one variable with itself.

Ideally we'd have a helper function that got something like `distribution`, `focus_var` and `focus_coords` and would return `distribution` without the focus_var subset and the dataarray for `focus_var` subset.

If that proves too difficult we could keep something similar but then we'd have to at least check if `focus_var` is multidimensional before negating it. I think it is better to show the pairplot of a variable with itself too often than to remove variables plots from the figure too eagerly.

src/arviz_plots/plots/pairs_focus_plot.py
```diff
@@ -0,0 +1,222 @@
...
+    labeller : labeller, optional
+    aes_by_visuals : mapping, optional
+        Mapping of visuals to aesthetics that should use their mapping in `plot_collection`
+        when plotted. Valid keys are the same as for `visuals`.
+    visuals : mapping of {str : mapping or False}, optional
+        Valid keys are:
+
+        * scatter -> passed to :func:`~.visuals.scatter_x`
+        * divergence -> passed to :func:`~.visuals.divergence_scatter`. Defaults to False.
+        * title -> :func:`~.visuals.labelled_title`
```
OriolAbril, line 70: This is not correct.

src/arviz_plots/plots/pairs_focus_plot.py
```diff
@@ -0,0 +1,222 @@
...
+    scatter_kwargs.setdefault("width", 0)
+    _, _, scatter_ignore = filter_aes(plot_collection, aes_by_visuals, "scatter", sample_dims)
+
+    plot_collection.map(
+        scatter_x,
+        "scatter",
+        ignore_aes=scatter_ignore,
+        y=y,
+        **scatter_kwargs,
+    )
```
OriolAbril, line 164: It probably feels a bit pointless right now, especially in this function that currently has very little elements, but it will get more elements at some point, 2d KDEs (countour only or filled) and hexbin plots, so we might want to show some combination of elements and not some others.

Consequently, this should follow the pattern in https://arviz-plots.readthedocs.io/en/latest/contributing/new_plot.html#adding-visuals-to-the-plot.

OriolAbril (commented):
I left some after the fact comments. I wasn't quick enough to review before, but there is still work to do on this function. In addition to all the comments, it needs tests too.

src/arviz_plots/plots/pairs_focus_plot.py
The-Broken-Keyboard: @aloctavodia , implementation would work because we are passing `focus_var_coords` which will handle multidimensions as well. for example: we can plot scatter plot of `theta-choate` against other `var-coord` combinations.

PR #279, opened 2025-06-06
Fixes share axis and share label feature in plotly

The-Broken-Keyboard:
As discussed in PR #276 , for plotly we need to find a way to enable users to share just x axis , as well as share both x axis and x label. Plotly and the other two backend libraries share just axis by default, but this causes problem in plotly because x label conflicts with the title of lower row plots. So we were forced to  share label also whenever sharex is true in plotly. 

But now , since we encountered the cases where we would like to share just the axis and not label in plotly. So this PR implements a work around which just affects the behaviour of plotly backend and behaviour of other two backends remain same. 

The user can now pass a new key named `share` with `False` value as `visuals = { "xlabel" : {"share": False}}`, which will enable user to not share the xlabels. 

For example:

```python
import arviz_plots as azp

azp.style.use("arviz-variat")
idata = load_arviz_data("centered_eight")
focus_var = "mu"
visuals = {"divergence": True}
pc = azp.plot_pairs_focus(
    idata,
    var_names=["theta","tau"],
    focus_var=focus_var,
    backend="plotly",
    visuals = visuals,
    figure_kwargs={"sharex": True}
)
pc.show()

```

the above code gives output:

image

Both axis and label is shared

and code with `share` as `False` :
```python
visuals = {"divergence": True , "xlabel": {"share": False}}
pc = azp.plot_pairs_focus(
    idata,
    var_names=["theta","tau"],
    focus_var=focus_var,
    backend="plotly",
    visuals = visuals,
    figure_kwargs={"sharex": True}
)
pc.show()
```

gives output as:

image

here just the axis is shared, not label.

### DIFF
--- a/src/arviz_plots/backend/bokeh/__init__.py
+++ b/src/arviz_plots/backend/bokeh/__init__.py
@@ -588,8 +588,9 @@ def ylabel(string, target, *, size=unset, color=unset, **artist_kws):
         setattr(target.yaxis, f"axis_label_{key}", value)
 
 
-def xlabel(string, target, *, size=unset, color=unset, **artist_kws):
+def xlabel(string, target, *, size=unset, share=True, color=unset, **artist_kws):
     """Interface to bokeh for adding a label to the x axis."""
+    _ = share  # noqa: F841
     kwargs = {"text_font_size": _float_or_str_size(size), "text_color": color}
     target.xaxis.axis_label = string
     for key, value in _filter_kwargs(kwargs, artist_kws).items():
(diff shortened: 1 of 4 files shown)


aloctavodia:
Ideally the API should be consistent across backends. This breaks that. A workaround will be to handle this internally for for plotly per plot, or keep thinking how to find a solution for plotly that mimics the other backends.

The-Broken-Keyboard:
@aloctavodia , problem is that plotly already behaves like other backends. It's just that xlabel overlaps with titles in plotly which creates problem in other plots in gallery. So we were forced to change this for plotly. 

In any case if we want to handle it internally in just plotly then we would have to tell plotly that we want to share just axis or we want to share label also with axis through some way. 

One more solution is that for every backend we can enforce use of `share` variable to decide if we want to share label also with axis ( when sharex is true) or not.

The-Broken-Keyboard:
Since, this PR is useless...so closing it.

PR #286, opened 2025-06-12
Minor fixes to tutorials

aloctavodia:
* I found issues running the `compose_own_plot` tutorial can you try to run it again @OriolAbril?
* For bokeh, we provide an automatic color cycler that takes theme colors into account.  https://github.com/arviz-devs/arviz-plots/blob/main/src/arviz_plots/backend/bokeh/__init__.py#L109-L120


OriolAbril:
> For bokeh, we provide an automatic color cycler that takes theme colors into account. https://github.com/arviz-devs/arviz-plots/blob/main/src/arviz_plots/backend/bokeh/__init__.py#L109-L120

The comment makes reference to something else entirely. There are no aesthetic mappings and we don't pass color explicitly either. We then add multiple elements to the same plot or figure. Each backend behaves differently (and has nothing to do with `arviz-plots`). This is actually the reason why we always set the default color from that `get_default_aes` function when it has no mapping. Otherwise the results would be backend dependent. I will try to generate some examples in a bit

OriolAbril:
> I found issues running the compose_own_plot tutorial can you try to run it again @OriolAbril?

That is quite weird, I made sure to rerun from scratch on a restarted kernel right before pushing. I will rerun it

docs/source/tutorials/plots_intro.ipynb
```diff
@@ -8,10 +8,9 @@
     "(plots_intro)=\n",
```
OriolAbril, line 8: <p>Re overlay, it would work the same if we used "overlay_school" instead. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/arviz-devs/arviz-plots/pull/286/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='36'/>

OriolAbril (approved):

Issue #100, opened 2024-10-18
Test common interface functions for each backend independently


OriolAbril:
We should have tests that check the use of common interface plotting functions work as expected. I think the easiest way to go about this is having kind of equivalent tests in backend specific files, so we check that create plotting grid generates the desired grid, shares axis correctly, sets the correct width/height ratios by inspecting the generated output, which will be backend specific.

EDIT: We now have the infrastructure in place to add such tests. See https://github.com/arviz-devs/arviz-plots/pull/409 for some specific examples. The general idea is to add two tests per function:

One using only the required arguments and focusing asserts on returned type and default behaviour (zorder for matplotlib, showlegend in plotly...).

The second one using [common interface arguments](https://python.arviz.org/projects/plots/en/latest/api/backend/index.html#module-arviz_plots.backend) and focusing asserts on ensuring those have been properly translated into the respective backend properties.

There will be some more complex functions that might need more tests, but I would treat those as exceptions.

OriolAbril:
Great, thanks! I will open a PR today to set the template and you can then build on top of that

PR #287, opened 2025-06-13
Adds Pair plot

The-Broken-Keyboard:
Adds Pair Plot

**Key Change required in PlotMatrix:**
When we send mask through `map_triangle` ( for divergence ), then it doesn't perform subsetting on `mask` data. Which causes dimension conflict. It was directly adding `**kwargs` to `fun_kwargs`. Now it performs subsetting before adding few keys (  dataArrays or datasets ) of `kwargs` into `fun_kwargs`.

for the following code:

```python

import arviz_plots as azp

azp.style.use("arviz-variat")

data = load_arviz_data("centered_eight")
visuals = {"divergence": True}
pc = azp.plot_pair(
    data,
    backend="bokeh" , 
    var_names=["mu","tau", "theta"],
    coords={"school": ["Choate", "Deerfield"]},
    visuals=visuals,
)
pc.show()
```

we get the following plot:

image

----
📚 Documentation preview 📚: https://arviz-plots--287.org.readthedocs.build/en/287/

### DIFF
--- a/src/arviz_plots/plots/pair_plot.py
+++ b/src/arviz_plots/plots/pair_plot.py
@@ -65,6 +65,7 @@ def plot_pair(
             "label",
             "xlabel",
             "ylabel",
+            "remove_axis",
         ],
         Mapping[str, Any] | Literal[False],
     ] = None,
@@ -148,7 +149,7 @@ def plot_pair(
           :meth:`arviz_plots.PlotMatrix.map_col` method. Not applied if ``triangle`` is
           "upper" or ``marginal=False``.
 
-        * default_remove_axis -> not passed anywhere.
+        * remove_axis -> not passed anywhere.
 
           It can only be set to ``False`` to disable the default removal of ``x`` and ``y`` axes
           from the plots of other half triangle. If ``triangle`` is "upper" then the lower triangle
@@ -204,10 +205,6 @@ def plot_pair(
     .. plot::
         :context: close-figs
 
-        >>> from arviz_plots import plot_pair, style
-        >>> style.use("arviz-variat")
-        >>> from arviz_base import load_arviz_data
-        >>> dt = load_arviz_data('centered_eight')
         >>> visuals = {"credible_interval":{"color":"red"},"point_estimate":{"color":"red"}}
         >>> plot_pair(
         >>>     dt,
@@ -226,10 +223,6 @@ def plot_pair(
     .. plot::
         :context: close-figs
 
-        >>> from arviz_plots import plot_pair, style
-        >>> style.use("arviz-variat")
-        >>> from arviz_base import load_arviz_data
-        >>> dt = load_arviz_data('centered_eight')
         >>> plot_pair(
         >>>     dt,
         >>>     coords = {"school":"Choate"},
@@ -497,28 +490,28 @@ def plot_pair(
                 **set_ticklabel_visibility_kwargs,
             )
 
-    # default removal of axis for better visualization
-    default_remove_axis_bool = visuals.get("default_remove_axis", True)
-    if default_remove_axis_bool:
-        _, _, default_remove_axis_ignore = filter_aes(
-            plot_matrix, aes_by_visuals, "default_remove_axis", sample_dims
+    #  removal of axis for better visualization
+    remove_axis_bool = visuals.get("remove_axis", True)
+    if remove_axis_bool:
+        _, _, remove_axis_ignore = filter_aes(
+            plot_matrix, aes_by_visuals, "remove_axis", sample_dims
         )
         # if triangle="upper" then remove the lower triangle axes
         if triangle == "upper":
             plot_matrix.map_triangle(
                 remove_matrix_axis,
-                "default_remove_axis",
+                "remove_axis",
                 triangle="lower",
                 axis="both",
-                ignore_aes=default_remove_axis_ignore,
+                ignore_aes=remove_axis_ignore,
             )
         # if triangle="lower" then remove the upper triangle axes
         elif triangle == "lower":
             plot_matrix.map_triangle(
                 remove_matrix_axis,
-                "default_remove_axis",
+                "remove_axis",
                 triangle="upper",
                 axis="both",
-                ignore_aes=default_remove_axis_ignore,
+                ignore_aes=remove_axis_ignore,
             )
     return plot_matrix


The-Broken-Keyboard:
@aloctavodia @OriolAbril please review this. I will add documentation and tests once we are done with final version of function

src/arviz_plots/plot_matrix.py
```diff
@@ -384,7 +384,24 @@ def map_triangle(
...
+                        **{
+                            key + "_x": process_kwargs_subset(values, var_name_x, sel_x)
+                            for key, values in kwargs.items()
+                            if isinstance(values, (xr.DataArray, xr.Dataset))
+                        },
+                        **{
+                            key + "_y": process_kwargs_subset(values, var_name_y, sel_y)
+                            for key, values in kwargs.items()
+                            if isinstance(values, (xr.DataArray, xr.Dataset))
+                        },
```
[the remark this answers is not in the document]
OriolAbril: I am a bit more sure after having gone over all the code in the PR. I'll still try to double check at some point today

src/arviz_plots/plots/pair_plot.py
```diff
@@ -0,0 +1,222 @@
...
+    aes_by_visuals : mapping, optional
+        Mapping of visuals to aesthetics that should use their mapping in `plot_matrix`
+        when plotted. Valid keys are the same as for `visuals`.
+        By default, there are no aesthetic mappings at all
+    visuals : mapping of {str : mapping or False}, optional
+        Valid keys are:
+
+        * scatter -> passed to :func:`~.visuals.scatter_couple`
+        * divergence -> passed to :func:`~.visuals.scatter_couple`. Defaults to False.
+        * label -> :func:`~.visuals.label_plot`
```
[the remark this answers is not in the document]
aloctavodia: ohh, I assumed that was because of some limitation of the current PlotMatrix implementation. We definitely want to show axes by default.

The-Broken-Keyboard:
@OriolAbril @aloctavodia 

I have incorporated all the changes suggested above

**MAJOR CHANGES**
- Added a `map_row_col` method which takes a variable, selection dict and orientation (row or col) and then maps the given function to whole row or col of matrix. Since we have this method in PlotMatrix, so we can also include similar kind of method  in PlotCollection and the work around implemented in #208 will not be needed now, because we can use this function in plotcollection to handle conflicting xlabels issue too. So, i removed the work around of #208 for now ( to incorporate this new function ). 
- Added parameters `triangle` and `remove_axis` ( in visuals ) .
- Added marginals ( supports kind="kde","hist" and "ecdf") on diagonal. 
- Also added diagonal xlabels if `triangle="upper"` is given
- Diagonal labels will be used automatically if `marginal=False`
- x labels for bottom plots and y labels for left plots will be used only if `marginal=True`, because otherwise diagonal labels will be there to label the rows and cols(automatically, as mentioned above).
- **Enforcing `remove_axis=False` if `marginal=True`. Because if we have marginals we definitely need xlabels and ylabels to label rows and cols ( since diagonal labels won't be there and removing axis will remove axis labels too).
- `sharex` and `sharey` is set to `True`. If `marginal=True` then we can't have `sharey=True`, so that is forced to be `False`.

** We are getting warning in Bokeh whenever we use upper or lower triangle, it says renderer wasn't found for some plots**

Examples:

### Code 1

```python

import arviz_plots as azp

azp.style.use("arviz-variat")

data = load_arviz_data("centered_eight")
pc = azp.plot_pair(
    data,
    backend="bokeh" , 
    var_names=["mu","tau","theta"],
    coords={"school": ["Choate", "Deerfield"]},
    marginal=True,
    marginal_kind="kde",
    triangle="both",
    visuals={"divergence": True},
)
pc.show()
```
### result 1
image

### code 2

```python

import arviz_plots as azp

azp.style.use("arviz-variat")

data = load_arviz_data("centered_eight")
pc = azp.plot_pair(
    data,
    backend="bokeh" , 
    var_names=["mu","tau","theta"],
    coords={"school": ["Choate", "Deerfield"]},
    marginal=True,
    marginal_kind="hist",
    triangle="both",
    visuals={"divergence": True},
)
pc.show()
```

### result 2
image

### code 3

```python

import arviz_plots as azp

azp.style.use("arviz-variat")

data = load_arviz_data("centered_eight")
pc = azp.plot_pair(
    data,
    backend="bokeh" , 
    var_names=["mu","tau","theta"],
    coords={"school": ["Choate", "Deerfield"]},
    marginal=True,
    marginal_kind="ecdf",
    triangle="both",
    visuals={"divergence": True},
)
pc.show()
```

### result 3
image

### code 4

```python

import arviz_plots as azp

azp.style.use("arviz-variat")

data = load_arviz_data("centered_eight")
pc = azp.plot_pair(
    data,
    backend="bokeh" , 
    var_names=["mu","tau","theta"],
    coords={"school": ["Choate", "Deerfield"]},
    marginal=True,
    marginal_kind="kde",
    triangle="upper",
    visuals={"divergence": True},
)
pc.show()
```

image

### code 5

```python

import arviz_plots as azp

azp.style.use("arviz-variat")

data = load_arviz_data("centered_eight")
pc = azp.plot_pair(
    data,
    backend="bokeh" , 
    var_names=["mu","tau","theta"],
    coords={"school": ["Choate", "Deerfield"]},
    marginal=True,
    marginal_kind="kde",
    triangle="lower",
    visuals={"divergence": True},
)
pc.show()
```

### result 5
image

### code 6
```python

import arviz_plots as azp

azp.style.use("arviz-variat")

data = load_arviz_data("centered_eight")
pc = azp.plot_pair(
    data,
    backend="bokeh" , 
    var_names=["mu","tau","theta"],
    coords={"school": ["Choate", "Deerfield"]},
    marginal=False,
    triangle="both",
    visuals={"divergence": True, "remove_axis":"both"},
)
pc.show()
```
### result 6
image

src/arviz_plots/plots/pair_plot.py
```diff
@@ -0,0 +1,395 @@
...
+        pc_kwargs = set_grid_layout(
+            pc_kwargs, plot_bknd, distribution, num_rows=n_pairs, num_cols=n_pairs
+        )
+        pc_kwargs["aes"] = pc_kwargs.get("aes", {}).copy()
+        if "chain" in distribution:
+            pc_kwargs["aes"].setdefault("overlay", ["chain"])
+        pc_kwargs["figure_kwargs"].setdefault("sharey", True)
+        pc_kwargs["figure_kwargs"].setdefault("sharex", True)
+        if marginal:
+            pc_kwargs["figure_kwargs"]["sharey"] = False
```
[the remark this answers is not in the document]
The-Broken-Keyboard: This required a little bit logic change in bokeh backend's create_plotting_grid function. Before the logic was such that we couldn't have sharex="col" and sharey="row" simultaneously. Since they were mutually exclusive so it was throwing error . Fixed that in backend.

Also since having just text on diagonal plots when marginal is False was little bit weird because we required to give x and y location of text which varied for different variables ( in case if we want them to be at centre). So now I added scatter plot for diagonal plots as well when marginal is False. Otherwise we will have marginals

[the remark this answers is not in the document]
OriolAbril: > This required a little bit logic change in bokeh backend's create_plotting_grid function

Thanks! Bokeh has always been the one to generate more trouble for us so it ends up being less used and worked on. The refactor is making things much better on this end but still not fully there.

> Also since having just text on diagonal plots when marginal is False was little bit weird because we required to give x and y location of text which varied for different variables ( in case if we want them to be at centre). So now I added scatter plot for diagonal plots as well when marginal is False. Otherwise we will have marginals

I don't think there is any case where we want the scatter against itself. It is ok as a subproduct in `plot_pair_focus` when removing it grealy increases complexity of the code, but I would not add it explicitly.

I see two options on this. One is doing some updates to the initialization of PlotMatrix in order to request a grid with one less column but same number of rows, and storing it like so in the "plot" dataarray (which would still have the same shape and coordinates) with `None` on the diagonal. Example for 3 variables, we generate a 3x2 grid:

```
None | 0, 0 | 0, 1
1, 0 | None | 1, 1
2, 0 | 2, 1 | None
```

This means there are no diagonal elements to be filled at all.

Alternatively, adapt the code for the labels along the diagonal. The backend text function takes vertical and horizontal alignment which we can set to center the text around the provided `x` and `y`. And as that would be for the case without marginals with axis sharing, the x and y would actually be the same value, and we can compute it as:

```
text_center = (distribution.max(dim=sample_dims) + distribution.min(dim=sample_dims)) / 2
```

which should get something that looks centered

src/arviz_plots/plot_matrix.py
```diff
@@ -174,10 +218,29 @@ def get_target(self, var_name, selection, var_name_y=None, selection_y=None):
...
+            selection_x = selection
+            selection_y = self.fixed_selection
+        elif self.orientation == "col":
+            var_name_x = self.fixed_var_name
+            var_name_y = var_name
+            selection_x = self.fixed_selection
+            selection_y = selection
+        else:
+            var_name_x = var_name
+            selection_x = selection
```
OriolAbril, line 238: Given the behaviour of `subset_matrix_da` I think this means that the simple fact of `var_name_y` and `selection_y` being None will already behave like "diagonal" even if the `orientation` attribute is not None. I think it is actually fine and we only need to check orientation `row` or `col`, with no orientation meaning "diagonal" which I think is a good default.

src/arviz_plots/plot_matrix.py
```diff
@@ -355,6 +418,9 @@ def map_triangle(
                 artist_dims=artist_dims,
                 ignore_aes=ignore_aes,
             )
+
+        self.orientation = None
```
OriolAbril, line 422: I would actually `raise` if `orientation` is set here (see other comments too).

src/arviz_plots/plot_matrix.py
```diff
@@ -158,6 +160,48 @@ def _generate_viz_dt(self, **figure_kwargs):
...
+        """Fixed dictionary for subsetting of targets."""
+        return self._fixed_selection
+
+    @fixed_selection.setter
+    def fixed_selection(self, value):
+        self._fixed_selection = value
+
+    @fixed_var_name.setter
+    def fixed_var_name(self, value):
+        self._fixed_var_name = value
```
OriolAbril, line 203: I think we could have them as "simple" attributes set to `None` in `__init__`. Moreover, I don't think we expect/want users to interact with them directly but more so as an internal thing to reduce duplication between `map` and `map_row_col`

src/arviz_plots/plot_matrix.py
```diff
@@ -439,10 +510,98 @@ def map(
...
+            fun=fun,
+            fun_label=fun_label,
+            data=data,
+            coords=coords,
+            ignore_aes=ignore_aes,
+            subset_info=subset_info,
+            store_artist=store_artist,
+            artist_dims=artist_dims,
+            **kwargs,
+        )
```
OriolAbril, line 555: I would revert `orientation` back to `None` before returning.

src/arviz_plots/plot_matrix.py
```diff
@@ -473,11 +632,11 @@ def map(
         --------
         arviz_plots.PlotMatrix.map_triangle
         """
+        self.set_fixed_var_attributes(index, orientation)
```
OriolAbril, line 635: Here also revert the 3 attributes to `None` before returning.

OriolAbril (commented):
next round of review. I think it is shaping up very well

src/arviz_plots/plots/pair_plot.py
```diff
@@ -0,0 +1,452 @@
...
+
+def plot_pair(
+    dt,
+    var_names=None,
+    filter_vars=None,
+    group="posterior",
+    coords=None,
+    sample_dims=None,
+    marginal=True,
+    marginal_kind="kde",
```
OriolAbril, line 38: default to rcparam or even none and pass to plot_dist so plot_dist changes none to the rcparam

docs/source/gallery/distribution/08_plot_pair_distribution.py
```diff
@@ -0,0 +1,27 @@
...
+
+azp.style.use("arviz-variat")
+
+data = load_arviz_data("centered_eight")
+pc = azp.plot_pair(
+    data,
+    var_names=["mu","theta","tau"],
+    coords= {"school": ["Choate", "Deerfield"]},
+    marginal=True,
+    marginal_kind="kde",
```
OriolAbril, line 24: I would remove these two arguments given they are the defaults

docs/source/gallery/inference_diagnostics/10_plot_pair.py
```diff
@@ -0,0 +1,29 @@
...
+from arviz_base import load_arviz_data
+
+import arviz_plots as azp
+
+azp.style.use("arviz-variat")
+
+dt = load_arviz_data("centered_eight")
+pc = azp.plot_pair(
+    dt,
+    var_names=["theta", "mu"],
```
OriolAbril, line 22:
```suggestion
    var_names=["theta", "tau"],
```

Tau is the interesting one to vizualize with divergences

docs/source/gallery/inference_diagnostics/10_plot_pair.py
```diff
@@ -0,0 +1,29 @@
...
+    coords={"school": ["Lawrenceville", "Mt. Hermon"]},
+    visuals={"divergence":True},
+    marginal=True,
+    marginal_kind="kde",
```
OriolAbril, line 26: maybe change this so the plot looks more different than the one in the distributions section

```suggestion
    marginal=False,
```

src/arviz_plots/backend/bokeh/__init__.py
```diff
@@ -317,31 +317,33 @@ def create_plotting_grid(
                 subplot_kws_i["y_range"] = shared_yrange[col]
             if width_ratios is not None:
                 subplot_kws["width"] = plot_widths[col]
+
+            if row * cols + (col + 1) > number:
```
OriolAbril, line 321: we should add tests for this, but we don't yet have the setup for backend specific tests. I will ping you once I start working on [this issue](https://github.com/arviz-devs/arviz-plots/issues/100) so we can both add tests for backend functions

src/arviz_plots/plot_matrix.py
```diff
@@ -439,10 +459,172 @@ def map(
...
+        index : int, default 0
+            Index of the row to be mapped by the given plotting function.
+        data : Dataset, optional
+            Data to be subsetted at each iteration and to pass to `fun` as first positional
+            argument. Defaults to the data used to initialize the ``PlotMatrix``.
+        coords : mapping, optional
+            Dictionary of {coordinate names : coordinate values} that should
+            be used to subset the aes, data and viz objects before any faceting
+            or aesthetics mapping is applied.
+        ignore_aes : set, optional
```
OriolAbril, line 554:
```suggestion
        ignore_aes : str or set of str, default "all"
```

src/arviz_plots/plot_matrix.py
```diff
@@ -439,10 +459,172 @@ def map(
...
+            Add the subset info from :func:`arviz_base.xarray_sel_iter` to
+            the keyword arguments passed to `fun`. If true, then `fun` must
+            accept the keyword arguments ``var_name``, ``sel`` and ``isel``.
+            Moreover, if those were to be keys present in `**kwargs` their
+            values in `**kwargs` would be ignored.
+        store_artist : boolean, default True
+        artist_dims : mapping of {hashable : int}, optional
+            Dictionary of sizes for proper allocation and storage when using
+            ``map`` with functions that return an array of :term:`visual`.
+        **kwargs : mapping, optional
```
OriolAbril, line 567:
```suggestion
        **kwargs
```

src/arviz_plots/plot_matrix.py
```diff
@@ -439,10 +459,172 @@ def map(
...
+            with :class:`~xarray.DataArray` of :class:`~xarray.Dataset` type
+            will be subsetted on the current selection (if possible) before calling `fun`.
+            Slicing with dims and coords is applied to the relevant subset present in the
+            xarray object so dimensions with mapped asethetics not being present is not an issue.
+            However, using Datasets that don't contain all the variable names in `data`
+            will raise an error.
+
+        See Also
+        --------
+        arviz_plots.PlotMatrix.map_triangle
```
OriolAbril, line 578: I think all 4: map, map_triangle, map_row and map_col should reference the other three in their see also sections, with the first one being the most relevant one, map_row<->map_col and map<->map_triangle.

tests/test_plots.py
```diff
@@ -332,6 +333,59 @@ def test_plot_mcse_models(self, datatree, datatree2, backend):
         assert "model" in pc.aes["color"].dims
         assert "/x" in pc.aes.groups
 
+    def test_plot_pair(self, datatree, backend):
```
OriolAbril, line 336: either here or in follow up tests we should make sure that `triangle` argument is taken into account. That is, with triangle both `scatter` only has `None` on the diagonal, with lower on the diagonal and above... Same thing for `marginal=False`

docs/source/gallery/distribution/08_plot_pair_distribution.py
```diff
@@ -0,0 +1,27 @@
+"""
+# Scatterplot all variable against each other
```
OriolAbril, line 2:
```suggestion
# Scatterplot all variables against each other
```

OriolAbril (commented):
mostly test and doc polishing left and we can merge

src/arviz_plots/backend/bokeh/__init__.py
The-Broken-Keyboard: sure

src/arviz_plots/plots/pair_plot.py
```diff
@@ -0,0 +1,495 @@
...
+        plot_matrix.map(
+            remove_axis,
+            "remove_axis",
+            ignore_aes=remove_axis_ignore,
+            **remove_axis_kwargs,
+        )
+
+    # default removal of axis for better visualization
+    default_remove_axis_kwargs = {}
+    default_remove_axis_kwargs["axis"] = "both"
```
OriolAbril, line 473: There is nothing where we are using this as dict, it is much easier to follow the code if instead you replace the `**default_remove_axis_kwargs` for `axis="both"`.

src/arviz_plots/plots/pair_plot.py
```diff
@@ -0,0 +1,495 @@
...
+        Passed to :class:`arviz_plots.PlotMatrix`
+
+
+    Returns
+    -------
+    PlotMatrix
+
+    Examples
+    --------
+    Default plot_pair
```
OriolAbril, line 148: The examples should focus on illustrating the behaviour that is specific to `plot_matrix` and can be inferred/learned from the general tutorial. At the same time doing that helps clarify the behaviour of unique/complicated arguments.

In this case I would generate multiple plots with `plot_pair(dt, var_names=["mu", "tau"], marginal=, triangle=)` for different values of marginal and triangle to show the less common/potentially surprising cases. For example, with lower triangle only, the behaviour of `marginal` True/False is quite simple to explain, if the diagonal is free the labels with variable names go there, otherwise they go at the left/bottom of the figure. However, the cases with upper are not as polished (which is ok) but should have extensive docs to minimize issues and surprises

src/arviz_plots/plots/pair_plot.py
```diff
@@ -0,0 +1,495 @@
...
+    if labeller is None:
+        labeller = BaseLabeller()
+    if backend is None:
+        if plot_matrix is None:
+            backend = rcParams["plot.backend"]
+        else:
+            backend = plot_matrix.backend
+
+    if marginal_kind is None:
+        marginal_kind = rcParams["plot.density_kind"]
```
OriolAbril, line 188: This is not used anywhere, so we can leave the substitution of None for the rcParam to `plot_dist`

src/arviz_plots/plots/pair_plot.py
```diff
@@ -0,0 +1,495 @@
...
+        dist_plot_stats["point_estimate"] = stats.get("point_estimate", {})
+
+        plot_matrix = plot_dist(
+            dt,
+            var_names,
+            filter_vars,
+            group,
+            coords,
+            sample_dims,
+            marginal_kind,
```
OriolAbril, line 306:
```suggestion
            distribution,
            sample_dims=sample_dims,
            kind=marginal_kind,
```

We have already done all the filtering/subsetting to get `distribution` and `plot_dist` doesn't need data from any other group, so we can pass distribution to avoid processing the same thing twice.

OriolAbril (approved):

PR #296, opened 2025-06-27
Add Violin Plot

The-Broken-Keyboard:
**Work in progress (few components remaining)**

----
📚 Documentation preview 📚: https://arviz-plots--296.org.readthedocs.build/en/296/


src/arviz_plots/plots/violin_plot.py
```diff
@@ -0,0 +1,331 @@
...
+            if "model" in distribution:
+                warnings.filterwarnings("ignore", message="Your data appears to have a single")
+            left_density = distribution.azstats.kde(
+                dim=left_density_dims, **stats.get("left_density", {})
+            )
+
+        # flip x and y coordinates
+        density_flipped = left_density.rename({"plot_axis": "flipped_plot_axis"})
+        density_flipped = density_flipped.assign_coords(flipped_plot_axis=["y", "x"])
+        density_flipped = density_flipped.rename({"flipped_plot_axis": "plot_axis"})
```
OriolAbril, line 245:
```suggestion
        # flip x and y coordinates
        density_flipped = left_density.sel(plot_axis=["x", "y"]).assign_coords(plot_axis=["y", "x"])
```

This should achieve the same, and adding the `.sel` will make it work even if for some reason the initial dataarray already has y, x as coord values

src/arviz_plots/plots/violin_plot.py
```diff
@@ -0,0 +1,331 @@
...
+            if "model" in distribution:
+                warnings.filterwarnings("ignore", message="Your data appears to have a single")
+            right_density = distribution.azstats.kde(
+                dim=right_density_dims, **stats.get("right_density", {})
+            )
+
+        # flip x and y coordinates
+        density_flipped = right_density.rename({"plot_axis": "flipped_plot_axis"})
+        density_flipped = density_flipped.assign_coords(flipped_plot_axis=["y", "x"])
+        density_flipped = density_flipped.rename({"flipped_plot_axis": "plot_axis"})
```
OriolAbril, line 274:
```suggestion
        density_flipped = left_density.sel(plot_axis=["x", "y"]).assign_coords(plot_axis=["y", "x"])
        density_flipped.loc[{"plot_axis": "x"}] = -density_flipped.sel(plot_axis="x")
```

You can get the data as negative like so instead of adding an argument to the visual function

src/arviz_plots/plots/violin_plot.py
```diff
@@ -0,0 +1,331 @@
...
+            "rug",
+        ],
+        Sequence[str],
+    ] = None,
+    visuals: Mapping[
+        Literal[
+            "dist",
+            "credible_interval",
+            "point_estimate",
+            "point_estimate_text",
```
OriolAbril, line 48: I don't think we'll want the point estimate text in this plot, the other ones can be nice, if you haven't added it yet I don't think it is necessary you do at all

OriolAbril (commented):
As it is a draft, I skimmed to get the general idea and try and flag places where you might be creating too much work or backing into a corner

PR #300, opened 2025-06-30
add parallel plot

The-Broken-Keyboard:
**WIP**

-  Initial commit
-  Refinement of xtick-labels remains ( vertical orientation)  [Done]
-  Polishing remains for overall function and it's logic [In Progress]
- Faster backend function for `bokeh` remains [ Done ]
- Tests and documentation remains [ In Progress ]

---

📚 Documentation preview 📚: https://arviz-plots--300.org.readthedocs.build/en/300/


The-Broken-Keyboard:
@OriolAbril review please

src/arviz_plots/plots/parallel_plot.py
```diff
@@ -0,0 +1,238 @@
...
+            ignore_aes=div_ignore,
+            store_artist=backend == "none",
+            **div_kwargs,
+        )
+
+    # x-axis label
+    x_labels = data.coords["label"].values
+    # process labels to get vertical alignment
+    processed_labels = [re.sub(r"(.*?)\[(.*?)\]", r"\1\n\2", label) for label in x_labels]
+    total = len(data.values[0])
```
OriolAbril, line 211: instead of `todal` devine `x_values = np.arange(len(x_labels))`, move to before any plotting and send `x_values` to the multiple_lines function as `x`

try numpy otherwise `xr.DataArray(x_values, dims=["label"])`

src/arviz_plots/plots/parallel_plot.py
```diff
@@ -0,0 +1,238 @@
...
+            plot_collection, aes_by_visuals, "divergence", new_sample_dims
+        )
+        if "color" not in div_aes:
+            div_kwargs.setdefault("color", "black")
+        if "alpha" not in div_aes:
+            div_kwargs.setdefault("alpha", 0.1)
+        plot_collection.map(
+            multiple_lines,
+            "divergence",
+            data=divergent_data,
```
OriolAbril, line 200: options: 1) data.T, 2) data.transpose("label", "sample"). See comment below too

The-Broken-Keyboard:
@OriolAbril , I have made suggested changes. 

I have opened one PR in arviz-base too (https://github.com/arviz-devs/arviz-base/pull/81) , to make changes in dataset_to_dataarray function to generate vertical labels as well. ( Docs build  is failing because arviz-base PR isn't merged yet).

Please review.

The-Broken-Keyboard:
@OriolAbril , I have made all the changes. Only issue known to me is `store_artist`. Since artists are getting plotted in chunks. First non-divergent ones and then divergent ones.

how to store these varying number of artists?

src/arviz_plots/plots/parallel_plot.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+def plot_parallel(
+    dt,
+    var_names=None,
+    filter_vars=None,
+    group="posterior",
+    coords=None,
+    sample_dims=None,
+    plot_collection=None,
```
OriolAbril, line 29: I would add label type as an argument here, and then pass that to `dataset_to_dataarray`. With very few variables vertical mode with horizontal labels looks nice, but we'll generally have more variables in which case I think `flat`+rotation=90 would be better.

I think we should also keep the norm_method argument. Personally I think "rank" is the best method as when diagnosing we don't really care about the actual values, only seeing if diverging and non diverging samples have clear differences in distribution. But I am not strongly opinionated on making it the default.

src/arviz_plots/plots/parallel_plot.py
```diff
@@ -0,0 +1,229 @@
...
+    # x-axis label
+    xticks_kwargs = copy(visuals.get("xticks", {}))
+    if xticks_kwargs is not False:
+        _, _, xticks_ignore = filter_aes(plot_collection, aes_by_visuals, "xticks", new_sample_dims)
+        plot_collection.map(
+            set_xticks,
+            "xticks",
+            labels=x_labels,
+            values=x_values,
+            rotation=0,
```
OriolAbril, line 223: rotation should be added to `xticks_kwargs` as `setdefault` so users can override it

src/arviz_plots/plots/parallel_plot.py
```diff
@@ -0,0 +1,229 @@
...
+        )
+
+    # get labels and x-values
+    x_labels = data.coords["label"].values
+    x_values = np.arange(len(x_labels))
+
+    new_sample_dims = [combined_dim, "label"]
+    plot_bknd = import_module(f".backend.{backend}", package="arviz_plots")
+    if plot_collection is None:
+        pc_kwargs.setdefault("cols", None)
```
OriolAbril, line 179: I think this is already the default in `.wrap` so we can skip this

src/arviz_plots/plots/parallel_plot.py
```diff
@@ -0,0 +1,229 @@
...
+    if sample_dims is None:
+        sample_dims = rcParams["data.sample_dims"]
+    if isinstance(sample_dims, str):
+        sample_dims = [sample_dims]
+    if visuals is None:
+        visuals = {}
+    if pc_kwargs is None:
+        pc_kwargs = {}
+    else:
+        pc_kwargs = pc_kwargs.copy()
```
OriolAbril, line 121: I didn't realize this when we updated the call signature of the functions. By using `**pc_kwargs` is is always a dictionary and it is defined internally when calling the function, it can't be a user defined dictionary that they also have stored as a variable. We can get rid of the whole if else.

We still need to copy mutable values inside `pc_kwargs` like figure_kwargs or aes before modifying them though.

src/arviz_plots/plots/parallel_plot.py
```diff
@@ -0,0 +1,229 @@
...
+    plot_bknd = import_module(f".backend.{backend}", package="arviz_plots")
+    if plot_collection is None:
+        pc_kwargs.setdefault("cols", None)
+        pc_kwargs["figure_kwargs"] = pc_kwargs.get("figure_kwargs", {}).copy()
+        pc_kwargs = set_wrap_layout(pc_kwargs, plot_bknd, data)
+        pc_kwargs["aes"] = pc_kwargs.get("aes", {}).copy()
+        colors = plot_bknd.get_default_aes("color", 1, {})
+        pc_kwargs["aes"].setdefault("color", ["diverging"])
+        pc_kwargs["aes"].setdefault("alpha", ["diverging"])
+        pc_kwargs["color"] = pc_kwargs.get("color", [colors[0], "black"])
```
OriolAbril, line 186: I am unsure about which/how to choose colors. The version in current arviz uses the opposed idea: non diverging samples are black, diverging are "C0". I think I prefer adding the extra emphasis to diverging samples.. But I think C0->non diverging and C1->diverging could also look good. I will try to play around next week and see if there is an alternative I like more, but if someone else has opinions on the matter please comment too.

src/arviz_plots/plots/parallel_plot.py
The-Broken-Keyboard: but it is required to use set_wrap_layout ahead.

src/arviz_plots/backend/plotly/__init__.py
```diff
@@ -694,18 +774,34 @@ def xlabel(string, target, *, size=unset, color=unset, **artist_kws):
...
     if labels is None:
         labels = [str(label) for label in labels]
-    target.update_xaxes(tickmode="array", tickvals=ticks, ticktext=labels, **artist_kws)
+    kwargs = {
+        "tickmode": "array",
+        "tickvals": ticks,
+        "ticktext": labels,
+    }
+    if rotation is not unset:
+        kwargs["tickangle"] = rotation
```
OriolAbril, line 787:
```suggestion
    kwargs = {
        "tickmode": "array",
        "tickvals": ticks,
        "ticktext": labels,
    }
    if rotation is not unset:
        kwargs["tickangle"] = -rotation
```

plotly seems to be interpreting the rotation in the oposite direction as the other two backends. In general I'd say use `"ticklabel": rotation` directly when defining kwargs, `unset` values are filtered out. But we can change the sign of `unset` so keep the if so we can change the sign and get the labels in the same direction in all backends. Same applies to `yticks`

src/arviz_plots/backend/plotly/__init__.py
```diff
@@ -519,30 +591,38 @@ def text(
...
+    annotations = []
+    for x_i, y_i, s_i in zip(x_vals, y_vals, strings):
+        ann = target.add_annotation(
+            x=x_i,
+            y=y_i,
+            text=str_to_plotly_html(s_i),
+            font=font_kwargs,
+            **_filter_kwargs(kwargs, artist_kws),
+        )
+        annotations.append(ann)
```
OriolAbril, line 624: I don't think we need to loop here. We were using `atleast_1d` to make it work with .Scatter, but I don't think we use or need (at least for now) sending arrays of text and their x,y positions directly to the backend.

I am not opposed to use `add_annotation` instead of `.Scatter` with `mode="text"` but we need to do extensive testing to make sure it behaves like the old version did. `horizontal_align` differing from the vertical one and using `align` instead of `xanchor` might have differences for example, the docs on `align` are not at all clear to me

OriolAbril:
I have added changes related to documentation and api consistency mostly. I will merge unless I broke tests. If there is any doubt about the things I did or why do not hesitate to review the PR even if already merged.

PR #305, opened 2025-07-11
add face in dist_plot

The-Broken-Keyboard:
This PR adds filled faces to the distribution plots of `kde`, `ecdf`  and `hist` kind. 

----
📚 Documentation preview 📚: https://arviz-plots--305.org.readthedocs.build/en/305/

### DIFF
--- a/src/arviz_plots/backend/plotly/__init__.py
+++ b/src/arviz_plots/backend/plotly/__init__.py
@@ -547,7 +547,10 @@ def scatter(
 def step(x, y, target, *, color=unset, alpha=unset, width=unset, linestyle=unset, **artist_kws):
     """Interface to plotly for a step line."""
     artist_kws.setdefault("showlegend", False)
-    line_kwargs = {"color": color, "width": width, "dash": linestyle}
+    edge_color = artist_kws.pop("edgecolor", color)
+    if edge_color is unset:
+        edge_color = color
+    line_kwargs = {"color": edge_color, "width": width, "dash": linestyle}
     line_artist_kws = artist_kws.pop("line", {"shape": "hv"}).copy()
     kwargs = {"opacity": alpha}
     line_object = go.Scatter(
--- a/src/arviz_plots/backend/plotly/__init__.py
+++ b/src/arviz_plots/backend/plotly/__init__.py
@@ -402,14 +402,16 @@ def hist(
             edgecolor = color
 
     if step:
-        step_x = np.repeat(np.concatenate(([l_e[0]], r_e)), 2)[1:-1]
-        step_y = np.repeat(np.concatenate(([0], height)), 2)
+        x_coords = np.concatenate((np.asarray(l_e), [r_e[-1]]))
+        y_coords = np.concatenate((np.asarray(height), [height[-1]]))
+
         hist_object = go.Scatter(
-            x=step_x,
-            y=step_y,
+            x=x_coords,
+            y=y_coords,
+            mode="lines",
+            line_shape="hv",
             fill="none",
             line={"color": edgecolor},
-            mode="lines",
             **_filter_kwargs({"opacity": alpha}, artist_kws),
         )
     else:
--- a/src/arviz_plots/backend/plotly/__init__.py
+++ b/src/arviz_plots/backend/plotly/__init__.py
@@ -544,14 +544,29 @@ def scatter(
     return scatter_object
 
 
-def step(x, y, target, *, color=unset, alpha=unset, width=unset, linestyle=unset, **artist_kws):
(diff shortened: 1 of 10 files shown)


src/arviz_plots/plots/dist_plot.py
```diff
@@ -244,6 +253,17 @@ def plot_dist(
...
+    col_dim_list = ["model", "group", "energy", "alpha"]
+    col_dim_list_bool = [dim in distribution for dim in col_dim_list]
+    if any(col_dim_list_bool):
+        aes_by_visuals.setdefault(
+            "face", {"color"}.union(plot_collection.aes_set.difference({"y", "linestyle"}))
+        )
+    else:
+        aes_by_visuals.setdefault(
+            "face", plot_collection.aes_set.difference({"y", "linestyle", "color"})
+        )
```
OriolAbril, line 265: I don't understand the logic here. And hardcoding the `col_dim_list` will become a headache soon. I think mappings should be the same as for "dist" or `plot_collection.aes_set.difference({"y", "linestyle"})`

src/arviz_plots/plots/dist_plot.py
```diff
@@ -254,6 +274,7 @@ def plot_dist(
 
     # density
     density_kwargs = copy(visuals.get("dist", {}))
+    face_kwargs = copy(visuals.get("face", {}))
```
OriolAbril, line 277: We should discuss defaults tomorrow at the meeting.

src/arviz_plots/plots/dist_plot.py
```diff
@@ -299,6 +330,33 @@ def plot_dist(
...
+            face_density = density.rename({"plot_axis": "kwarg"})
+            face_density = face_density.assign_coords(
+                kwarg=[
+                    "y_top" if coord == "y" else coord
+                    for coord in face_density.coords["kwarg"].values
+                ]
+            )
+            zeros = xr.full_like(face_density.sel(kwarg="x"), 0)
+            zeros = zeros.assign_coords(kwarg=["y_bottom"])
+            face_density = xr.concat([face_density, zeros], dim="kwarg")
```
OriolAbril, line 346: we can use the `.pad` method for this so we don't need to build the dataset of zeros manually:

```python
face_density = (
    density
    .rename(plot_axis="kwarg")
    .sel(kwarg=["x", "y"])
    .pad(kwarg=(0, 1), constant_value=0)
    .assign_coords(kwarg=["x", "y_top", "y_bottom"])
)
```

src/arviz_plots/plots/dist_plot.py
OriolAbril: `face_density` not having `chain` means `density` doesn't have chain either but `dist` doesn't have any of this logic and it works. This means that the plots that call `plot_dist` internally are defining `aes_by_visual` to make sure things work. There are probably values being set for `dist` but not for `face` because it didn't exist yet. Let's discuss a bit tomorrow at the meeting, but I see two solutions:

* default to `set(aes_by_visuals["dist"]).difference("linestyle")` (this is right after the setdefault for `dist` key in aes_by_visual so will always exist)
* update the plots that use `plot_dist` to ensure they define something different explicitly when the default won't make sense (note that even if we do option 1 we might still need to make updates to some of the plots)

src/arviz_plots/backend/plotly/__init__.py
```diff
@@ -496,7 +483,10 @@ def scatter(
 def step(x, y, target, *, color=unset, alpha=unset, width=unset, linestyle=unset, **artist_kws):
     """Interface to plotly for a step line."""
     artist_kws.setdefault("showlegend", False)
-    line_kwargs = {"color": color, "width": width, "dash": linestyle}
+    edge_color = artist_kws.pop("edgecolor", color)
+    if edge_color is unset:
+        edge_color = color
+    line_kwargs = {"color": edge_color, "width": width, "dash": linestyle}
     line_artist_kws = artist_kws.pop("line", {"shape": "hv"}).copy()
```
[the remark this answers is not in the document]
OriolAbril: Thanks for checking the correct translations! We can default to `before` (I think it is already the case) but I think it should be configurable so visuals can choose that using our own convention (thus, set keyword and values independently of the backend like we do for horizontal and vertical align for example).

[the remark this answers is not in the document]
The-Broken-Keyboard: Yeah, `after` value shows consistent results with the `hist`  function. I have tested this out. So I have set it to `after` in case of `step_hist`.

OriolAbril (approved):

PR #311, opened 2025-07-23
Allow using "AX" notation for aliases to default aes cycle

OriolAbril:
I will try to combine the functionality with what we have in `combine_plots` (or relegate that part of `combine_plots` so it always happens automatically). The main question I have is we currently use `{aes}_{i}` as placeholders in the none backend, so `color_12`. This initial version allows both notations, `color_12`, `linestyle_3` or `C12`, `L3` but I think we should choose only one. We can either change the placeholders of the none backend to the `C12` convention or support dealiasing only the verbose placeholders of the none backend.

Example:

```python
azp.plot_dist(data, visuals={"dist": {"color": "C3", "linestyle": "L1"}});
```

example

### DIFF
--- a/src/arviz_plots/backend/matplotlib/__init__.py
+++ b/src/arviz_plots/backend/matplotlib/__init__.py
@@ -119,6 +119,37 @@ def get_default_aes(aes_key, n, kwargs=None):
     return get_agnostic_default_aes(aes_key, n, kwargs)
 
 
+def dealiase_aes_value(aes, value):
+    try:
+        if value.startswith(f"{aes}_"):
+            index = int(value.rsplit("_")[-1])
+        else:
+            index = int(value[1:])
+        dealiased_value = get_default_aes(aes, index+1)[index]
+    except ValueError:
+        return value
+    return dealiased_value
+
+
+def expand_aesthetic_aliases(plot_fn):
+    def _dealiased_plot_fn(*args, **kwargs):
+        for aes in ("color", "facecolor", "edgecolor"):
+            if (
+                aes in kwargs 
+                and isinstance(value := kwargs[aes], str)
+                and (value[0].lower() in ("c", aes[0]) or value.startswith(f"{aes}_"))
+            ):
+                kwargs[aes] = dealiase_aes_value(aes, value)
+        for aes in ("linestyle", "marker"):
+            if (
+                aes in kwargs 
+                and isinstance(value := kwargs[aes], str)
+                and (value[0].lower() == aes[0] or value.startswith(f"{aes}_"))
+            ):
+                kwargs[aes] = dealiase_aes_value(aes, value)
+        return plot_fn(*args, **kwargs)
+    return _dealiased_plot_fn
+
 def scale_fig_size(figsize, rows=1, cols=1, figsize_units=None):
     """Scale figure properties according to figsize, rows and cols.
 
@@ -319,6 +350,7 @@ def hist(
     )
 
 
+@expand_aesthetic_aliases
 def line(x, y, target, *, color=unset, alpha=unset, width=unset, linestyle=unset, **artist_kws):
     """Interface to matplotlib for a line plot."""
     artist_kws.setdefault("zorder", 2)


src/arviz_plots/backend/matplotlib/__init__.py
```diff
@@ -119,6 +119,37 @@ def get_default_aes(aes_key, n, kwargs=None):
     return get_agnostic_default_aes(aes_key, n, kwargs)
 
 
+def dealiase_aes_value(aes, value):
+    try:
+        if value.startswith(f"{aes}_"):
+            index = int(value.rsplit("_")[-1])
+        else:
+            index = int(value[1:])
+        dealiased_value = get_default_aes(aes, index+1)[index]
```
OriolAbril, line 128: add check about unique elements and error out if `index` is larger than that.

src/arviz_plots/backend/matplotlib/__init__.py
```diff
@@ -119,6 +119,37 @@ def get_default_aes(aes_key, n, kwargs=None):
...
+
+
+def expand_aesthetic_aliases(plot_fn):
+    def _dealiased_plot_fn(*args, **kwargs):
+        for aes in ("color", "facecolor", "edgecolor"):
+            if (
+                aes in kwargs 
+                and isinstance(value := kwargs[aes], str)
+                and (value[0].lower() in ("c", aes[0]) or value.startswith(f"{aes}_"))
+            ):
```
OriolAbril, line 141: allow only `C0`, `L2`... no `color_0`

aloctavodia: Now only "CX" syntax is allowed. I think it's fine to use only "CX" instead of CX, LX, MX. With "C" being from "cycler", not color.

src/arviz_plots/backend/matplotlib/__init__.py
aloctavodia: I think this is not necessary, I may have said before that this was necessary. But the matplotlib syntax for "CN" allows N to be larger than the length of the cycler. So C100 is a valid syntax.

aloctavodia:
We could add a section to the overview https://arviz-plots.readthedocs.io/en/latest/tutorials/overview.html 
We should also update this example https://arviz-plots.readthedocs.io/en/latest/gallery/add_reference_bands.html

src/arviz_plots/backend/bokeh/core.py
```diff
@@ -0,0 +1,769 @@
...
+
+
+def savefig(figure, path, **kwargs):
+    """Save the figure to a file.
+
+    Parameters
+    ----------
+    figure : bokeh.plotting.Figure
+        The figure to save.
+    filename : pathlib.Path
```
tomicapretto:
```suggestion
    path : pathlib.Path
```

src/arviz_plots/backend/alias_utils.py
```diff
@@ -0,0 +1,89 @@
...
+
+    The main aliases are "C0", "C1", "C13" which indicate to use the n-th
+    element of the default cycler for that aesthetic. This is available
+    for color, facecolor, edgecolor, marker and linestyle.
+
+    Moreover, the first three properties which are color based also allow
+    aliases for colors that depend on the background color:
+
+    * B0 -> the background color of the currently active theme
+    * B1 -> a color with high contrast with B0 (21:1 for black or white backgrounds)
```
tomicapretto:
```suggestion
    * B1 -> a color with high contrast with B0 (21:1 contrast for black or white backgrounds)
```

src/arviz_plots/backend/alias_utils.py
```diff
@@ -0,0 +1,89 @@
...
+                ):
+                    kwargs[aes] = dealiase_aes_value(aes, value)
+            for aes in ("linestyle", "marker"):
+                if (
+                    aes in kwargs
+                    and isinstance(value := kwargs[aes], str)
+                    and value[0] == "C"
+                    and (value[1].isdigit())
+                ):
+                    kwargs[aes] = dealiase_aes_value(aes, value)
```
tomicapretto: Same here, a regular expression could be more straightforward.

Also, seeing `dealiase_aes_value` is only used internally... do we need to have the `try-except`?

(NB: please don't feel you have to do it to get this merged).

OriolAbril:
<p>I am not sure if it is intentional but maybe to keep the plots looking nicer it might be better to use the radon dataset for this one for example</p>

---

 <a href='https://app.reviewnb.com/arviz-devs/arviz-plots/pull/311/discussion/#comment-3407532865'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

src/arviz_plots/backend/alias_utils.py
```diff
@@ -0,0 +1,89 @@
...
+                    case 1:
+                        dealiased_value = get_contrast_colors(bg_color=bg_color)[0]
+                    case 2:
+                        dealiased_value = get_contrast_colors(bg_color=bg_color)[1]
+                    case 3:
+                        dealiased_value = get_contrast_colors(bg_color=bg_color)[2]
+                    case _:
+                        raise ValueError("Unrecognized background dependent color alias")
+        except ValueError:
+            return value
```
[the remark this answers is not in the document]
OriolAbril: I will update to use regular expressions yes

tomicapretto:
<p>Sounds like a good idea!</p>

---

 <a href='https://app.reviewnb.com/arviz-devs/arviz-plots/pull/311/discussion/#comment-3407532865'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

PR #326, opened 2025-08-17
Allow values in visuals to be True

### LINKED ISSUE #302: Allow values in `visuals` dictionary to be `True`?
In such cases the `True` would be an alias to an empty dict to make sure plotting is triggered for elements that can be added but default to not being present (for example divergences or rug plots).

I think it is ok and we already do in a couple places, but if we allow it we should document it properly and ensure it works in general.

aloctavodia:
closes #302

### DIFF
--- a/docs/source/contributing/new_plot.md
+++ b/docs/source/contributing/new_plot.md
@@ -86,7 +86,7 @@ def plot_xyz(
         when plotted. Valid keys are the same as for `visuals`.
 
         [[Description of default aesthetic mappings]]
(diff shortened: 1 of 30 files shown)


OriolAbril (commented):
I would also update https://github.com/arviz-devs/arviz-plots/blob/main/tests/test_hypothesis_plots.py#L45 so we have tests checking it is valid to use `True` and it works as expected

OriolAbril (approved):

PR #328, opened 2025-08-18
Add plot_lm

### LINKED ISSUE #317: add plot_lm
Add a simplified version of `plot_lm` in current ArviZ. I think to start we could limit ourselves to something with 3 elements only: CI band, mean line and observation scatter. For example the plot from this section https://bambinos.github.io/bambi/notebooks/distributional_models.html#constant-alpha (adding link in addition to copy of the image so you have code to generate data like that to test the function, it might not match all the assumptions below though but it has the data and reorganizing it should be easier than generating it I think):

Image

We need to access potentially 4 different data elements:

* observations for target variable. It should be in `idata.observed_data`
* observations for independent variable. It should be in `idata.constant_data` in general, but I think we could try allowing string/list (access from constant data) or dataarray/dataset to use directly.
* predictions/posterior predictive. I think as default we can get the same variables there are in observed_data group from the posterior_predictive group. I would restrict this to string/list only. But we might want to allow `posterior_predictive/y`/`predictions/y` to allow indicating which groups holds the data we want to plot. We'd then compute the point estimate and credible interval according to rcParams
  - open question: cc @aloctavodia do we want to integrate smoothing here? or support smoothing only if done manually and then pre-computed summaries are provided via `stats` dict?
* independent variable data for ci/mean. I think this should be optional and default to the observations for independent variable, but it should be possible to change. One automatic change in default could be when using data from `predictions` group, get data with same same as independent variable observations but from `predictions_constant_data` instead. And like the observation analog, it should also allow dataarray/dataset.

---

I would start allowing to generate only one plot in the figure. After that we can discuss how to incorporate faceting/mappings to that for multiple outcomes, multiple covariates, both at the same time?

The-Broken-Keyboard:
Documentations and tests are still remaining. 

closes #317 

----
📚 Documentation preview 📚: https://arviz-plots--328.org.readthedocs.build/en/328/

### DIFF
--- a/src/arviz_plots/plots/lm_plot.py
+++ b/src/arviz_plots/plots/lm_plot.py
@@ -84,7 +84,8 @@ def plot_lm(
         Independent variable(s) for predictions.
 
         If None (default), and if group is "predictions", all variables corresponding to x data
-        in "predictions_constant_data" group are used. If group is not "predictions", x is used.
+        in "predictions_constant_data" group are used. If group is "posterior_predictive",
+        x is used.
     filter_vars: {None, “like”, “regex”}, default None
         If None (default), interpret var_names as the real variables names.
         If “like”, interpret var_names as substrings of the real variables names.
--- a/src/arviz_plots/plots/lm_plot.py
+++ b/src/arviz_plots/plots/lm_plot.py
@@ -1,0 +1,324 @@
+"""lm plot code."""
+from collections.abc import Mapping, Sequence
+from copy import copy
+from importlib import import_module
+from typing import Any, Literal
+
+import numpy as np
+import xarray as xr
+from arviz_base import rcParams
+from scipy.interpolate import griddata
+from scipy.signal import savgol_filter
+
+from arviz_plots.plot_collection import PlotCollection
+from arviz_plots.plots.utils import (
+    filter_aes,
+    get_contrast_colors,
+    process_group_variables_coords,
+    set_wrap_layout,
+)
+from arviz_plots.visuals import fill_between_y, labelled_x, labelled_y, line_xy, scatter_xy
+
+
+def plot_lm(
+    dt,
+    target_data=None,
+    independent_data=None,
+    prediction_data=None,
+    ci_mean_independent_data=None,
+    filter_vars=None,
+    prediction_group="predictions",
+    coords=None,
+    sample_dims=None,
+    ci_prob=None,
+    ci_kind=None,
+    smooth=True,
+    plot_collection=None,
+    backend=None,
+    aes_by_visuals: Mapping[
+        Literal[
+            "line",
+            "mean_line",
+            "fill",
+            "scatter",
+            "xlabel",
+            "ylabel",
+        ],
+        Sequence[str],
+    ] = None,
+    visuals: Mapping[
+        Literal[
+            "line",
+            "mean_line",
+            "fill",
+            "scatter",
+            "xlabel",
+            "ylabel",
+        ],
+        Mapping[str, Any] | Literal[False],
+    ] = None,
+    stats: Mapping[
+        Literal["credible_interval",],
+        Mapping[str, Any] | xr.Dataset,
+    ] = None,
+    **pc_kwargs,
+):
+    """Posterior predictive and mean plots for regression-like data..
+
+    Parameters
+    ----------
+    dt : DataTree
+        Input data
+
+    """
+    if sample_dims is None:
+        sample_dims = rcParams["data.sample_dims"]
+    if isinstance(sample_dims, str):
+        sample_dims = [sample_dims]
+    if visuals is None:
+        visuals = {}
+    if pc_kwargs is None:
+        pc_kwargs = {}
+    else:
+        pc_kwargs = pc_kwargs.copy()
+
+    if backend is None:
+        if plot_collection is None:
+            backend = rcParams["plot.backend"]
+        else:
+            backend = plot_collection.backend
+    if ci_prob is None:
+        ci_prob = rcParams["stats.ci_prob"]
+    if ci_kind is None:
+        ci_kind = rcParams["stats.ci_kind"] if "stats.ci_kind" in rcParams else "eti"
+
+    if aes_by_visuals is None:
+        aes_by_visuals = {}
+    else:
+        aes_by_visuals = aes_by_visuals.copy()
+
+    if stats is None:
+        stats = {}
+    else:
+        stats = stats.copy()
+
+    if target_data is None:
+        target_data = process_group_variables_coords(
+            dt, group="observed_data", var_names="y", filter_vars=filter_vars, coords=coords
+        )
+    elif isinstance(target_data, str):
+        target_data = process_group_variables_coords(
+            dt, group="observed_data", var_names=target_data, filter_vars=filter_vars, coords=coords
+        )
+
+    if independent_data is None:
+        independent_data = process_group_variables_coords(
+            dt, group="constant_data", var_names="x", filter_vars=filter_vars, coords=coords
+        )
+    elif isinstance(independent_data, str):
+        independent_data = process_group_variables_coords(
+            dt,
+            group="constant_data",
+            var_names=independent_data,
+            filter_vars=filter_vars,
+            coords=coords,
+        )
+
+    (target_var,) = target_data.data_vars
+    (independent_var,) = independent_data.data_vars
+    # pred_target_var = target_var
+    pred_independent_var = independent_var
+
+    if prediction_data is None:
+        prediction_data = process_group_variables_coords(
+            dt, group=prediction_group, var_names=target_var, filter_vars=filter_vars, coords=coords
+        )
+    elif isinstance(prediction_data, str):
+        # pred_target_var = prediction_data
+        prediction_data = process_group_variables_coords(
+            dt,
+            group=prediction_group,
+            var_names=prediction_data,
+            filter_vars=filter_vars,
+            coords=coords,
+        )
+
+    if ci_mean_independent_data is None:
+        if prediction_group == "predictions":
+            ci_mean_independent_data = process_group_variables_coords(
+                dt,
+                group="predictions_constant_data",
+                var_names=independent_var,
+                filter_vars=filter_vars,
+                coords=coords,
+            )
+        else:
+            ci_mean_independent_data = independent_data
+    elif isinstance(ci_mean_independent_data, str):
+        if prediction_group == "predictions":
+            pred_independent_var = ci_mean_independent_data
+            ci_mean_independent_data = process_group_variables_coords(
+                dt,
+                group="predictions_constant_data",
+                var_names=ci_mean_independent_data,
+                filter_vars=filter_vars,
+                coords=coords,
+            )
+        else:
+            ci_mean_independent_data = independent_data
+
+    plot_bknd = import_module(f".backend.{backend}", package="arviz_plots")
+    bg_color = plot_bknd.get_background_color()
+    _, contrast_gray_color = get_contrast_colors(bg_color=bg_color, gray_flag=True)
+    if plot_collection is None:
+        pc_kwargs.setdefault("cols", None)
+        pc_kwargs["figure_kwargs"] = pc_kwargs.get("figure_kwargs", {}).copy()
+        pc_kwargs["aes"] = pc_kwargs.get("aes", {}).copy()
+        pc_kwargs = set_wrap_layout(pc_kwargs, plot_bknd, prediction_data)
+        plot_collection = PlotCollection.wrap(
+            prediction_data,
+            backend=backend,
+            **pc_kwargs,
+        )
+
+    if aes_by_visuals is None:
+        aes_by_visuals = {}
+    else:
+        aes_by_visuals = aes_by_visuals.copy()
+    aes_by_visuals.setdefault("line", plot_collection.aes_set)
+
+    # calculations for credible interval
+    if ci_kind == "eti":
+        ci = prediction_data.azstats.eti(prob=ci_prob, **stats.get("credible_interval", {}))
+    elif ci_kind == "hdi":
+        ci = prediction_data.azstats.hdi(prob=ci_prob, **stats.get("credible_interval", {}))
+    ci_upper = ci.y.sel(ci_bound="upper").values
+    ci_lower = ci.y.sel(ci_bound="lower").values
+
+    mean_data = prediction_data.mean(dim=["chain", "draw"])
+    pred_x_vals = ci_mean_independent_data[pred_independent_var].values
+    colors = plot_bknd.get_default_aes("color", 2, {})
+
+    ci_data = np.stack([ci_lower, ci_upper], axis=1)
+
+    if smooth is True:
+        smooth_kwargs = {}
+        smooth_kwargs.setdefault("window_length", 55)
+        smooth_kwargs.setdefault("polyorder", 2)
+        x_data = np.linspace(pred_x_vals.min(), pred_x_vals.max(), 200)
+        x_data[0] = (x_data[0] + x_data[1]) / 2
+        hdi_interp = griddata(pred_x_vals, ci_data, x_data)
+        y_data = savgol_filter(hdi_interp, axis=0, **smooth_kwargs)
+    else:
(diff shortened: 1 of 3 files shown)


src/arviz_plots/plots/lm_plot.py
```diff
@@ -0,0 +1,324 @@
...
+    ci_upper = ci.y.sel(ci_bound="upper").values
+    ci_lower = ci.y.sel(ci_bound="lower").values
+
+    mean_data = prediction_data.mean(dim=["chain", "draw"])
+    pred_x_vals = ci_mean_independent_data[pred_independent_var].values
+    colors = plot_bknd.get_default_aes("color", 2, {})
+
+    ci_data = np.stack([ci_lower, ci_upper], axis=1)
+
+    if smooth is True:
```
aloctavodia, line 204: We need to add a smoothing function in arviz-stats

src/arviz_plots/plots/lm_plot.py
```diff
@@ -0,0 +1,324 @@
...
+        y_data = savgol_filter(hdi_interp, axis=0, **smooth_kwargs)
+    else:
+        idx = np.argsort(pred_x_vals)
+        x_data = pred_x_vals[idx]
+        y_data = ci_data[idx]
+
+    ci_lower = y_data[:, 0]
+    ci_upper = y_data[:, 1]
+
+    # upper and lower lines of credible interval
```
aloctavodia, line 220: This visual could be named "ci_line" or something similar. Not sure we want this given that we already have "fill", Maybe we want it, but it should be false by default, as we already have "fill". Both lines should be of the same color (maybe constant_gray_color?), and dashed.

src/arviz_plots/plots/lm_plot.py
```diff
@@ -0,0 +1,324 @@
...
+        data,
+        dims=["kwarg", "index"],
+        coords={
+            "kwarg": ["x", "y_bottom", "y_top"],
+            "index": np.arange(len(x_data)),
+        },
+        name="ci_bounds",
+    )
+
+    # fill between lines of credible interval
```
aloctavodia, line 254: It would be nice to support multiple intervals.

src/arviz_plots/plots/lm_plot.py
```diff
@@ -0,0 +1,324 @@
...
+
+    # calculations for credible interval
+    if ci_kind == "eti":
+        ci = prediction_data.azstats.eti(prob=ci_prob, **stats.get("credible_interval", {}))
+    elif ci_kind == "hdi":
+        ci = prediction_data.azstats.hdi(prob=ci_prob, **stats.get("credible_interval", {}))
+    ci_upper = ci.y.sel(ci_bound="upper").values
+    ci_lower = ci.y.sel(ci_bound="lower").values
+
+    mean_data = prediction_data.mean(dim=["chain", "draw"])
```
aloctavodia, line 198: We should offer an option to compute the mean and median. 
Should we also offer an option to "smooth" the results?

src/arviz_plots/plots/lm_plot.py
The-Broken-Keyboard: where in arviz-stats? In `helper_stats.py`?

src/arviz_plots/plots/lm_plot.py
The-Broken-Keyboard: so you mean we will support a list of ci_probs and calculate intervals for each probability and then plot them separately? Let me know if I misunderstood something

src/arviz_plots/plots/lm_plot.py
aloctavodia: Not sure what the final form of this smooth function should be, and not sure where it should be, helper_stats contains other "half-backed" computations, so it may be a good place in the meantime. Maybe @OriolAbril has a clearer idea.

src/arviz_plots/plots/lm_plot.py
The-Broken-Keyboard: yes that sounds good....then we will need to make changes in arviz-stats also probably, so that it can calculate multiple intervals together

OriolAbril: we could add an alpha aesthetic mapping when `ci_prob` is not a scalar so the same call we already have to fill_between works

src/arviz_plots/plots/lm_plot.py
```diff
@@ -0,0 +1,324 @@
...
+
+    # x-axis label
+    xlabel_kwargs = copy(visuals.get("xlabel", {}))
+
+    if xlabel_kwargs is not False:
+        _, _, xlabel_ignore = filter_aes(plot_collection, aes_by_visuals, "xlabel", sample_dims)
+        plot_collection.map(
+            labelled_x,
+            "xlabel",
+            text="Predictor",
```
OriolAbril, line 307: we can probably try and get a better default (`independent_data` as a dataarray might have a name we could use here). But in either case, this should go into `**xlabel_kwargs` through `setdefault` so users can choose the text if desired. Same for ylabel

src/arviz_plots/plots/lm_plot.py
```diff
@@ -0,0 +1,324 @@
...
+)
+from arviz_plots.visuals import fill_between_y, labelled_x, labelled_y, line_xy, scatter_xy
+
+
+def plot_lm(
+    dt,
+    target_data=None,
+    independent_data=None,
+    prediction_data=None,
+    ci_mean_independent_data=None,
```
OriolAbril, line 28: we should discuss these argument names tomorrow

src/arviz_plots/plots/lm_plot.py
```diff
@@ -0,0 +1,324 @@
...
+    filter_vars=None,
+    prediction_group="predictions",
```
OriolAbril, line 30: given all the other groups are hardcoded, I think this could simply be `group`

src/arviz_plots/plots/lm_plot.py
```diff
@@ -0,0 +1,324 @@
...
+        stats = stats.copy()
+
+    if target_data is None:
+        target_data = process_group_variables_coords(
+            dt, group="observed_data", var_names="y", filter_vars=filter_vars, coords=coords
+        )
+    elif isinstance(target_data, str):
+        target_data = process_group_variables_coords(
+            dt, group="observed_data", var_names=target_data, filter_vars=filter_vars, coords=coords
+        )
```
OriolAbril, line 112: as written now, this can be reduce to the elif as an if instead and `target_data="y"` in the call signature. I think we can probably have a better default if we discuss the behaviour tomorrow (and potentially support multiple variables) so I would change it to something like:

```
if target_data is None:
    target_data = "y"
if not isinstance(target_data, xr.DataArray | xr.Dataset):
    target_data = process_group...
```

src/arviz_plots/plots/lm_plot.py
```diff
@@ -0,0 +1,324 @@
...
+    if original_scatter_kwargs is not False:
+        _, _, scatter_ignore = filter_aes(plot_collection, aes_by_visuals, "scatter", sample_dims)
+        plot_collection.map(
+            scatter_xy,
+            "scatter",
+            x=independent_data[independent_var].values,
+            y=target_data[target_var].values,
+            alpha=0.3,
+            color=contrast_gray_color,
+            width=0,
```
OriolAbril, line 294: all of these should be added through `setdefault` and only when on aesthetic is mapped to them. The scatter/line/area can be enough distinction between observed/mean/credible interval and we might want all to have a mapping from color to group like we did in https://arviz-plots.readthedocs.io/en/latest/tutorials/compose_own_plot.html#option-2-specific-functions-for-each-visual-element

src/arviz_plots/plots/lm_plot.py
```diff
@@ -0,0 +1,324 @@
...
+    if pc_kwargs is None:
+        pc_kwargs = {}
+    else:
+        pc_kwargs = pc_kwargs.copy()
+
+    if backend is None:
+        if plot_collection is None:
+            backend = rcParams["plot.backend"]
+        else:
+            backend = plot_collection.backend
```
OriolAbril, line 89: I think this was needed a while ago before we simplified all the automatic figsizing logic. `PlotCollection.grid` and `.wrap` can take `None` as backend and will use the rcParam, so we can simplify this to

```
backend = plot_collection.backend
```

if we put it after we initialize the plot_collection

src/arviz_plots/plots/lm_plot.py
OriolAbril: target_data -> y 
independent_data -> x
prediction_data -> y_pred
ci_mean_independent_data -> x_pred

src/arviz_plots/plots/lm_plot.py
```diff
@@ -0,0 +1,324 @@
...
+                coords=coords,
+            )
+        else:
+            ci_mean_independent_data = independent_data
+
+    plot_bknd = import_module(f".backend.{backend}", package="arviz_plots")
+    bg_color = plot_bknd.get_background_color()
+    _, contrast_gray_color = get_contrast_colors(bg_color=bg_color, gray_flag=True)
+    if plot_collection is None:
+        pc_kwargs.setdefault("cols", None)
```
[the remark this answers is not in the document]
OriolAbril: ~~If we do this it means that everything that is related `y` or `y_pred` needs to be a Dataset, and everything that is related to `x` or `x_pred` needs to be a DataArray.~~ **We could swap and have x as Dataset and y as DataArray** default to this second option for now, which means use data from `x` to initialize the PlotCollection

The-Broken-Keyboard:
Few examples :

1. Multiple independent variables with multiple bands

image

2. Multiple iv with single ci_band

image

3. Single iv with single band

image

The-Broken-Keyboard:
example:

image

src/arviz_plots/plots/lm_plot.py
```diff
@@ -0,0 +1,412 @@
...
+    sample_dims : iterable, optional
+        Dimensions to reduce unless mapped to an aesthetic.
+        Defaults to ``rcParams["data.sample_dims"]``
+    ci_kind : {"hdi", "eti"}, optional
+        Which credible interval to use. Defaults to ``rcParams["stats.ci_kind"]``
+    ci_prob : float or list of float, optional
+        Indicates the probabilities that should be contained within the plotted credible intervals.
+        Defaults to ``rcParams["stats.ci_prob"]``
+    line_kind : {"mean", "median"}, optional
+        Which point estimate to use for the line. Defaults to ``rcParams["stats.point_estimate"]``
```
OriolAbril, line 106: mode is also available now

src/arviz_plots/plots/lm_plot.py
```diff
@@ -0,0 +1,412 @@
...
+    plot_collection : PlotCollection, optional
+    backend : {"matplotlib", "bokeh"}, optional
+    labeller : labeller, optional
+    aes_by_visuals : mapping, optional
+        Mapping of visuals to aesthetics that should use their mapping in `plot_matrix`
```
OriolAbril, line 111:
```suggestion
        Mapping of visuals to aesthetics that should use their mapping in `plot_collection`
```

src/arviz_plots/plots/lm_plot.py
```diff
@@ -0,0 +1,412 @@
...
+
+        If None (default), and if group is "predictions", all variables corresponding to x data
+        in "predictions_constant_data" group are used. If group is "posterior_predictive",
+        x is used.
+    filter_vars: {None, “like”, “regex”}, default None
+        If None (default), interpret var_names as the real variables names.
+        If “like”, interpret var_names as substrings of the real variables names.
+        If “regex”, interpret var_names as regular expressions on the real variables names.
+    group : str, default "posterior_predictive"
+        Group to use for plotting. Defaults to "posterior_predictive".
```
OriolAbril, line 94:
```suggestion
        Group to use for plotting.
```

already indicated above

src/arviz_plots/plots/lm_plot.py
```diff
@@ -0,0 +1,412 @@
...
+    x_pred : str or list of str or DataArray or Dataset, optional
+        Independent variable(s) for predictions.
+
+        If None (default), and if group is "predictions", all variables corresponding to x data
+        in "predictions_constant_data" group are used. If group is "posterior_predictive",
+        x is used.
+    filter_vars: {None, “like”, “regex”}, default None
+        If None (default), interpret var_names as the real variables names.
+        If “like”, interpret var_names as substrings of the real variables names.
+        If “regex”, interpret var_names as regular expressions on the real variables names.
```
OriolAbril, line 92: Indicate when it is used and when it is ignored in addition to the behaviour when used. I think it is only relevant for non xarray inputs, but used for any of the x, y, y_pred, x_pred that are str/list[str]

src/arviz_plots/plots/lm_plot.py
```diff
@@ -0,0 +1,412 @@
...
+            "line",
+            "ci_fill",
+            "scatter",
+            "xlabel",
+            "ylabel",
+        ],
+        Mapping[str, Any] | Literal[False],
+    ] = None,
+    stats: Mapping[
+        Literal["credible_interval",],
```
OriolAbril, line 65: the central line/point estimate should also be in here

src/arviz_plots/plots/lm_plot.py
```diff
@@ -0,0 +1,412 @@
...
+    plot_bknd = import_module(f".backend.{backend}", package="arviz_plots")
+    bg_color = plot_bknd.get_background_color()
+    _, contrast_gray_color = get_contrast_colors(bg_color=bg_color, gray_flag=True)
+    if plot_collection is None:
+        pc_kwargs.setdefault("cols", "__variable__")
+        pc_kwargs["figure_kwargs"] = pc_kwargs.get("figure_kwargs", {}).copy()
+        pc_kwargs["aes"] = pc_kwargs.get("aes", {}).copy()
+        if isinstance(ci_prob, Sequence):
+            alpha_dims = pc_kwargs["aes"].get("alpha", None)
+            if alpha_dims is None:
```
OriolAbril, line 246: we don't use alpha dims so we can update the condition to

```suggestion
            if "alpha" not in pc_kwargs["aes"]:
```

src/arviz_plots/plots/lm_plot.py
```diff
@@ -0,0 +1,412 @@
...
+    else:
+        aes_by_visuals = aes_by_visuals.copy()
+    aes_by_visuals.setdefault("line", plot_collection.aes_set.difference("alpha"))
+    if isinstance(ci_prob, Sequence):
+        aes_by_visuals.setdefault("ci_line", {"alpha"})
+        aes_by_visuals.setdefault("ci_fill", {"alpha"})
+
+    # calculations for credible interval
+    ci_fun = azs.hdi if ci_kind == "hdi" else azs.eti
+    ci_dims, _, _ = filter_aes(plot_collection, aes_by_visuals, "ci_fill", sample_dims)
```
OriolAbril, line 274: you can already define `fill_aes` and `fill_ignore` here and skip calling it inside the `if fill_kwargs is not False`

src/arviz_plots/plots/lm_plot.py
```diff
@@ -0,0 +1,412 @@
...
+            ],
+            dim="prob",
+        )
+    else:
+        ci_data = ci_fun(y_pred, dim=ci_dims, prob=ci_prob, **stats.get("credible_interval", {}))
+
+    if line_kind == "mean":
+        line_data = y_pred.mean(dim=["chain", "draw"])
+    elif line_kind == "median":
+        line_data = y_pred.median(dim=["chain", "draw"])
```
OriolAbril, line 291: chain and draw should not be hardcoded

OriolAbril (commented):
> example:
> image

Try generating this same plot but with `"color": ["__variable__"]` aesthetic defined that is active for `fill`, `line` and `scatter`.

The-Broken-Keyboard:
@OriolAbril I have incorporated the changes you suggested, there is just one difference. I kept color of scatter as contrasting_gray_color and color of central_line as contrasting color. While color of ci_fill is mapped to color aesthetics of plot_collection which is mapped to variable dimension. I didn't activate color aesthetic for scatter and central line because they look indifferentiable when kept same... therefore..

below are examples in different themes:

arviz-variat:

image

arviz-tumma:

image

Issue #51, opened 2024-06-04
Plot orientation


OriolAbril:
We have requests from time to time about rotating the plots, for example plot_forest. Is this something we want to support? If so, for a subset of the plots like plot_forest/plot_ridge only for as a general argument for all plots?

aloctavodia:
At least as a workaround. Is it enough to rotate all text elements and then the user manually rotates the figure? Or do we want a single argument that rotates all elements in the plot?

OriolAbril:
I think the 2nd option makes more sense. In most of the cases the plots are generated but not stored into a proper image, much less edited as images, but it can still make sense to prefer it horizontal because it uses screen space better, it fits better within an educational notebook...

PR #329, opened 2025-08-21
plot_compare allow to rotate 

aloctavodia:
This adds the `rotate` argument (default to False). 

It also introduces other changes
* Change "shade" visuals for "similar_line", by default, this plots a dotted gray line. This is done to avoid visual clashing due to the next change  (but we can discuss other options).
* Add `ref_band`, add a band spanning the entire plot representing the standard error of the ELPD estimate. This is similar to `ref_line`, which does the same but for the point estimate.
* Adds `hide_top_model` argument (see example below).
* Removed the "ranked model" x/y_label.

Defaults
default

```rotated=True```
rotated

```{python}
pc = azp.plot_compare(cmp_df,
                      relative_scale=True,
                      visuals={"similar_line":True},
                      )
```
Notice that `similar_line` is always at -4 when using the relative scale.
relative_similar

```{python}
pc = azp.plot_compare(cmp_df,
                      relative_scale=True,
                      hide_top_model=True,
                      rotated=True,
                      visuals={"ref_band":True, "title": False},
                      )
```
This is intended to be the default in Kulprit. But I see it can be useful in other settings too. Here, the `cmp_df` dataframe includes a "reference model" against which we want to compare the rest of the models.  The main goal of the plot is to allow for easy finding of the smallest model that is "indistinguishable" from the reference model.  Thus, instead of representing the reference model explicitly (with a point + line), it becomes useful to represent it implicitly with the `ref_line` and `ref_band`.
kulprit


src/arviz_plots/plots/compare_plot.py
```diff
@@ -27,44 +35,62 @@ def plot_compare(
...
+        :func:`~arviz_plots.backend.none.vline` depending on the
+        ``rotated`` parameter.
+        * ref_band -> passed to :func:`~arviz_plots.backend.none.hspan` or
+        :func:`~arviz_plots.backend.none.vspan` depending on the
+        ``rotated`` parameter. Defaults to ``False``.
+        * similar_line -> passed to :func:`~arviz_plots.backend.none.hline` or
+        :func:`~arviz_plots.backend.none.vline` depending on the
+        ``rotated`` parameter. Defaults to ``False``.
+        * labels -> passed to :func:`~arviz_plots.backend.none.xticks` and
+        :func:`~arviz_plots.backend.none.yticks`
```
OriolAbril, line 83: multiline bullet points need to be indented, otherwise sphinx thinks the list ends there and we are starting a new paragraph (but sends a warning which caught it: https://app.readthedocs.org/projects/arviz-plots/builds/29276643/#283097490--347)

```suggestion
        * ref_line -> passed to :func:`~arviz_plots.backend.none.hline` or
          :func:`~arviz_plots.backend.none.vline` depending on the
          ``rotated`` parameter.
        * ref_band -> passed to :func:`~arviz_plots.backend.none.hspan` or
          :func:`~arviz_plots.backend.none.vspan` depending on the
          ``rotated`` parameter. Defaults to ``False``.
        * similar_line -> passed to :func:`~arviz_plots.backend.none.hline` or
          :func:`~arviz_plots.backend.none.vline` depending on the
          ``rotated`` parameter. Defaults to ``False``.
        * labels -> passed to :func:`~arviz_plots.backend.none.xticks` and
          :func:`~arviz_plots.backend.none.yticks`
```

OriolAbril (approved):
mentioning https://github.com/arviz-devs/arviz-plots/issues/51 so the issue gets the PR linked in the history/discussion

PR #343, opened 2025-09-11
Refactor plot_lm 

aloctavodia:
I started working on adding a smooth option to `plot_lm` as presented in #317. But after playing a little bit with the function, I decided to incorporate other changes, which may be a better fit given the design and expectations of other plots in ArviZ.  But given that  `plot_lm` has always been a weird plot for us, and I may be too focused on what I want to do, and then miss other uses. So, to make things more concrete, I will show the steps of the two main use cases I have in mind. Notice the `combine` function I am adding in this PR; I may be overcomplicating things there, or not making the correct assumptions (~e.g., it fails when ci_prob is a list instead of a float~).

A common pattern in ArviZ-plots is to create a DataTree first, plot second. So I follow that pattern here. For instance, the arguments "x" and "y" are equivalent to "var_names" in other plots.

For the most straightforward example, we have a linear regression with only one predictor. Here I am using the bikes dataset and a NegativeBinomial family, the model is essentially.

`rented ~ temperature`

Once the model is sampled, we do 

```python
idata_lb.add_groups({"constant_data": xr.DataArray(bikes.temperature)})
```

And then we can plot the predictions with

```python
azp.plot_lm(idata_lb)
```

Multiple bands

```python
pc = azp.plot_lm(idata_lb,
                 ci_prob=[0.5, 0.9, 0.95],
                 visuals={"observed_scatter":False,
                          "pe_line": False})
```

output

Notice the y-axis labels are not correct yet.

More often than not. We have regression models with more than one covariate. For those cases, what we want to plot is the marginal predictions (or marginals eta/linear-terms).

`rented ~ temperature + humidity`

For such models, we need to compute the marginals somehow. PyMC-BART has custom functions for this that take advantage of the tree structure to efficiently compute predictions. Bambi has the interpret module for the slope/predictions. I mention those for context, but also because, at least in the PyMC-BART case, it would be nice to replace the plotting part with arviz-plots.  

Assuming we want to compute the marginals manually, we could do something like this 

```python
import xarray as xr
from xarray_einstats.stats import XrDiscreteRV
import preliz as pz

# Compute marginal "slopes"
alpha = idata_lb.posterior["α"]
beta = idata_lb.posterior["β"]
scale = idata_lb.posterior["scale"]


mut = np.exp(alpha + beta.sel(β_dim_0=0) * xr.DataArray(bikes.temperature, dims="μ_t_dim_0") + beta.sel(β_dim_0=1) * bikes.humidity.mean())
muh = np.exp(alpha + beta.sel(β_dim_0=0) * bikes.temperature.mean() + beta.sel(β_dim_0=1) * xr.DataArray(bikes.humidity, dims="μ_h_dim_0"))

idata_lb.posterior["μ_t"] = mut
idata_lb.posterior["μ_h"] = muh

# Compute marginal predictions
y_pred_t = XrDiscreteRV(pz.NegativeBinomial, mut, scale).rvs()
y_pred_h = XrDiscreteRV(pz.NegativeBinomial, muh, scale).rvs()

idata_lb.posterior_predictive["y_pred_t"] = y_pred_t
idata_lb.posterior_predictive["y_pred_h"] = y_pred_h

dt_cd = azb.from_dict(
    {"constant_data": {
        "temperature": bikes.temperature.to_numpy(),
        "humidity": bikes.humidity.to_numpy(),
    }},
    dims={"temperature": ["dim_0"], "humidity": ["dim_0"]},
)
idata_lb["constant_data"] = dt_cd["constant_data"].to_dataset()
```
After that, we can plot the marginal values i.e., how the mean of the rented bikes changes with temperature when we keep the humidity at its mean value. and the other way around

```python
pc = azp.plot_lm(idata_lb,
                x=["temperature", "humidity"],
                y=["y_pred_t", "y_pred_h"],
                y_obs="y_pred",
                visuals={"ci_band":{},  "pe_line":{}, "observed_scatter":False},
                )
```

We may want to plot the posterior means, instead of predictions

```python
pc = azp.plot_lm(idata_lb,
                group="posterior",
                x=["temperature", "humidity"],
                y=["μ_t", "μ_h"],
                y_obs="y_pred",
                visuals={"ci_bounds":True, "pe_line":{}, "observed_scatter":False},
                )
``` 

----
📚 Documentation preview 📚: https://arviz-plots--343.org.readthedocs.build/en/343/

### DIFF
--- a/src/arviz_plots/plots/lm_plot.py
+++ b/src/arviz_plots/plots/lm_plot.py
@@ -54,7 +54,7 @@ def plot_lm(
             "pe_line",
             "ci_band",
             "ci_bounds",
-            "ci_line_y",
+            "ci_vlines",
             "observed_scatter",
             "xlabel",
             "ylabel",
@@ -66,7 +66,7 @@ def plot_lm(
             "pe_line",
             "ci_band",
             "ci_bounds",
-            "ci_line_y",
+            "ci_vlines",
             "observed_scatter",
             "xlabel",
             "ylabel",
@@ -74,7 +74,7 @@ def plot_lm(
         Mapping[str, Any] | Literal[False],
     ] = None,
     stats: Mapping[
-        Literal["credible_interval", "point_estimate", "smooth"],
+        Literal["credible_interval", "pe_line", "smooth"],
         Mapping[str, Any] | xr.Dataset,
     ] = None,
     **pc_kwargs,
@@ -85,11 +85,13 @@ def plot_lm(
     ----------
     dt : DataTree
         Input data
-    x : str, optional
+    x : str or sequence of str, optional
         Independent variable. If None, use the first variable in group.
         Data will be taken from the constant_data group unless the `group` argument is
         "predictions" in which case it is taken from the predictions_constant_data group.
-    y : str, optional
+
+        The plots and visuals in the generated ``PlotCollection`` object will use `x` for naming.
+    y : str or sequence of str, optional
         Response variable or linear term. If None, use the first variable in observed_data group.
     y_obs : str or DataArray, optional
         Observed response variable. If None, use `y`.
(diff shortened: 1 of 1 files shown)


aloctavodia:
It is doing similar things. Maybe we don't need to have a separate function like `plot_hdi` if this is easy to use and flexible enough.

juanitorduz:
Agree! This looks great! Thank you!

OriolAbril (commented):
I like most of the changes but we need more examples both to showcase that in the docs and for user testing.

aloctavodia:
I have addressed some of your comments @OriolAbril, the easy ones. Regarding what to do with the `combine` function, I am not entirely sure what the best approach is, but I agree that the current one feels hacky. We could have a meeting to discuss details, or maybe you can try an alternative yourself, or both.

src/arviz_plots/plots/lm_plot.py
```diff
@@ -1,91 +1,95 @@
...
-        If None (default), the variable in the specified group with the same name as y is used.
-    x_pred : str or list of str or DataArray or Dataset, optional
-        Independent variable(s) for predictions.
-
-        If None (default), and if group is "predictions", all variables corresponding to x data
-        in "predictions_constant_data" group are used. If group is "posterior_predictive",
-        x is used.
+    x : str  optional
+        Independent variable. If None, use the first variable in constant_data group.
+    y : str optional
```
tomicapretto:
```suggestion
    y : str, optional
```

src/arviz_plots/plots/lm_plot.py
```diff
@@ -1,91 +1,95 @@
...
+        Response variable or linear term. If None, use the first variable in observed_data group.
+    y_obs : str or DataArray, optional
```
tomicapretto: I think it would be good to explain the difference between `y` and `y_obs` a bit more.  They're looked up in different groups, right? I would mention it. 

Also, why does it make different assumptions about `y` and `y_obs`? I can imagine it's related to variables available in the different groups, but I'm not sure.

src/arviz_plots/plots/lm_plot.py
```diff
@@ -115,10 +119,30 @@ def plot_lm(
     visuals : mapping of {str : mapping or bool}, optional
         Valid keys are:
 
-        * ci_line -> passed to :func:`~.visuals.line_xy`. Defaults to False
-        * central_line -> passed to :func:`~.visuals.line_xy`.
-        * ci_fill -> passed to :func:`~.visuals.fill_between_y`.
-        * scatter -> passed to :func:`~.visuals.scatter_xy`.
+        * pe_line-> passed to :func:`~.visuals.line_xy`.
```
tomicapretto: Are spaces between lines and around `->` meaningful for doc building? Asking to make sure things are lay out as expected.

src/arviz_plots/plots/lm_plot.py
```diff
@@ -165,76 +189,71 @@ def plot_lm(
...
-    if isinstance(y, xr.Dataset):
-        raise TypeError(
-            "y can't be a dataset because multiple target variables are not supported yet."
-        )
-
-    if not isinstance(y, xr.DataArray):
-        y = process_group_variables_coords(
-            dt, group="observed_data", var_names=y, filter_vars=filter_vars, coords=coords
-        )
+        y = list(obs_data.data_vars)[:1]
```
tomicapretto: What happens with `x` and `y` if either `observed_data` or `constant_data` do not exist? Would `get_group` work? Maybe it makes sense to first check availability of the groups, then availability of data vars within the groups, and then raise appropriate errors if needed?

src/arviz_plots/plots/lm_plot.py
```diff
@@ -269,20 +289,20 @@ def plot_lm(
     aes_by_visuals.setdefault(
         "central_line", plot_collection.aes_set.difference({"alpha", "color"})
     )
-    if isinstance(ci_prob, Sequence):
-        aes_by_visuals.setdefault("ci_line", {"alpha"})
+    if isinstance(ci_prob, (list | tuple | np.ndarray)):
```
tomicapretto: I see this check in several places. As far as I understand, you use it to act different in the presence of multiple intervals instead of one.

Would it make sense to convert `ci_prob` to a one-dimensional numpy array early on, and then always act as if they are many? Perhaps something custom can be done if `len(arr) == 1`.

src/arviz_plots/plots/lm_plot.py
```diff
@@ -298,125 +318,212 @@ def plot_lm(
...
 
     return plot_collection
+
+
+def combine(x_pred, pe_value, ci_data, x_vars, y_vars, smooth, smooth_kwargs):
+    """
+    Combine and sort x_pred, pe_value, ci_data into a dataset.
+
+    The resulting dataset will have a dimension plot_axis=['x','y','y_bottom','y_top'],
+    and will sort each variable by its x values, and optionally smooth along dim_0.
```
tomicapretto: I think `dim_0` is wrong?

OriolAbril:
I did some experiments, and I like the new API, I am trying to use xarray more to simplify a bit the smoothing and data wrangling.

My main doubt (whose answer might be use a different plot) was about trying to use `plot_lm` for posterior predictive checks. After noticing some calibration issues, could it be useful in some scenarios to plot with 

```
x=["temperature", "humidity"],
y=["y_pred", "y_pred"],  # won't work now but I think it wouldn't be too hard to make work with the changes I am adding
```

so we can compare the bands that come from the posterior predictive with the observations and see if we can identify specific regions in temperature or humidity where the model doesn't work well?

PR #345, opened 2025-09-16
update default ci_prob for ecdf-pit plots

aloctavodia:
These plots are diagnostic plots, instead of summary plots.  And hence the role of `ci_prob` is slightly different from that in the other plots. I suggest we use the value 0.99, while arbitrary, it matches the one used in bayesplot.

Alternatively, we could rename the argument `ci_prob`  for these plots and add a matching entry in rcParams. Maybe something like  `env_prob` or `envelope_prob` and take the default value from there.


aloctavodia:
I wanted your opinion @OriolAbril, not Copilot's

Issue #353, opened 2025-10-10
Add Pareto shape plot


jordandeklerk:
Should we add the Pareto shape plot like in old ArviZ https://python.arviz.org/en/stable/examples/plot_khat.html?

PR #356, opened 2025-10-21
Add Pareto shape plot

jordandeklerk:
Closes https://github.com/arviz-devs/arviz-plots/issues/353

```python
from arviz_plots import plot_khat
from arviz_stats import loo

dt = load_arviz_data("radon")
elpd_data = loo(dt, pointwise=True)
plot_khat(elpd_data)
```

image

```python
plot_khat(elpd_data, show_bins=True, show_hlines=True)
```

image

----
📚 Documentation preview 📚: https://arviz-plots--356.org.readthedocs.build/en/356/

### DIFF
--- a/src/arviz_plots/plots/khat_plot.py
+++ b/src/arviz_plots/plots/khat_plot.py
@@ -8,11 +8,11 @@ import numpy as np
 import xarray as xr
 from arviz_base import rcParams
 from arviz_base.labels import BaseLabeller
+from arviz_stats.base.stats_utils import calculate_khat_bin_edges
 
 from arviz_plots.plot_collection import PlotCollection
 from arviz_plots.plots.utils import (
     annotate_bin_text,
-    calculate_khat_bin_edges,
     enable_hover_labels,
     filter_aes,
     format_coords_as_labels,
@@ -305,24 +305,6 @@ def plot_khat(
             )
 
     scalar_ds = xr.Dataset({"pareto_k": xr.DataArray(0)})
-    khat_kwargs = get_visual_kwargs(visuals, "khat")
-    if khat_kwargs is not False:
-        _, khat_aes, khat_ignore = filter_aes(plot_collection, aes_by_visuals, "khat", [])
-
-        if "color" not in khat_aes:
-            khat_kwargs.setdefault("color", color if color is not None else "C0")
-
-        if "marker" not in khat_aes and marker is not None:
-            khat_kwargs.setdefault("marker", marker)
-
-        plot_collection.map(
-            scatter_xy,
-            "khat",
-            data=khat_dataset,
-            ignore_aes=khat_ignore,
-            **khat_kwargs,
-        )
-
     hlines_kwargs = get_visual_kwargs(visuals, "hlines", default=False)
     if hlines_kwargs is not False and hline_values:
 
@@ -348,6 +330,24 @@ def plot_khat(
                 ignore_aes="all",
                 **hline_kwargs,
             )
+
+    khat_kwargs = get_visual_kwargs(visuals, "khat")
+    if khat_kwargs is not False:
+        _, khat_aes, khat_ignore = filter_aes(plot_collection, aes_by_visuals, "khat", [])
+
+        if "color" not in khat_aes:
+            khat_kwargs.setdefault("color", color if color is not None else "C0")
+
+        if "marker" not in khat_aes and marker is not None:
+            khat_kwargs.setdefault("marker", marker)
+
+        plot_collection.map(
+            scatter_xy,
+            "khat",
+            data=khat_dataset,
+            ignore_aes=khat_ignore,
+            **khat_kwargs,
+        )
 
     bin_text_kwargs = get_visual_kwargs(visuals, "bin_text", default=False)
     if bin_text_kwargs is not False:
--- a/src/arviz_plots/plots/utils.py
+++ b/src/arviz_plots/plots/utils.py
@@ -507,50 +507,6 @@ def format_coords_as_labels(data, skip_dims=None, labeller=None):
     return np.array(labels, dtype=object)
 
 
-def calculate_khat_bin_edges(values, thresholds, tolerance=1e-9):
-    """Calculate bin edges for Pareto k diagnostic bins.
-
-    Parameters
-    ----------
-    values : array_like
-        Pareto k values to bin
-    thresholds : sequence of float
-        Diagnostic threshold values to use as potential bin edges (e.g., [0.7, 1.0])
-    tolerance : float, default 1e-9
-        Numerical tolerance for edge comparisons to avoid duplicate edges
-
-    Returns
-    -------
-    bin_edges : list of float or None
-        Calculated bin edges suitable for np.histogram, or None if edges cannot
-        be computed.
-    """
-    if not values.size:
-        return None
-
-    ymin = np.nanmin(values)
-    ymax = np.nanmax(values)
-
-    if not (np.isfinite(ymin) and np.isfinite(ymax)):
-        return None
-
-    bin_edges = [ymin]
-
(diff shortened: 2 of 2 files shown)


src/arviz_plots/plots/utils.py
jordandeklerk: Not sure if the new helpers here should be in the public API or not. Not too sure how frequently these will be used outside of the Pareto shape plot.

src/arviz_plots/plots/khat_plot.py
```diff
@@ -0,0 +1,557 @@
...
+
+    if show_bins:
+        bin_text_kwargs = get_visual_kwargs(visuals, "bin_text")
+        if bin_text_kwargs is not False:
+            _, bin_text_aes, _ = filter_aes(plot_collection, aes_by_visuals, "bin_text", [])
+            if "color" not in bin_text_aes:
+                bin_text_kwargs.setdefault("color", "B1")
+            bin_text_kwargs.setdefault("horizontal_align", "center")
+
+            bin_edges = calculate_khat_bin_edges(khat_values, [good_k_threshold, 1.0])
```
aloctavodia: All this computatiob logic should be in arviz-stats. There were have some functions that are not exposed to user and only serve arviz-plots

src/arviz_plots/plots/khat_plot.py
```diff
@@ -0,0 +1,557 @@
...
+
+        if backend == "matplotlib" and "color" not in khat_aes:
+            base_color = khat_kwargs.pop("color", default_color)
+            point_rgba = alpha_scaled_colors(base_color, khat_data.values, good_k_threshold)
+
+            if fillable_marker:
+                khat_kwargs.setdefault("facecolor", point_rgba)
+                khat_kwargs.setdefault("edgecolor", point_rgba)
+            else:
+                khat_kwargs.setdefault("color", point_rgba)
```
OriolAbril: all functions that take facecolor and edgecolor also take color as an alias to set both at the same time, so we can skip this check without repercusions

jordandeklerk:
> I have to leave now, will continue the review tomorrow hopefully it is somewhat helpful already

Very useful, thank you! I will need to re-work somethings you've mentioned already and some that @aloctavodia mentioned as well over slack. Hoping to get these changes in early next week!

src/arviz_plots/plots/utils.py
```diff
@@ -453,3 +455,270 @@ def add_bands(
         plot_collection.map(plot_func, "ref_band", data=ref_ds, ignore_aes=ref_ignore, **ref_kwargs)
 
     return plot_collection
+
+
+def format_coords_as_labels(data, skip_dims=None):
```
jordandeklerk: This is the same formatting util from legacy ArviZ. Do we need to adpat this in any way for arviz-plots? This seems to work as it is now.

src/arviz_plots/plots/utils.py
```diff
@@ -453,3 +455,270 @@ def add_bands(
...
+    if len(coord_labels) == 0:
+        return np.array([], dtype=object)
+
+    if isinstance(coord_labels[0], tuple):
+        fmt = ", ".join(["{}" for _ in coord_labels[0]])
+        return np.array([fmt.format(*x) for x in coord_labels], dtype=object)
+    return np.array([f"{s}" for s in coord_labels], dtype=object)
+
+
+def calculate_khat_bin_edges(values, thresholds, tolerance=1e-9):
```
jordandeklerk: Should this live in arviz-stats since it's computational? Could go here if we need to move it https://github.com/arviz-devs/arviz-stats/blob/main/src/arviz_stats/utils.py.

src/arviz_plots/plots/khat_plot.py
```diff
@@ -0,0 +1,512 @@
...
+                    h_kwargs.setdefault("alpha", 0.7)
+
+                h_ds = xr.Dataset({"pareto_k": xr.DataArray(value)})
+                plot_collection.map(
+                    hline,
+                    f"hline_{idx}",
+                    data=h_ds,
+                    ignore_aes="all",
+                    **h_kwargs,
+                )
```
OriolAbril: we should define a dataarray/dataset with a `bin` or `hline` related dimension (kind of like a concatenated version of the `h_ds` ones). Then add an overlay aesthetic to it. Otherwise if we manually loop here I don't see how it would be possible to add an aesthetic mapping to the lines.

Another option is deciding we won't allow aesthetic mappings for this element in which case we can simplify the `aes_by_visual` dict and docs as well as all the `in hline_aes` checks because we do `ignore_aes="all"`

OriolAbril (commented):
I am a bit unsure about the things we'd want to support and the things we don't want to allow. There are several things we could integrate more with plotcollection machinery which would give more flexibility to the plot, but plotcollection also usually gives more flexibility than what is really needed which can end up meaning more work and less clear behaviour.

src/arviz_plots/plots/utils.py
jordandeklerk: I think they are different enough to leave this here

src/arviz_plots/plots/khat_plot.py
```diff
@@ -0,0 +1,512 @@
...
+    set_xticks,
+)
+
+
+def plot_khat(
+    elpd_data,
+    threshold=None,
+    show_hlines=False,
+    show_bins=False,
+    hover_label=False,
```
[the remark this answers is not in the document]
jordandeklerk: I like removing them as top level arguments. I feel like that's more inline with the other plots.

src/arviz_plots/plots/khat_plot.py
jordandeklerk: Let me know if you think the most recent change for this is good.

src/arviz_plots/plots/khat_plot.py
```diff
@@ -0,0 +1,557 @@
...
+
+        plot_collection.map(
+            scatter_xy,
+            "khat",
+            data=khat_dataset,
+            ignore_aes=khat_ignore,
+            **khat_kwargs,
+        )
+
+    if show_hlines and hline_values:
```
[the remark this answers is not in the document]
jordandeklerk: Will changing this to `zorder=3` resolve that? I think that's what I had before, but changed it based on feedback here. I could be wrong though. I think the problem is that scatter points have `zorder=2` as well so which ever is drawn first in `plot_khat()` takes priority. 

Or should I just keep this `zorder` for `hlines` and just plot the `hlines` before the scatter in the actual function so they overlay on top?

```python
@expand_aesthetic_aliases
def hline(y, target, *, color=unset, alpha=unset, width=unset, linestyle=unset, **artist_kws):
    """Interface to matplotlib for a horizontal line spanning the whole axes."""
    artist_kws.setdefault("zorder", 2)
    kwargs = {"color": color, "alpha": alpha, "linewidth": width, "linestyle": linestyle}
    return target.axhline(y, **_filter_kwargs(kwargs, Line2D, artist_kws))
```

docs/source/gallery/model_comparison/04_plot_khat_facet_grid.py
```diff
@@ -0,0 +1,48 @@
+"""
```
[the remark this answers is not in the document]
jordandeklerk: I think the current approach we're using for `plot_khat` is a good balance personally. I think for functions like `plot_khat`, having a simple example in the docstring plus more detailed gallery examples makes sense because the gallery can showcase visual variations without cluttering the docstrings.

Maybe the key is to prioritize based on how users will interact with each function. Functions that are commonly used or have interesting visual diversity benefit from gallery examples, while more specialized or utility functions can rely on docstring examples alone. This would keep the standard of having a simple usage example in the docstring and then more variations in the gallery if the function emits different visual facets.

PR #373, opened 2025-11-20
Improve usability of style functions

aloctavodia:
This is intended to improve the usability of the `style` functions.

* Now `style.available` included an entry `"common"` that lists the styles that are intended to work across backends.  
* The docstring are more explicit
* Remove hard-coded styles for bokeh
* Extend get to work with bokeh.

### DIFF
--- a/src/arviz_plots/style.py
+++ b/src/arviz_plots/style.py
@@ -10,7 +10,7 @@ def use(name):
     The style will be set for all backends that support it and have it available.
     The supported backends are Matplotlib, Plotly and Bokeh.
     You can use ``arviz_plots.style.available()`` to check which styles are available.
-    The ones that works for all backends are listed under the 'common' key.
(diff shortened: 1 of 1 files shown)


src/arviz_plots/style.py
```diff
@@ -1,16 +1,27 @@
...
 from arviz_base import rcParams
 
 
 def use(name):
     """Set an arviz style as the default style/template for all available backends.
 
+    The style will be set for all backends that support it and have it available.
+    The supported backends are Matplotlib, Plotly and Bokeh.
+    You can use ``arviz_plots.style.available()`` to check which styles are available.
+    The ones that works for all backends are listed under the 'common' key.
```
aloctavodia:
```suggestion
    The ones that work for all backends are listed under the 'common' key.
```

PR #379, opened 2025-11-25
Implemented Joint Plot in arviz-plots architecture

PoojasPatel013:
## Joint Plot
This PR introduces the `plot_joint` function to the `arviz-plots` library. This function is a critical component of the ArviZ plotting suite, allowing users to visualize the joint distribution of two variables along with their marginal distributions. This implementation leverages the new `PlotMatrix` architecture to ensure flexibility and consistency with the v1.0 refactor.

joint_plot_test

Here is the output generated by the test script using the new plot_joint implementation.

## What was Added
- **`src/arviz_plots/plots/joint_plot.py`**: A new module containing the `plot_joint` function.
    - **Functionality**: Creates a 2x2 grid layout using `PlotMatrix`.
    - **Joint Distribution**: Plots a scatter plot (or other 2D representations) in the main panel (bottom-left).
    - **Marginal Distributions**: Plots marginal densities (KDE, Hist, ECDF) in the side panels (top-left for X, bottom-right for Y).
        - **Rotated Marginals**: Implemented custom logic (`line_rotated`, `fill_between_rotated`) to correctly plot the Y-variable marginal on the right side.
        - *Note*: I used `fill_betweenx` directly on the target (Matplotlib Axes). In the future, we might want to expose this method in the backend interface wrapper to maintain strict backend agnosticism.
    - **Integration**: Fully integrated with `arviz-plots` backend and aesthetics system.
    - **Safety**: Added checks for variable count and handled empty plots (top-right).

## What was Changed
- **`src/arviz_plots/plots/__init__.py`**: Modified to import and expose `plot_joint` at the top level of the plots module.
- **Architecture Usage**: The implementation demonstrates how to use `PlotMatrix` for non-square, custom grids (2x2 with marginals), extending the usage patterns of the core library.

## Usage
```python
from arviz_plots import plot_joint

dt = load_arviz_data("centered_eight")
plot_joint(dt, var_names=["mu", "tau"])
```
This will create a 2x2 grid with:
- `mu` vs `tau` scatter plot in the bottom-left.
- `mu` marginal density (KDE) on the top-left.
- `tau` marginal density (KDE, rotated) on the bottom-right.
- Top-right axis removed.

## Why
- **Feature Parity**: `plot_joint` is a standard and popular plot in the legacy `arviz` library. Porting it to `arviz-plots` is essential for the "ArviZ 1.0" release.
- **Modular Design**: By implementing this in the new system, we ensure that joint plots benefit from the performance and flexibility improvements of the refactor, such as backend agnosticism (Matplotlib, Bokeh, etc.) and fit well with `arviz-stats`.
- **Custom Grid Layouts**: This implementation serves as a reference for creating non-square, custom grid layouts using `PlotMatrix`, demonstrating its flexibility beyond simple pair plots or facet grids.

## Verification
- Ran `test_joint_plot.py` which successfully generated a joint plot using synthetic data.
- Verified that the `PlotMatrix` initializes correctly without errors.
- Verified that marginals are plotted correctly (X on top, Y on right, and rotated).

### DIFF
--- a/src/arviz_plots/plots/joint_plot.py
+++ b/src/arviz_plots/plots/joint_plot.py
@@ -112,7 +112,63 @@ def plot_joint(
     stats: Mapping[Literal["dist"], Mapping[str, Any] | xr.Dataset] = None,
     **pc_kwargs,
 ):
-    """Plot a joint distribution of two variables with marginals."""
+    """
+    Plot a joint distribution of two variables with marginals.
+
+    This function creates a 2x2 grid (PlotMatrix) where the main panel (bottom-left)
+    displays the joint relationship (e.g., scatter plot) and the side panels
+    display the marginal distributions (e.g., KDE or Histogram).
+
+    Parameters
+    ----------
+    dt : DataTree or InferenceData
+        Input data containing the posterior or prior groups.
+    var_names : list of str, optional
+        Variables to be plotted. Must be exactly two variables.
+    filter_vars : str, optional
+        Regex expression to select variables.
+    group : str, optional
+        Group to be plotted. Defaults to "posterior".
+    coords : dict, optional
+        Coordinates to slice the data.
+    sample_dims : list of str, optional
+        Dimensions to be treated as samples (e.g., ["chain", "draw"]).
+    kind : {"scatter", "kde", "hist"}, default "scatter"
+        Type of plot for the joint distribution (main panel).
+    marginal_kind : {"dist", "kde", "hist"}, default "dist"
+        Type of plot for the marginal distributions (side panels).
+    plot_matrix : PlotMatrix, optional
+        Existing PlotMatrix object to plot onto.
+    backend : str, optional
+        Backend to use (e.g., "matplotlib", "bokeh").
+    labeller : Labeller, optional
+        Class to handle plot labeling.
+    aes_by_visuals : dict, optional
+        Mapping of visual elements to aesthetics.
+    visuals : dict, optional
+        Kwargs passed to the visual functions (e.g., color, alpha).
+    stats : dict, optional
+        Pre-computed statistics for the plots.
+    **pc_kwargs
+        Additional keyword arguments passed to the PlotMatrix constructor.
+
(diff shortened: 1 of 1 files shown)


PoojasPatel013:
I've implemented plot_joint directly in src/arviz_plots/plots/ to match the existing library structure. Let me know if you'd prefer any changes to the file organization.

PoojasPatel013:
> The output of
> 
> ```python
> from arviz_plots import plot_pair
> from arviz_base import load_arviz_data
> 
> dt = load_arviz_data("centered_eight")
> plot_pair(dt, var_names=["mu", "tau"])
> ```
> 
> is output
> 
> `plot_pair` already uses `PlotMatrix`

> Thanks for pointing this out and sharing the image\! I see now that `plot_pair` in the new architecture is versatile enough to handle the asymmetrical layout and marginals.
>
> However, I believe `plot_joint` should still exist as a distinct function for **API parity** and **User Experience**, even if the internal logic overlaps.
>
> **The Case for plot\_joint:**
>
> 1.  **API Discovery:** Users migrating from ArviZ 0.x specifically look for `plot_joint` to analyze 2-variable relationships.
> 2.  **Enforced Constraints:** `plot_joint` strictly validates that exactly 2 variables are passed, preventing accidental N\*N matrix generation.
> 3.  **Opinionated Defaults:** It defaults to the specific "Scatter + Marginals" view without requiring the user to manually configure grid ratios or toggle `marginals=True`.
>
> **Proposed Path Forward:**
> Since `plot_pair` already handles the heavy lifting of the layout, **should I refactor my implementation to be a lightweight wrapper around `plot_pair`?**
>
> This would define `plot_joint` as:
>
> ```python
> def plot_joint(dt, **kwargs):
>     # Enforce 2 variables
>     # Set default kwargs for marginals/layout
>     return plot_pair(dt, marginals=True, **kwargs)
> ```
>
> Or do you prefer keeping the implementation separate to allow for distinct visual defaults (e.g. hexbin support) in the future?

PR #383, opened 2025-12-02
Goodness of fit for density visualizations

aloctavodia:
This shows the basic computations described in https://arxiv.org/abs/2503.01509 for assessing the quality of a density representation (histogram, KDE, or quantile dot plot). The method used the PIT evaluated using the computed density. 

EDIT

This implements the functions `plot_dgof` and `plot_dgof_dist`. The first one only includes the diagnostics and the second also includes the density in the left column.

Not happy with the `dgof` name.

depends on https://github.com/arviz-devs/arviz-stats/pull/273

~For the visualization part, I guess we would like to have a two-column plot with the dist on the left and the diagnostic on the right. Something like `plot_trace_dist` and `plot_rank_dist`. To avoid duplicating the computation, we should compute the density once and use that information to both plot it on the left and calculate the visualization on the right.~

~For the computation part, I am not sure if we want the PIT to be included with the output of histogram/kde/qds, or do we want a separate function that uses that output.~

While working on this, I realized we currently don't support `bins="auto"` (for more than one variable), and hence this diagnostic is usually very bad for the default histogram we compute, https://github.com/arviz-devs/arviz-stats/issues/262

```python
dt = azb.from_dict({"posterior": {"mu":pz.Normal(0, 1).rvs((4, 200), random_state=42)}})
```

```python
azp.plot_dgof(dt, kind="hist")
```
hist

```python
azp.plot_dgof(dt, kind="hist", stats={"dist": {"bins":"auto"}})
```
hist_auto

```python
azp.plot_dgof(dt, kind="kde")
```
kde

```python
azp.plot_dgoft(dt, kind="qds")
```

qds

### DIFF
--- a/src/arviz_plots/plots/dgof_dist_plot.py
+++ b/src/arviz_plots/plots/dgof_dist_plot.py
@@ -62,12 +62,14 @@ def plot_dgof_dist(
     var_names : str or list of str, optional
         One or more variables to be plotted.
         Prefix the variables by ~ when you want to exclude them from the plot.
-    filter_vars : {None, “like”, “regex”}, optional, default=None
+    filter_vars : {None, "like", "regex"}, optional, default=None
         If None (default), interpret var_names as the real variables names.
-        If “like”, interpret var_names as substrings of the real variables names.
-        If “regex”, interpret var_names as regular expressions on the real variables names.
+        If "like", interpret var_names as substrings of the real variables names.
+        If "regex", interpret var_names as regular expressions on the real variables names.
     group : str, default "posterior"
         Group to be plotted.
+    coords : dict, optional
+        Coordinates to be used to index data variables.
     sample_dims : str or sequence of hashable, optional
         Dimensions to reduce unless mapped to an aesthetic.
         Defaults to ``rcParams["data.sample_dims"]``
--- a/src/arviz_plots/plots/dgof_plot.py
+++ b/src/arviz_plots/plots/dgof_plot.py
@@ -5,6 +5,7 @@ from typing import Any, Literal
 
 import arviz_stats  # pylint: disable=unused-import
 import xarray as xr
+from arviz_base import rcParams
 from arviz_stats.ecdf_utils import (
     compute_pit_for_histogram,
     compute_pit_for_kde,
@@ -22,7 +23,7 @@ def plot_dgof(
     group="posterior",
     coords=None,
     sample_dims=None,
-    kind="kde",
+    kind=None,
     ci_prob=0.99,
     plot_collection=None,
     backend=None,
@@ -65,12 +66,14 @@ def plot_dgof(
     var_names : str or list of str, optional
         One or more variables to be plotted.
         Prefix the variables by ~ when you want to exclude them from the plot.
-    filter_vars : {None, “like”, “regex”}, optional, default=None
+    filter_vars : {None, "like", "regex"}, optional, default=None
         If None (default), interpret var_names as the real variables names.
-        If “like”, interpret var_names as substrings of the real variables names.
-        If “regex”, interpret var_names as regular expressions on the real variables names.
+        If "like", interpret var_names as substrings of the real variables names.
+        If "regex", interpret var_names as regular expressions on the real variables names.
     group : str, default "posterior"
         Group to be plotted.
+    coords : dict, optional
+        Coordinates to be used to index data variables.
     sample_dims : str or sequence of hashable, optional
         Dimensions to reduce unless mapped to an aesthetic.
         Defaults to ``rcParams["data.sample_dims"]``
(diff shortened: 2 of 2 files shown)


src/arviz_plots/plots/dens_pit_plot.py
```diff
@@ -0,0 +1,251 @@
+"""pit density diagnostic."""
```
tomicapretto:
```suggestion
"""PIT density diagnostic."""
```

tomicapretto:
> Thanks for the comment and discussion.
> 
> The easy thing to do is to have some utility function `arviz-stats` to compute the diagnostics, and then here we should have `plot_x` and then a `plot_x_dist` (like we have, for instance, with `plot_rank` and `plot_rank_dist`). This will give users the main features (the diagnostic plot) and then, in the future, we can think about how to improve the internals and/or expose the computation of the diagnostic to the user if we decide to do so.

Sounds good to me!

tomicapretto:
When I run `tox -e docs` I get the following

```
Extension error (gallery_generator)!

Versions
========

* Platform:         linux; (Linux-5.15.167.4-microsoft-standard-WSL2-x86_64-with-glibc2.35)
* Python version:   3.12.9 (CPython)
* Sphinx version:   8.2.3
* Docutils version: 0.21.2
* Jinja2 version:   3.1.6
* Pygments version: 2.19.2

Last Messages
=============

None.

Loaded Extensions
=================

None.

Traceback
=========

    Traceback (most recent call last):
      File "/home/tomas/oss/arviz-devs/arviz-plots/.tox/docs/lib/python3.12/site-packages/sphinx/events.py", line 404, in emit
        results.append(listener.handler(self.app, *args))
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/tomas/oss/arviz-devs/arviz-plots/docs/sphinxext/gallery_generator.py", line 174, in main
        exec(compile(mpl_noshow_code, basename, "exec"))  # pylint: disable=exec-used
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "plot_ppc_pit", line 8, in <module>
      File "/home/tomas/oss/arviz-devs/arviz-plots/.tox/docs/lib/python3.12/site-packages/arviz_plots/plots/ppc_pit_plot.py", line 216, in plot_ppc_pit
        ds_ecdf = difference_ecdf_pit(
                  ^^^^^^^^^^^^^^^^^^^^
    TypeError: difference_ecdf_pit() got multiple values for argument 'n_simulations'
    
    The above exception was the direct cause of the following exception:
    
    Traceback (most recent call last):
      File "/home/tomas/oss/arviz-devs/arviz-plots/.tox/docs/lib/python3.12/site-packages/sphinx/cmd/build.py", line 414, in build_main
        app = Sphinx(
              ^^^^^^^
      File "/home/tomas/oss/arviz-devs/arviz-plots/.tox/docs/lib/python3.12/site-packages/sphinx/application.py", line 332, in __init__
        self._init_builder()
      File "/home/tomas/oss/arviz-devs/arviz-plots/.tox/docs/lib/python3.12/site-packages/sphinx/application.py", line 414, in _init_builder
        self.events.emit('builder-inited')
      File "/home/tomas/oss/arviz-devs/arviz-plots/.tox/docs/lib/python3.12/site-packages/sphinx/events.py", line 415, in emit
        raise ExtensionError(
    sphinx.errors.ExtensionError: Handler <function main at 0x7f9bd3bdad40> for event 'builder-inited' threw an exception (exception: difference_ecdf_pit() got multiple values for argument 'n_simulations'
```

I think it's because we're passing more arguments than what the function can handle?
However, I'm surprised all checks are green here.

src/arviz_plots/plots/dgof_dist_plot.py
```diff
@@ -0,0 +1,240 @@
...
+    ----------
+    dt : DataTree
+        Input data
+    var_names : str or list of str, optional
+        One or more variables to be plotted.
+        Prefix the variables by ~ when you want to exclude them from the plot.
+    filter_vars : {None, “like”, “regex”}, optional, default=None
+        If None (default), interpret var_names as the real variables names.
+        If “like”, interpret var_names as substrings of the real variables names.
+        If “regex”, interpret var_names as regular expressions on the real variables names.
```
tomicapretto:
```suggestion
    filter_vars : {None, "like", "regex"}, optional, default=None
        If None (default), interpret var_names as the real variables names.
        If "like", interpret var_names as substrings of the real variables names.
        If "regex", interpret var_names as regular expressions on the real variables names.
```

tomicapretto:
@aloctavodia, only minor comments.

I don't have better alternatives for the names. At first sight, they looked a bit strange, I was not understanding what they meant. After a moment I realized "gof" is "goodness of fit" and "dgof" is "density goodness of fit"... I think they're OK.

PR #386, opened 2025-12-04
Add quantile dots plot

### LINKED ISSUE #25: Add quantile-dot to plot_dist

aloctavodia:
closes #25 

Depends on https://github.com/arviz-devs/arviz-stats/pull/266

In the future, we may want to use something more flexible than `scatter_xy`, but this already provides basic functionality.

Some tests generate warnings because we sometimes need to use fewer quantiles than the default 100, which is expected. Therefore, the tests don't fail due to these warnings.


docs/source/gallery/distribution/02b_plot_dist_qds.py
OriolAbril: I think I hardcoded removing the first 3 characters of the filename, so now the url gets a leading underscore. We should update the numbers in the other filenames (which won't change the url) or update the gallery generator to have more robust stripping of numbers for sorting.

jordandeklerk (approved):
LGTM! Small, nonbreaking comment: we have this validation in `dist_plot.py`

```python
if kind not in ("kde", "hist", "ecdf", "dot"):
    raise ValueError("kind must be either 'kde', 'hist', 'ecdf' or 'dot'")
```

but we don't have this in other modules like `ppc_dist_plot.py` where we do have the `kind` argument. Might be worthwhile to add to other modules that have this arg.

Issue #387, opened 2025-12-05
plotly legends are not interactive in arviz.


isofer:
one of the feature that plotly supports is to be able to interact with the legend. You can isolate traces by double-clicking them or turn a trace on/off by clicking on them in the legend. For some reason this ability was lost in arviz.
To see it you can simply check the [docs](https://arviz-plots.readthedocs.io/en/latest/gallery/plot_psense.html?backend=plotly#synchronised-tabs). you can see that clicking on the legend doesn't affect the plots, compared to clicking on native plotly legends (for instance, on this[ page](https://plotly.com/python/legend/)).

Issue #68, opened 2024-07-06
Tooltip and grouping information?


OriolAbril:
We might want to provide tooltip and grouping information somehow as part of the common plotting interface. Even if some backends can't include that into the plot, it would be a great win for those that can.

OriolAbril:
Here is an example of getting an interactive legend working with `plot_dist`. As we can pass arbitrary elements as aesthetics and as plot_kwargs everything is feasible already, but we might want to set that behaviour as the default (maybe only when there are multiple models for example, with docs on doing that also for plot_dist with chains as different colors...).

Note: to make that a default the `legendgroup` argument should become part of the common interface (with this or another name) and be ignored for backends where it makes no sense.

[Interactive legend in plot_dist](https://github.com/arviz-devs/arviz-plots/assets/23738400/f9494d11-e80a-47e3-97b6-b79134188979)

<details><summary>Code to reproduce</summary>

At the time of writing, this requires the plotly_poc branch

```python
import numpy as np
import arviz_plots as azp
from arviz_base import load_arviz_data

import plotly.io as pio

pio.renderers.default = "browser"

c = load_arviz_data("centered_eight")
n = load_arviz_data("non_centered_eight")

pc = azp.plot_dist(
    {"centered": c, "non_centered": n},
    backend="plotly",
    pc_kwargs={"aes": {"legendgroup": ["model"]}, "legendgroup": ["centered", "non_centered"]},
    aes_map={
        "credible_interval": {"color", "y", "legendgroup"},
        "point_estimate": {"color", "y", "legendgroup"},
    },
    plot_kwargs={
        "kde": {"showlegend": False},
        "credible_interval": {"showlegend": False},
        "point_estimate": {"showlegend": False},
        "point_estimate_text": {"showlegend": False},
    }
)

# generate interactive legend
colors = pc.aes["mu"]["color"].values
plot = pc.viz["chart"].item()
obj_centered = pc.viz["mu"]["kde"].sel(model="centered").item()
obj_non_centered = pc.viz["mu"]["kde"].sel(model="non_centered").item()
plot.update_traces(
    selector=lambda trace: (trace.y.shape == obj_centered.y.shape) and np.allclose(trace.y, obj_centered.y),
    name="centered",
    legendgroup="centered",
    showlegend=True,
)
plot.update_traces(
    selector=lambda trace: (trace.y.shape == obj_non_centered.y.shape) and np.allclose(trace.y, obj_non_centered.y),
[...]

PR #388, opened 2025-12-08
Plotly legends: support interactivity and multiple legends

OriolAbril:
The goal of the PR is to allow the legend generating code to update visuals at legend creation time.
Still early stages and breaks legend related things for other backends right now,
but I think it can be interesting.

It also has an important caveat which is we need to loop over all traces to update them.
This means it won't work on visuals/traces that aren't stored in the plotcollection,,
and if we know beforehand what we want to do we are much better off adding a `legendgroup`
aes from the start instead of doing this update step.

related to #387 and #68

### DIFF
--- a/src/arviz_plots/backend/plotly/legend.py
+++ b/src/arviz_plots/backend/plotly/legend.py
@@ -1,8 +1,10 @@
 """Plotly legend generation."""
+from functools import partial
+
 import numpy as np
 import xarray as xr
 from arviz_base import xarray_sel_iter
-from plotly.graph_objects import Bar, Contour, Heatmap, Scatter
+from plotly.graph_objects import Bar, Scatter
 
 from .core import expand_aesthetic_aliases
 
@@ -12,6 +14,19 @@ def dealiase_line_kwargs(**kwargs):
     """Convert arviz common interface properties to plotly ones."""
     prop_map = {"linewidth": "width", "linestyle": "dash"}
     return {prop_map.get(key, key): value for key, value in kwargs.items()}
+
+
+def _trace_matcher(trace, target_viz):
+    """See if a plotly trace matches a target trace `target_viz`."""
+    return (
+        (getattr(trace, "mode", "na") == getattr(target_viz, "mode", "na"))
+        and (getattr(trace, "line", "na") == getattr(target_viz, "line", "na"))
+        and (trace.marker == target_viz.marker)
+        and (trace.text == target_viz.text)
+        and (trace.x.shape == target_viz.x.shape)
+        and np.allclose(trace.y, target_viz.y)
+        and np.allclose(trace.x, target_viz.x)
+    )
 
 
 LINE_SUBKEYS = [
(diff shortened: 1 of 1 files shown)


src/arviz_plots/backend/plotly/legend.py
```diff
@@ -36,25 +63,84 @@ def legend(
     None
         The legend is added to the target figure inplace
     """
-    if artist_kwargs is None:
-        artist_kwargs = {}
+    figure = plot_collection.get_viz("figure")
+    if "legend" in plot_collection.viz.children:
```
tomicapretto:
```suggestion
    # NOTE: Legend IDs in Plotly must be 'legend', 'legend2', 'legend3', etc.
    if "legend" in plot_collection.viz.children:
```

src/arviz_plots/backend/plotly/legend.py
```diff
@@ -36,25 +63,84 @@ def legend(
...
+        legend_number = len(plot_collection.viz["legend"].data_vars) + 1
+        legend_id = f"legend{legend_number}"
+    else:
+        legend_number = 1
+        legend_id = "legend"
+    kwargs.setdefault("legend_y", {1: 1, 2: 0, 3: 0.5}[legend_number])
+    kwargs["legend_title_text"] = title
+    legend_kwargs = kwargs.pop(legend_id, {}).copy()
+    kwargs_list = list(kwargs.items())
+    for key, value in kwargs_list:
```
tomicapretto: Would this be a bit more explicit?
```suggestion
    for key, value in kwargs.copy().items():
```

src/arviz_plots/backend/plotly/legend.py
```diff
@@ -36,25 +63,84 @@ def legend(
...
+                        (getattr(trace, "mode", "na") == getattr(target_viz, "mode", "na"))
+                        and (getattr(trace, "line", "na") == getattr(target_viz, "line", "na"))
+                        and (trace.marker == target_viz.marker)
+                        and (trace.text == target_viz.text)
+                        and (trace.x.shape == target_viz.x.shape)
+                        and np.allclose(trace.y, target_viz.y)
+                        and np.allclose(trace.x, target_viz.x)
+                    )
+
+                target_plot = plot_collection.get_target(var_name, sel)
```
tomicapretto: Is this `.get_target(var_name, sel)` similar to `viz_ds[var_name].sel(sel).item()`?

I imagine they're not exactly the same, but close.
It would be good to have a comment explaining what the bigger loop does. 
Something along the lines of "Updates the legend group for visuals in the plot that share this and that, only when X and Y"

src/arviz_plots/plot_collection.py
```diff
@@ -1268,6 +1268,15 @@ def add_legend(
             dim = (dim,)
         else:
             dim = tuple(dim)
+        update_visuals = False
+        if "legendgroup" not in self.aes.children and self.backend == "plotly":
+            # TODO: keep if to avoid duplicating legendgroup but don't make it plotly specific
+            # also, should we add an "interactive" argument to disable this behaviour
+            # even if it would be possible?
```
tomicapretto: I would consider it after we consider the legend implementation to be done.

src/arviz_plots/backend/none/legend.py
```diff
@@ -6,20 +6,38 @@
...
-    artist_type : {"line", "scatter", "rectangle"}, default "line"
-    artist_kwargs : mapping, optional
+    visual_type : {"line", "scatter", "rectangle"}, default "line"
+    visual_kwargs : mapping, optional
         Passed to all visuals when generating legend miniatures.
+    legend_dim : str or sequence of str, optional
+        Dimension or dimensions whose mappings should be used to generate the legend.
+    update_visuals : bool, optional
+        If relevant for the backend, update objects representing :term:`visual` elements
+        of the plot to improve or allow interactivity for the legend.
```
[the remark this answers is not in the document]
OriolAbril: I agree the description needs improving, but the goal of the argument isn't actually related to their visibility. Given we have many visual elements in the whole figure but only a handful of entries in the legend, we want multiple elements to be toggled on/off with the same trigger. So we need to update the `legendgroup` property of the existing objects. This doesn't change anything about the generated figure visually, it only makes clicking on the plotly legend work as expected (for the first legend).

There are cases when we don't want to do this, for example with the 2+ legend in the same figure. But maybe more importantly, if we know beforehand which legend we'll generate we can define the legendgroup property/mapping beforehand which should be more efficient. In such cases we want to skip this step too.

PR #390, opened 2025-12-11
Add bfmi to energy plot

aloctavodia:
~I would like to have some level of formatting for the legend. This is just a suggestion~.

output

EDIT: I changed the plot from the one above to a two-column plot. One reason is that I find it annoying that the old one features a figure with combined chains, but includes information about individual chains. Instead in the new one, all the information is plotted. I also default to face true, because the goal is to evaluate overlap.

output


tomicapretto (approved):
I like it, looks good

PR #393, opened 2025-12-17
Change defaults for plot_ppc_dist

aloctavodia:
This removes the `data_pairs` arguments, which we already agree we don't want.
This also change the default for "hist" (each predictive draw was of a different color and the observed hist was filled) and for "dot", only plots the top dots. For reference, this is how the four kinds look by default.

output_hist

output_dots
output_kde
output_ecdf

~I think we also want to remove the y-axis by default for hist, as now histograms represent densities.~

### DIFF
--- a/src/arviz_plots/plots/dist_plot.py
+++ b/src/arviz_plots/plots/dist_plot.py
@@ -217,7 +217,7 @@ def plot_dist(
         visuals = {}
     else:
         visuals = visuals.copy()
-    if kind in ("hist", "ecdf"):
+    if kind == "ecdf":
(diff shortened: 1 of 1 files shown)


OriolAbril:
> currently on arviz-stats, the bins argument only supports integers

it should allow a dataarray, if it doesn't we should open an issue for it. For plot_ppc_dist purposes I think there are cases where I want the shared bins but also cases where it ends up generating too much overlap between lines and becomes hard to read. That being said, plot_ppc_dist with histograms is generally very hard to read, no matter how the histograms are done

PR #404, opened 2026-01-14
Fix example in docstring for plot_dgof_dist and plot_dgof

aloctavodia:


tomicapretto:
Do you know what could have caused the doc build error?

I've never seen it:

```
File "/home/docs/checkouts/readthedocs.org/user_builds/arviz-plots/envs/404/lib/python3.12/site-packages/arviz_base/datasets.py", line 135, in load_arviz_data
        raise OSError(
    OSError: /home/docs/arviz_data/rugby.nc has an SHA256 checksum (e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855) differing from expected ({remote.checksum}), file may be corrupted. Run `arviz.clear_data_home()` and try again, or please open an issue.
```

aloctavodia:
Yes, figshare has not been working properly. We have been getting different errors, including 403 errors or empty files, like in this case.  Figshare may fix the issue at some point, but it is not entirely clear when. In the meantime, I propose to use an alternative source for files in case figshare fails https://github.com/arviz-devs/arviz-base/pull/127

PR #415, opened 2026-01-21
Add facet_map methods to PlotCollection

### LINKED ISSUE #224: Set y/x_lim
Is there a more direct way of doing this?

```python
crabs_poisson = azb.load_arviz_data('crabs_poisson')
pc = azp.plot_ppc_rootogram(crabs_poisson)
pc.viz["satellite"].plot.item().set_xlim(-0.5, 20)

Shlokpalrecha:
Closes #224

## Summary
Added `facet_map()` method to `PlotCollection` for applying visual functions to specific subsets of plots using the standard `var_names`/`coords` interface.

## Motivation
Previously discussed implementing `set_xlim()` and `set_ylim()` convenience methods, but maintainers suggested a more general approach to avoid the "slippery slope" of adding many specific methods. 

The new `facet_map()` method allows applying ANY visual function (including `set_xlim`, `set_ylim`, or custom functions) to filtered subsets of plots, making it much more flexible and consistent with existing patterns in the library.

### DIFF
--- a/src/arviz_plots/plots/combine.py
+++ b/src/arviz_plots/plots/combine.py
@@ -165,7 +165,7 @@ def combine_plots(
             attrs = ds.attrs
             pc.map(
                 render,
-                fun_label=f"{viz_group}_{name}",
+                f"{viz_group}_{name}",
                 data=ds.dataset,
                 ignore_aes=attrs.get("ignore_aes", frozenset()),
             )


OriolAbril:
Exactly, internally that would create a dataset from the `plot` object in the plotcollection and `var_names`+`coords`;  and then use it as `data` arg to call `.map` along with `ignore_aes="all"`. We can also easily allow the first argument to be a string which is then interpreted as a function in the visuals module for example (maybe even for map itself)

OriolAbril:
@Shlokpalrecha sorry for hijacking the PR. We have been quite swamped so I had no time to review but we want to make sure it gets included in the upcoming release (which will be today or tomorrow). This is definitely not standard practice though. Do not hesitate to ask about any changes I did to the PR. Note you can still review it even if merged.

Issue #417, opened 2026-01-22
Ensure consistent signature


aloctavodia:
From @OriolAbril comment in #413 

It doesn't look like we have been very consistent with this, but I think it might be best to use

    *required_args, var_names, filter_vars..., sample_dims, function_arg1, function_arg2...

In the call signature, whenever var_names, group... make sense to be there. 

We might also want to add the *, at some point, ensure arguments are used as keyword ones instead of positional.

PR #419, opened 2026-01-29
Ensure a consistent signature and enforce most arguments to be keyword-only

AAlexxis222:
This change introduces the *, separator after the required dt parameter in plot_rank, forcing all optional parameters (var_names, filter_vars, etc.) to be passed as keyword arguments.

This is the pilot PR for issue #417 - standardizing function signatures across all plotting functions.

Breaking change: Positional arguments after dt are no longer accepted. Users must use keyword arguments (e.g., plot_rank(data, var_names=['x']) instead of plot_rank(data, ['x'])).

### DIFF
--- a/src/arviz_plots/plots/lm_plot.py
+++ b/src/arviz_plots/plots/lm_plot.py
@@ -38,11 +38,11 @@ def plot_lm(
     y=None,
     y_obs=None,
     plot_dim=None,
-    smooth=True,
     filter_vars=None,
     group="posterior_predictive",
     coords=None,
     sample_dims=None,
+    smooth=True,
     ci_kind=None,
     ci_prob=None,
     point_estimate=None,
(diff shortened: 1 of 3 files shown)


aloctavodia:
I changed the title to make it more general. You can now add more commits to this PR for the rest of the functions. 

Just to be sure we are on the same page, we want this order: 

`*required_args, var_names, filter_vars..., sample_dims, function_arg1, function_arg2...`

For most functions, `required_args` is just `dt`, but there may be some exceptions. For instance, `combine_plots` has `dt` and `plots` as required arguments. 

For many functions, the order is not correct. For instanc, we have `plot_bf(dt, var_names, ref_val=0, kind=None, sample_dims=None, plot_collection, ...)`  but we should have  ``plot_bf(dt, var_names, sample_dims=None, ref_val=0, kind=None, plot_collection, ...)``

src/arviz_plots/plots/lm_plot.py
```diff
@@ -33,6 +33,7 @@
 
 def plot_lm(
     dt,
+    *,
     x=None,
```
aloctavodia: This is a weird one, `smooth` should be in the block after `sample_dims`. The others can stay, I have doubts about `plot_dim` @OriolAbril, what do you think?

    x=None,
    y=None,
    y_obs=None,
    plot_dim=None,
    smooth=True,

PR #435, opened 2026-02-25
Add hypothesis test for plot_compare

### LINKED ISSUE #63: Add hypothesis tests for all test functions
Tests with hypothesis do a very good job of making sure the API works and is coherent. Today for example I realized about #62 which would have been caught by these tests.

PiyushPanwarFST:
Address issue #63 

This PR adds Hypothesis testing for the `plot_compare` function.

- Added `compare_df_strategy` to generate random Pandas DataFrames matching arviz_stats.compare output.
- Implemented `test_plot_compare` covering rotated, relative_scale, and hide_top_model variants.
- Limited ticklabels to only use {} or False because the none backend for ticks doesn't support extra styling like colors and was causing errors.

I have started with `plot_compare` for now. I will follow up with tests for the remaining functions (like `plot_lm` and `combine_plots`) 

closes #63

### DIFF
--- a/tests/test_hypothesis_plots.py
+++ b/tests/test_hypothesis_plots.py
@@ -187,6 +187,9 @@ def test_plot_compare(cmp_df, relative_scale, rotated, hide_top_model, visuals):
         visuals=visuals,
     )
     assert "plot" in pc.viz.data_vars
+    for visual, value in visuals.items():
+        if value is False:
+            assert visual not in pc.viz.data_vars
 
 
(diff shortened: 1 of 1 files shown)


OriolAbril:
Given `xticks` and `yticks` backend function is about both tick location and tick labels they should have both `color` and `fontsize` arguments as part of the call signature for all backends. Otherwise it is inconsistent with https://python.arviz.org/projects/plots/en/latest/api/backend/index.html#module-arviz_plots.backend. Also as they are getting a color argument, the decorator for aesthetic aliases should also be added to the two functions (in all backends except none).

PR #438, opened 2026-02-27
test: comprehensive tests for combine_plots (fixes #236)

### LINKED ISSUE #236: User test and improve `combine_plots`
It would be nice to have someone test it out and find issues with it. Some of them might be "won't fix" but others might be low hanging fruits.

I realized for example that the `sample_dims` argument to `combine_plots` is not passed to the plotting functions it calls so if you try to use `sample_dims` different than the rcParam you need to set it to the combine_plots call and to all specific plot kwargs.

**IMPORTANT NOTE: Why good first issue?** This issue is labeled "good first issue" because working on it doesn't require opening a PR nor reading the ArviZ contributing guide. That being said, the issue is about **user testing** so if you aren't already an intermediate ArviZ user you'll need to become one (consequently learning the basics about [exploratory analysis of bayesian models](https://arviz-devs.github.io/EABM/) in that process) before you can start working on this issue.

jinukuntlaakhilakumargoud-web:
## Summary

Added comprehensive test suite for `combine_plots` which previously had zero tests. This addresses issue #236 ("User test and improve combine_plots").

### Tests added (9 total)

| Test | What it validates |
|------|-------------------|
| `test_combine_basic_column` | Default `expand="column"` layout |
| `test_combine_basic_row` | `expand="row"` layout |
| `test_combine_custom_plot_names` | Custom names appear as coords |
| `test_combine_sample_dims_forwarded` | `sample_dims` reaches child plots |
| `test_combine_invalid_expand` | ValueError on bad expand value |
| `test_combine_multiple_plots` | Three plots in one figure |
| `test_combine_with_kwargs` | Child plot kwargs forwarded |
| `test_combine_ppc_plots` | PPC plots with `group="posterior_predictive"` |
| `test_combine_with_forest_xfail` | Documents `plot_forest` incompatibility |

### Key findings

1. **`sample_dims` forwarding** - Already fixed in the current codebase (line 156 of `combine.py`)
2. 2. **`no_artist_kwargs` incompatibility** - `combine_plots` internally calls child plots with `backend="none"`, so the `no_artist_kwargs` test fixture cannot be used
3. 3. **Grid-based plot incompatibility** (bug) - `plot_forest`, `plot_ridge`, `plot_ess`, and `plot_trace` fail when used with `combine_plots` because their internal grid dimensions conflict with `combine_plots`'s `column`/`row` expand dimension, causing `backend_from_object()` to receive an xarray object instead of a plotting target
### Bug details

The `render()` function in `combine.py` calls `backend_from_object(target)` which inspects `target.__module__` to determine the backend. When grid-based plots (forest, ridge, ess, trace) are combined, the `PlotCollection.map()` call may resolve `target` to an xarray DataArray instead of a matplotlib Axes, causing a `ValueError`.

This is documented via an `xfail` test (`test_combine_with_forest_xfail`).

### DIFF
--- a/tests/test_combine_plots.py
+++ b/tests/test_combine_plots.py
@@ -1,134 +1,159 @@
 # pylint: disable=no-self-use, redefined-outer-name
-"""Tests for the combine_plots function.
+"""Tests for combine_plots — issue #236.
 
-Covers:
-- Basic two-plot combination (column and row expand modes)
-- Custom plot_names
-- sample_dims propagation (key regression: issue #236)
-- Multi-model dict input
-- PlotCollection structure validation
+``combine_plots`` internally renders child plots with ``backend='none'`` before
+mapping them onto a real grid, so the outer backend must be a real plotting
+backend (matplotlib, bokeh, plotly) and the ``no_artist_kwargs`` fixture must be
+omitted.
 """
 import pytest
 
-from arviz_plots import PlotCollection, combine_plots, plot_dist, plot_trace
+from arviz_plots import (
+    PlotCollection,
+    combine_plots,
+    plot_dist,
+    plot_forest,
+    plot_ppc_dist,
+    plot_ppc_pit,
+)
 
 pytestmark = [
     pytest.mark.usefixtures("clean_plots"),
     pytest.mark.usefixtures("check_skips"),
-    pytest.mark.usefixtures("no_artist_kwargs"),
 ]
 
 
-@pytest.mark.parametrize("backend", ["matplotlib", "bokeh", "plotly", "none"])
+@pytest.mark.parametrize("backend", ["matplotlib", "bokeh", "plotly"])
 class TestCombinePlots:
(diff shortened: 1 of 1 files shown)


OriolAbril:
> Could you please guide me on what would be needed to close https://github.com/arviz-devs/arviz-plots/issues/236?

The issue is about user testing. We need people to, from the role of ArviZ users, play around with `combine_plots` in order to find limitations in the design and interesting edge cases to add unit tests for. We should have unit tests form combine_plots too like the ones you are adding, but those can't find design flaws with the function nor highlight interesting combination patterns that are not supported by the function but we then see they should be.

I will update the issue description to be clearer, this is one of the reasons the "good first issue" label can be misleading. This is an extremely easy issue for an ArviZ user as it doesn't require opening a PR or learning about the contributing workflow, but extremely difficult for someone who is not an ArviZ user (even if they are a pro at contributing, PRs, CI...) because it requires first becoming an ArviZ user, then working on the issue.

PR #444, opened 2026-03-01
Scoped strict warnings: alias_utils (backend)

sjiang83:
Hi everyone! Following up on the GSoC DevOps ideas and our discussion in #2526, I wanted to take a small, safe first step toward the “treating warnings as errors” goal.

To keep the review burden light and avoid breaking anything globally, this is scoped strictly to tests/test_backend.py (backend alias utilities).

What this does:

- Configures pytest (via pytest.mark.filterwarnings) to treat DeprecationWarning and FutureWarning as errors only for tests/test_backend.py.
- Adds a new CI job strict-warnings (alias_utils) that runs a focused subset with strict flags: 
      tox -e py312 -- tests/test_backend.py -W error::DeprecationWarning -W error::FutureWarning
- Leaves the existing tox matrix (Python 3.12, 3.13, 3.14) unchanged.
- Adds DEVELOPMENT.md with the minimal pattern to extend this gradually to other areas.

I chose the alias_utils backend layer for this pilot because it’s low-risk, self-contained, and easy to verify. Feedback welcome on this gradual approach.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,22 +7,6 @@ on:
       - "docs/"
 
 jobs:
-  strict-warnings-alias-utils:
-    name: strict-warnings (alias_utils)
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v6
-      - name: Set up Python 3.12
-        uses: actions/setup-python@v6
-        with:
-          python-version: "3.12"
-      - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install tox
-      - name: Run scoped strict warnings
-        run: tox -e py312 -- tests/test_backend.py -W error::DeprecationWarning -W error::FutureWarning
-
   test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


ahartikainen:
Can you find a way to introduce these checks as tests and not change the CI file? I think that adds unnecessary complication to CI process? And how would you run this automatically locally?

OriolAbril (commented):
I agree with not trying to edit all tests at once in a single PR but I am also quite confident (and hopefully not to biased and wrong on this) that the vast majority of tests already emit no warnings.

Consequently, I think the ideal aproach would be:

1. Try adding the warnings as error setting globally in [pyproject.toml like we do in arviz-base](https://github.com/arviz-devs/arviz-base/blob/447713079345ec07d4b3fa2344ce7ad4aa99ef04/pyproject.toml#L160)
2. If all tests continue passing everything is done
3. If some tests fail, add the filterwarnings mark to as many files as possible and report the problematic tests so we know what needs addressing

PR #448, opened 2026-03-03
docs: add kwargs conventions guide for the plots module (#66)

### LINKED ISSUE #66: Define and document conventions on artist naming and keys for xyz_kwargs in plots module

HusseinElhaddad:
Adds a contributing guide documenting the current conventions for `visuals`, `aes_by_visuals`, and `stats` parameters across batteries-included plots. Also fixes a typo in the [get_visual_kwargs](cci:1://file:///mnt/02E440FEE440F607/Hussein/Programming/gsoc26/arviz-plots/src/arviz_plots/plots/utils.py:55:0-78:21) docstring.

Closes #66


aloctavodia:
This is very confusing. The guide describes features that are implemented, with features that have been discussed but not implemented. This is not useful in this state. This should be written by someone familiar with the library. 
Additionally, please keep PR descriptions short, reserve them to explain details that may not be very clear, or to justify decisions, or to ask for help.

PR #450, opened 2026-03-03
Improve docstrings for visual interfaces and fix hline function 

aatifjunaid:
- Improved and standardized docstrings for visual interface functions.
- Fixed incorrect `"function": "vline"` reference in `hline`.

### DIFF
--- a/src/arviz_plots/backend/none/core.py
+++ b/src/arviz_plots/backend/none/core.py
@@ -183,8 +183,12 @@ def set_figure_title(figure, string, *, color=unset, size=unset, **artist_kws):
         Color of the title text.
     size : optional
         Font size of the title.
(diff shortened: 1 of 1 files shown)


src/arviz_plots/backend/none/core.py
```diff
@@ -504,7 +580,35 @@ def text(
     horizontal_align=unset,
     **artist_kws,
 ):
-    """Interface to text annotation inside a plot."""
+    """Interface to a text annotation inside a plot.
+
+    Parameters
+    ----------
+    x, y : float or array-like
```
OriolAbril: here I am quite sure only scalars are valid and arrays or array-like objects would error out.

PR #452, opened 2026-03-04
Add new uniformity test

aloctavodia:
Add new uniformity tests as presented in https://arxiv.org/abs/2603.02928 depends on arviz-devs/arviz-stats/pull/316

The reason that we need this new test is that the previous method, with the envelope, assumes independence. So we could keep using it in those cases, but when there is a correlation, we should use the methods implemented here. 

In the paper, the focus of the discussion and main motivation is LOO-PIT, so in this PR I am changing the default method only for `plot_loo_pit` but these methods could potentially be useful elsewhere (as acknowledged in the paper), like our `dgof` functions (or other ecdf_pit functions), but that requires at least some more testing to be sure. 

With the envelope
output_00

with the new default

output_01


aloctavodia:
The tests are passing locally, not sure what I am missing here.

PR #466, opened 2026-03-19
Add nightlies to tox and fix Hypothesis test deadline Extended issue

### LINKED ISSUE #453: Nightlies tests fail due to matplotlib.style.core deprecation
I tried adding nightlies tox environment for arviz-plots to test future releases.

When running `tox -e nightlies`, tests importing top level packages of arviz_plots fail with:
`AttributeError: module 'matplotlib.style' has no attribute 'core'`

following tests fail 
- tests/test_backend.py
- tests/test_hypothesis_plots.py
- tests/test_plot_collection.py
- tests/test_plot_matrix.py
- tests/test_plots.py

### Reason : 

- In the latest matplotlib pre-release (3.11 (dev)), matplotlib.style.core is deprecated.
**Reference** 
[Next API changes matplotlib.style.core](https://matplotlib.org/devdocs/api/next_api_changes/deprecations/30163-AL.html)

### Questions

- In the places where `matplotlib.style.core` is called , for example, In` arviz_plots/__init__.py`, would replacing `mplstyle.core.USER_LIBRARY_PATHS` and `mplstyle.core.reload_library()` with `mplstyle.USER_LIBRARY_PATHS` and `mplstyle.reload_library()` respectively fix the issue?
- Is it okay to submit the nightlies PR now, even though some tests fail due to recent changes in matplotlib?

### LINKED ISSUE #500: AttributeError: module 'matplotlib.style' has no attribute 'core' on import with matplotlib 3.11.0 pre-release
Importing arviz_plots (and therefore arviz, which does from arviz_plots import *) fails on matplotlib 3.11.0 / pre-release with:

```
AttributeError: module 'matplotlib.style' has no attribute 'core'
Traceback

File ".../arviz_plots/__init__.py", line 37, in <module>
    mplstyle.core.USER_LIBRARY_PATHS.append(_arviz_style_path)
    ^^^^^^^^^^^^^
AttributeError: module 'matplotlib.style' has no attribute 'core'
````

## Cause

arviz_plots/__init__.py reaches into the private matplotlib.style.core submodule via attribute access:

```
from matplotlib.pyplot import style as mplstyle
...
mplstyle.core.USER_LIBRARY_PATHS.append(_arviz_style_path)
mplstyle.core.reload_library()
In matplotlib's current pre-release, matplotlib.style/__init__.py only does from .core import available, context, library, reload_library, use and no longer leaves core bound as an attribute on the parent module, so mplstyle.core raises AttributeError. The surrounding try block catches ImportError, not AttributeError, so the failure is not swallowed.
```

## Suggested fix

Use the public API instead of poking at the private submodule. reload_library is already re-exported from matplotlib.style, and explicitly importing core would also work.

### PR DESCRIPTION
**Pull request by Sakshee-D**
### Description
This PR adds a `nightlies` test environment in `tox.ini` for arviz-plots.

### Key Implementation Details

- Added a nightlies environment in tox.ini.
- `hasattr` check for `mplstyle.USER_LIBRARY_PATHS` has been added for compatibility with newer versions.
- While testing, a `DeadlineExceeded` error was encountered in `tests/test_hypothesis_plots.py::test_plot_dgof_dist`. This was addressed by adding a `@settings(deadline=6000)` decorator to the affected test.

All tests passed locally.
Closes #453

### DIFF
--- a/src/arviz_plots/backend/matplotlib/core.py
+++ b/src/arviz_plots/backend/matplotlib/core.py
@@ -39,7 +39,7 @@ class SquareRootScale(mscale.ScaleBase):
     name = "sqrt"
 
     def __init__(self, axis, **kwargs):  # pylint: disable=unused-argument
-        mscale.ScaleBase.__init__(self, axis)
+        mscale.ScaleBase.__init__(axis, self)
 
     def set_default_locators_and_formatters(self, axis):
         """Set the locators and formatters to default."""
--- a/src/arviz_plots/backend/matplotlib/core.py
+++ b/src/arviz_plots/backend/matplotlib/core.py
@@ -33,13 +33,10 @@ class UnsetDefault:
 unset = UnsetDefault()
 
 
-class SquareRootScale(mscale.ScaleBase):
+class SquareRootBaseScale(mscale.ScaleBase):
     """ScaleBase class for generating square root scale."""
 
     name = "sqrt"
-
-    def __init__(self, axis, **kwargs):  # pylint: disable=unused-argument
-        mscale.ScaleBase.__init__(axis, self)
 
     def set_default_locators_and_formatters(self, axis):
         """Set the locators and formatters to default."""
@@ -87,7 +84,25 @@ class SquareRootScale(mscale.ScaleBase):
         return self.SquareRootTransform()
 
 
-mscale.register_scale(SquareRootScale)
+class SquareRootScale(SquareRootBaseScale):
+    """ScaleBase class for generating square root scale for matplotlib<3.11."""
+
+    def __init__(self, axis):  # pylint: disable=all
+        super().__init__(axis)
+
+
+class SquareRootScale311(SquareRootBaseScale):
+    """ScaleBase class for generating square root scale for matplotlib>=3.11."""
+
+    def __init__(self):  # pylint: disable=all
+        pass
+
+
+try:
+    mscale.LinearScale()  # noqa
+    mscale.register_scale(SquareRootScale311)
+except TypeError:
+    mscale.register_scale(SquareRootScale)
 
 
 def get_background_color():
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Review by OriolAbril — COMMENTED** [review]
I will now push the fix for the square root scale issue. Make sure to pull before addressing the review comments.

File: src/arviz_plots/backend/matplotlib/core.py [hunk 1]
```diff
@@ -422,7 +437,7 @@ def scatter(
         "s": size,
         "marker": marker,
         "alpha": alpha,
-        "c": facecolor,
+        "color": facecolor,
```
OriolAbril: I got an error at some point and changed that, can you check if keeping `c` the nightlies test still pass? I tried several things to it might be something else I did.

OriolAbril (commented):
> However, when running all tests, there are multiple failures ValueErrors (reshape array) unrelated to the "c" change.

I will try to look at this at some point during the week. There might have been a behaviour change in xarray we were not aware was coming. We definitely need to fix these too.

Sakshee-D:
I will also look into the failing tests and trace the chain of failures to see where things are breaking. I will update my findings here.

Sakshee-D:
I looked into the nightly failures. They are caused by a `NumPy` development version change where generic units in np.timedelta64 are deprecated.

> Deprecated since version 2.5: The generic units of [timedelta64](https://numpy.org/devdocs/reference/arrays.scalars.html#numpy.timedelta64) are deprecated since NumPy 2.5 and will raise an error in the future.
[reference ](https://numpy.org/devdocs/reference/arrays.datetime.html)

`xarray` currently uses `np.timedelta64("NaT")` without specifying a unit, which now raises a `DeprecationWarning` . This happens during importing xarray .

Sakshee-D:
Thanks for the fix!
 I will re-run the tests on this PR.

PR #503, opened 2026-05-13
Add KDE2D to plot_pair

### LINKED ISSUE #344: Add 2D KDE plots
I don't think we have 2D KDE contour plots right now. Maybe this could be added to `plot_dist()` via the `kind` arg? We can also address this https://github.com/arviz-devs/arviz/issues/2473.

aloctavodia:
This adds 2D KDE to plot_pair, but also more basic functions like contour and contourf visuals. Close #344 

Depends on https://github.com/arviz-devs/arviz-stats/pull/362

### DIFF
--- a/src/arviz_plots/plots/pair_plot.py
+++ b/src/arviz_plots/plots/pair_plot.py
@@ -359,26 +359,6 @@ def plot_pair(
             **scatter_kwargs,
         )
 
-    contour_kwargs = get_visual_kwargs(visuals, "contour", False)
-    if contour_kwargs is not False:
-        _, contour_aes, contour_ignore = filter_aes(
-            plot_matrix, aes_by_visuals, "contour", sample_dims
-        )
-
-        if "color" not in contour_aes:
-            contour_kwargs.setdefault("color", "C0")
-
-        plot_matrix.map_triangle(
-            _kde_couple,
-            "contour",
-            triangle=triangle,
-            data=distribution,
-            ignore_aes=contour_ignore,
-            levels=levels,
-            sample_dims=sample_dims,
-            **contour_kwargs,
-        )
-
     contourf_kwargs = get_visual_kwargs(visuals, "contourf", False)
     if contourf_kwargs is not False:
         _, contourf_aes, contourf_ignore = filter_aes(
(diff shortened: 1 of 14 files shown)


OriolAbril (commented):
I have not been able to do any progress in trying to find out why plotly's areas and borders do not match even though the input data is aparently the same. Despite that (or due to that) I would not make it a blocker. I think we can merge everything else once we are happy with it, I'll try to review the rest of the things to approve, in the meantime the main thing missing is adding the relevant functions to visuals and [backend](https://github.com/arviz-devs/arviz-plots/blob/main/docs/source/api/backend/index.rst#geoms) docs

aloctavodia:
Yeah, in favor of merging this (possible after some further feedback) and opening an issue to solve the plotly part in the future.

src/arviz_plots/plots/pair_plot.py
```diff
@@ -317,6 +359,48 @@ def plot_pair(
...
+        )
+
+    contour_kwargs = get_visual_kwargs(visuals, "contour", False)
+    if contour_kwargs is not False:
+        _, contour_aes, contour_ignore = filter_aes(
+            plot_matrix, aes_by_visuals, "contour", sample_dims
+        )
+
+        if "color" not in contour_aes:
+            contour_kwargs.setdefault("color", "C0")
```
OriolAbril:
```suggestion
            contour_kwargs.setdefault("color", "B1")
```

I think this default will make more sense in general given for the filling doing nothing means using the viridis cmap. It might also be the reason the example script has the explicit `color=black` piece to avoid the weird looking plot.

src/arviz_plots/plots/pair_plot.py
```diff
@@ -508,3 +592,44 @@ def plot_pair(
...
+
+def _kde_couple(da_x, da_y, target, filled=False, levels=None, sample_dims=None, **kw):
+    if isinstance(levels, int):
+        n_levels = levels
+        hdi_probs = None
+    else:
+        hdi_probs = np.sort(levels)[::-1]
+        n_levels = None
+
+    result = kde2d(da_x, da_y, dim=sample_dims, hdi_probs=hdi_probs)
```
OriolAbril, line 605: We should open an issue about precomputing the 2dkde (at least for this plot only, maybe we can store the results in something with the same format as we do the `plot` variable). Otherwise it will be quite slow once the number of variables increases.

src/arviz_plots/plots/pair_plot.py
```diff
@@ -121,6 +133,7 @@ def plot_pair(
         Valid keys are:
 
         * scatter -> passed to :func:`~.visuals.scatter_couple`
+        * contour -> passed to :func:`~.visuals.contour`. Defaults to False.
```
OriolAbril, line 136: contourf is missing

OriolAbril (commented):
I think this is mostly good to go, most comments are small things. The main thing to fix are docs for visuals and backend functions: https://github.com/arviz-devs/arviz-plots/blob/main/docs/source/api/visuals.rst and https://github.com/arviz-devs/arviz-plots/blob/main/docs/source/api/backend/index.rst

PR #504, opened 2026-05-17
Forward labeller in plot_ridge labels

chrberrig:
Fixes `plot_ridge` not applying the `labeller` kwarg to ridge labels.

`plot_ridge` already accepts/constructs a labeller, and `annotate_label` already supports one, but the labeller was not forwarded in the label mapping call. This PR passes `labeller=labeller` and adds a test checking mapped labels across backends.

Validated with:
- `pytest tests/test_plots.py -k forwards_labeller`
- `tox -e check`
- manual notebook render

### DIFF
--- a/src/arviz_plots/plots/ridge_plot.py
+++ b/src/arviz_plots/plots/ridge_plot.py
@@ -503,6 +503,7 @@ def plot_ridge(
             subset_info=True,
             coords={"column": "labels"},
             ignore_aes=lab_ignore,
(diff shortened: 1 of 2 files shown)


aloctavodia:
thanks @chrberrig

chrberrig:
Thank you for your quick response :) 

I've just found that plot forest suffers from the same missing kwarg issue: the fix is virtually identical, and I just PR'd at #506

PR #509, opened 2026-06-02
plot_pair_focus: support extra dimensions

aloctavodia:
For instance, we may want to do 

```python
idata_0 = az.load_arviz_data("centered_eight")
idata_1 = az.load_arviz_data("non_centered_eight")

combined = xr.concat(
    [idata_0, idata_1],
    dim=xr.DataArray(["model_0", "model_1"], dims="__model__")
)

az.plot_pair_focus(combined, var_names=["mu"], focus_var="tau")
```
to get 

output

### DIFF
--- a/src/arviz_plots/plots/pair_focus_plot.py
+++ b/src/arviz_plots/plots/pair_focus_plot.py
@@ -150,7 +150,7 @@ def plot_pair_focus(
             f"focus_var should be a string or DataArray, got {type(focus_var)} instead."
         )
 
-    dims_y = [dim for dim in y.dims if dim not in sample_dims]
+    dims_y = [dim for dim in y.dims if dim not in distribution.dims]
     if len(dims_y) > 0:
         raise ValueError(f"focus variable has unexpected dimensions: {dims_y}.")
 


src/arviz_plots/plots/pair_focus_plot.py
```diff
@@ -150,7 +150,7 @@ def plot_pair_focus(
             f"focus_var should be a string or DataArray, got {type(focus_var)} instead."
         )
 
-    dims_y = [dim for dim in y.dims if dim not in sample_dims]
+    dims_y = [dim for dim in y.dims if dim not in distribution.dims]
     if len(dims_y) > 0:
         raise ValueError(f"focus variable has unexpected dimensions: {dims_y}.")
```
OriolAbril, line 155: I think the plot used as example in the description should error out. I am not completely sure the `dims_y` use as currently defined is completely right though, but I think this makes the plot not error out so we probably want this exact change.

Both models have a `tau` variable so it is not clear what should be used as focus var. For `plot_pair_focus` to work with the concatenated data in the example it should be necessary to use `focus_var_coords` to indicate which model to take `tau` from.

OriolAbril (changes requested):

src/arviz_plots/plots/pair_focus_plot.py
aloctavodia: I was expecting this to return a comparison per model, i.e. tau from model 0 against mu from model0, and then, tau from model 1 against mu from model 1

OriolAbril: This might be what is happening, otherwise matplotlib would error out due to tau and mu[model0] having different shapes. If we want the plot to handle this we'd only have the labeling issue for the y axes to look at, but I am not sure we want this to be the case given the general plot behaviour and defaults.

If we included more variables we'd probably end up with alternating tau0 va mu0, tau1 va mu1, tau0 vs theta0_choate, tau1 vs theta1_choate, ... I think it could be interesting to have a plot like the one you mention but it would be better if we have columns/rows mapped to the variable+other dimensions and the rows/columns mapped to model. I have been doing some improvements to combine_plots but this use case I think would still not be supported to combine two calls to plot_focus_var.

aloctavodia: Fine if you think we should close this.

aloctavodia:
Should we close this @OriolAbril?

OriolAbril:
probably best yes

PR #511, opened 2026-06-07
Bring arviz-darkgrid style back  

### LINKED ISSUE #510: Bring `az.style.use("arviz-darkgrid")` back
Hi team! This is a selfish request. I really loved the `az.style.use("arviz-darkgrid")` style! Any chance we can bring it back :) ?
Thanks!

juanitorduz:
Closes https://github.com/arviz-devs/arviz-plots/issues/510

I also added some unit-tests to verify all styles work with all backends :)

```python
import arviz_base
import arviz_plots as azp

azp.style.use("arviz-darkgrid")

idata = arviz_base.load_arviz_data("centered_eight")

pc = azp.plot_forest(idata, backend="matplotlib")
pc.viz["figure"].item();
```
image

```python
pc = azp.plot_forest(idata, backend="plotly")
pc.viz["figure"].item();
```

image

### DIFF
--- a/src/arviz_plots/backend/bokeh/core.py
+++ b/src/arviz_plots/backend/bokeh/core.py
@@ -130,8 +130,6 @@ def _set_sqrt_scale(target, axis):
 
 def get_hex_from_color_name(color_name: str) -> str:
     """Convert a standard CSS color name into its HEX code using Bokeh."""
-    if isinstance(color_name, str) and color_name.startswith("#"):
-        return color_name
     try:
         color_obj: Color = getattr(named_colors, color_name.lower())
         return color_obj.to_hex()
@@ -146,8 +144,9 @@ def get_background_color():
         from bokeh.io import curdoc
 
         bg_color = curdoc().theme._json["attrs"]["Plot"]["background_fill_color"]
-        hex_bg_color = get_hex_from_color_name(bg_color)
-        return hex_bg_color
+        if isinstance(bg_color, str) and bg_color.startswith("#"):
+            return bg_color
+        return get_hex_from_color_name(bg_color)
     except (ImportError, KeyError):
         return "#ffffff"
 
@@ -507,21 +506,49 @@ def hist(
         if edgecolor is unset:
             edgecolor = color
 
-    kwargs = {"bottom": bottom, "fill_color": facecolor, "line_color": edgecolor, "alpha": alpha}
+    kwargs = {
+        "bottom": bottom,
+        "fill_color": facecolor,
+        "line_color": edgecolor,
+        "alpha": alpha,
+    }
 
     return target.quad(top=y, left=l_e, right=r_e, **_filter_kwargs(kwargs, artist_kws))
 
 
 @expand_aesthetic_aliases
-def line(x, y, target, *, color=unset, alpha=unset, width=unset, linestyle=unset, **artist_kws):
+def line(
+    x,
+    y,
+    target,
+    *,
+    color=unset,
+    alpha=unset,
+    width=unset,
+    linestyle=unset,
+    **artist_kws,
+):
     """Interface to bokeh for a line plot."""
-    kwargs = {"color": color, "alpha": alpha, "line_width": width, "line_dash": linestyle}
+    kwargs = {
+        "color": color,
+        "alpha": alpha,
+        "line_width": width,
+        "line_dash": linestyle,
+    }
     return target.line(np.atleast_1d(x), np.atleast_1d(y), **_filter_kwargs(kwargs, artist_kws))
 
 
 @expand_aesthetic_aliases
 def multiple_lines(
-    x, y, target, *, color=unset, alpha=unset, width=unset, linestyle=unset, **artist_kws
+    x,
+    y,
+    target,
+    *,
+    color=unset,
+    alpha=unset,
+    width=unset,
+    linestyle=unset,
+    **artist_kws,
 ):
     """Interface to bokeh for multiple lines."""
     y = y.T
@@ -530,7 +557,12 @@ def multiple_lines(
     if len(x) != len(y):
         raise ValueError("x and y must have the same length")
     source = ColumnDataSource(data={"x": x, "y": y})
-    kwargs = {"line_color": color, "line_alpha": alpha, "line_width": width, "line_dash": linestyle}
+    kwargs = {
+        "line_color": color,
+        "line_alpha": alpha,
+        "line_width": width,
+        "line_dash": linestyle,
+    }
     return target.multi_line(xs="x", ys="y", source=source, **_filter_kwargs(kwargs, artist_kws))
 
 
@@ -646,7 +678,12 @@ def fill_between_y(x, y_bottom, y_top, target, **artist_kws):
 @expand_aesthetic_aliases
 def vline(x, target, *, color=unset, alpha=unset, width=unset, linestyle=unset, **artist_kws):
     """Interface to bokeh for a vertical line spanning the whole axes."""
-    kwargs = {"line_color": color, "line_alpha": alpha, "line_width": width, "line_dash": linestyle}
+    kwargs = {
+        "line_color": color,
+        "line_alpha": alpha,
+        "line_width": width,
+        "line_dash": linestyle,
+    }
     span_element = Span(location=x, dimension="height", **_filter_kwargs(kwargs, artist_kws))
     target.add_layout(span_element)
     return span_element
@@ -655,7 +692,12 @@ def vline(x, target, *, color=unset, alpha=unset, width=unset, linestyle=unset, 
 @expand_aesthetic_aliases
 def hline(y, target, *, color=unset, alpha=unset, width=unset, linestyle=unset, **artist_kws):
     """Interface to bokeh for a horizontal line spanning the whole axes."""
-    kwargs = {"line_color": color, "line_alpha": alpha, "line_width": width, "line_dash": linestyle}
+    kwargs = {
+        "line_color": color,
+        "line_alpha": alpha,
+        "line_width": width,
+        "line_dash": linestyle,
+    }
     span_element = Span(location=y, dimension="width", **_filter_kwargs(kwargs, artist_kws))
     target.add_layout(span_element)
     return span_element
@@ -693,7 +735,12 @@ def ciliney(
     **artist_kws,
 ):
     """Interface to bokeh for a line from y_bottom to y_top at given value of x."""
-    kwargs = {"color": color, "alpha": alpha, "line_width": width, "line_dash": linestyle}
+    kwargs = {
+        "color": color,
+        "alpha": alpha,
+        "line_width": width,
+        "line_dash": linestyle,
+    }
     x = np.atleast_1d(x)
     y_bottom = np.atleast_1d(y_bottom)
     y_top = np.atleast_1d(y_top)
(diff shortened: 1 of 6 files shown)


src/arviz_plots/backend/bokeh/core.py
```diff
@@ -130,6 +130,8 @@ def _set_sqrt_scale(target, axis):
 
 def get_hex_from_color_name(color_name: str) -> str:
     """Convert a standard CSS color name into its HEX code using Bokeh."""
+    if isinstance(color_name, str) and color_name.startswith("#"):
```
aloctavodia: Why is this needed now?

src/arviz_plots/backend/plotly/templates.py
```diff
@@ -92,6 +92,37 @@
...
+arviz_darkgrid_template.data.bar = h_template
+arviz_darkgrid_template.layout.paper_bgcolor = "white"
+arviz_darkgrid_template.layout.plot_bgcolor = "#eeeeee"
+arviz_darkgrid_template.layout.polar.bgcolor = "#eeeeee"
+arviz_darkgrid_template.layout.ternary.bgcolor = "#eeeeee"
+arviz_darkgrid_template.layout.margin = {"l": 50, "r": 10, "t": 40, "b": 45}
+arviz_darkgrid_template.layout.xaxis = axis_darkgrid
+arviz_darkgrid_template.layout.yaxis = axis_darkgrid
+arviz_darkgrid_template.layout.colorway = [
+    "#2a2eec",
```
aloctavodia, line 113: We now use 8 colors, following recommendations in https://clauswilke.com/dataviz/

src/arviz_plots/backend/bokeh/core.py
juanitorduz: because the `bokeh` backend was failing otherwise

src/arviz_plots/backend/plotly/templates.py
juanitorduz: removed the last 2 https://github.com/arviz-devs/arviz-plots/pull/511/commits/d574ab22721834058c6647b46bba3015b33f7953

src/arviz_plots/backend/bokeh/core.py
juanitorduz: this is the error:

```python
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
File ~/Documents/arviz-plots/src/arviz_plots/backend/bokeh/core.py:136, in get_hex_from_color_name(color_name)
    135 try:
--> 136     color_obj: Color = getattr(named_colors, color_name.lower())
    137     return color_obj.to_hex()

AttributeError: module 'bokeh.colors.named' has no attribute '#eeeeee'

The above exception was the direct cause of the following exception:

ValueError                                Traceback (most recent call last)
Cell In[2], line 1
----> 1 pc = azp.plot_forest(idata, backend="bokeh")
      2 pc.viz["figure"].item()

File ~/Documents/arviz-plots/src/arviz_plots/plots/forest_plot.py:472, in plot_forest(dt, var_names, filter_vars, group, coords, sample_dims, combined, point_estimate, ci_kind, ci_probs, labels, shade_label, plot_collection, backend, labeller, aes_by_visuals, visuals, stats, **pc_kwargs)
    470     if x == len(labels) - 1:
    471         lab_kwargs.setdefault("horizontal_align", "right")
--> 472     plot_collection.map(
    473         annotate_label,
    474         f"{label.strip('_')}_label",
    475         data=y,
    476         x=x,
    477         dim=None if label == "__variable__" else label,
    478         subset_info=True,
    479         coords={"column": "labels"},
    480         ignore_aes=lab_ignore,
    481         labeller=labeller,
    482         **lab_kwargs,
    483     )
    484     x += 1
    485 ticklabel_kwargs = get_visual_kwargs(visuals, "ticklabels")

File ~/Documents/arviz-plots/src/arviz_plots/plot_collection.py:1329, in PlotCollection.map(self, func, func_label, data, coords, ignore_aes, subset_info, store_artist, artist_dims, **kwargs)
   1326 if subset_info:
   1327     func_kwargs = {**func_kwargs, "var_name": var_name, "sel": sel, "isel": isel}
-> 1329 aux_artist = func(da, target=target, **func_kwargs)
   1330 if store_artist:
   1331     if np.size(aux_artist) == 1:

File ~/Documents/arviz-plots/src/arviz_plots/visuals/__init__.py:405, in annotate_label(da, target, var_name, sel, isel, x, y, dim, labeller, **kwargs)
    403     text = labeller.sel_to_str(sel, isel)
    404 plot_backend = backend_from_object(target)
--> 405 return plot_backend.text(
    406     x,
    407     y,
    408     text,
    409     target,
    410     **kwargs,
    411 )

File ~/Documents/arviz-plots/src/arviz_plots/backend/alias_utils.py:83, in create_aesthetic_handlers.<locals>.expand_aesthetic_aliases.<locals>._dealiased_plot_fn(*args, **kwargs)
     77 for aes in ("color", "facecolor", "edgecolor"):
     78     if (
     79         aes in kwargs
     80         and isinstance(value := kwargs[aes], str)
     81         and (cycle_pattern.match(value) or background_pattern.match(value))
     82     ):
---> 83         kwargs[aes] = dealiase_aes_value(aes, value)
     84 for aes in ("linestyle", "marker"):
     85     if (
     86         aes in kwargs
     87         and isinstance(value := kwargs[aes], str)
     88         and cycle_pattern.match(value)
     89     ):

File ~/Documents/arviz-plots/src/arviz_plots/backend/alias_utils.py:61, in create_aesthetic_handlers.<locals>.dealiase_aes_value(aes, value)
     59     dealiased_value = get_default_aes(aes, index + 1)[index]
     60 else:
---> 61     bg_color = get_background_color()
     62     match index:
     63         case 0:

File ~/Documents/arviz-plots/src/arviz_plots/backend/bokeh/core.py:149, in get_background_color()
    146     from bokeh.io import curdoc
    148     bg_color = curdoc().theme._json["attrs"]["Plot"]["background_fill_color"]
--> 149     hex_bg_color = get_hex_from_color_name(bg_color)
    150     return hex_bg_color
    151 except (ImportError, KeyError):

File ~/Documents/arviz-plots/src/arviz_plots/backend/bokeh/core.py:140, in get_hex_from_color_name(color_name)
    137     return color_obj.to_hex()
    139 except AttributeError as exc:
--> 140     raise ValueError(f"Color '{color_name}' is not a valid Bokeh named color.") from exc

ValueError: Color '#eeeeee' is not a valid Bokeh named color.
```

aloctavodia: Should we fix it at this level or here 

bg_color = curdoc().theme._json["attrs"]["Plot"]["background_fill_color"]
hex_bg_color = get_hex_from_color_name(bg_color..strip("#"))

juanitorduz: like https://github.com/arviz-devs/arviz-plots/pull/511/changes/f8a09622a3c92778e870325868eb60b2d3de0683 ? (up the re-formatting, I reverted that in the next commit)

aloctavodia: No, I mean if we should fix it here File ~/Documents/arviz-plots/src/arviz_plots/backend/bokeh/core.py:149,

juanitorduz: something like https://github.com/arviz-devs/arviz-plots/pull/511/commits/734e81a2c9bebe9040baca16bcd77a905a629e32 ?

tests/test_styles.py
```diff
@@ -0,0 +1,79 @@
...
+
+STYLES_DIR = Path(azp.__file__).parent / "styles"
+# Source of truth: every style we ship a matplotlib ``.mplstyle`` for.
+# The list auto-updates when a style is added/removed.
+ARVIZ_STYLES = sorted(path.stem for path in STYLES_DIR.glob("*.mplstyle"))
+BACKENDS = ["matplotlib", "bokeh", "plotly"]
+
+
+@pytest.fixture
+def restore_style():
```
OriolAbril: I was about to propose switching to `addfinalizer` instead of yield to ensure teardown still runs if setup runs into an error but then realized we don't actually need any setup in this fixture. All 3 libraries already have internal objects that represent/revert back to defaults.

To revert back in matplotlib we can call `matplotlib.rcdefaults()`, for plotly we can set `plotly.io.templates.default = "plotly"` and for bokeh `curdoc().theme = bokeh.themes.default` (even if we import and call all of these after setting up a different theme to be used. I would also replace the try excepts for checking of the `backend` parametrization like we do for example in `check_skips`. Something like:

```python
@pytest.fixture(scope="function")
def restore_style(request):
    yield
    if ("backend" in request.fixturenames):
        if any("matplotlib" in key for key in request.keywords.keys()):
            import matplotlib
            matplotlib.rcdefaults()
        if any("bokeh" in ...)
```

tests/test_styles.py
```diff
@@ -0,0 +1,79 @@
...
+            assert style_name in available[backend], f"{style_name} missing for {backend}"
+    if "common" in available:
+        assert style_name in available["common"]
+
+
+@pytest.mark.parametrize("backend", BACKENDS)
+@pytest.mark.parametrize("style_name", ARVIZ_STYLES)
+@pytest.mark.usefixtures("restore_style", "clean_plots", "check_skips")
+def test_style_renders(datatree, style_name, backend):
+    pytest.importorskip(backend)
```
OriolAbril:
```suggestion
```

We have been using only `check_skips` to handle running of tests when one/some of the backends are not installed. If one of the backends is not installed, the `--skip-bkn` flag should be used instead of tests skipping automatically. It adds a bit of friction to running tests with missing optional dependencies which is mostly what we want.

tests/test_styles.py
juanitorduz: Thanks @OriolAbril , in https://github.com/arviz-devs/arviz-plots/pull/511/changes/01565e0d2fed9a35413892a5f5137e960d8df58f I tried to address this suggestion. 

**One comment:** I was talking to Claude and it brought the following remark: 
*Adopted the no-setup + reset-to-defaults approach 👍. One nuance: style.use() sets the default on all three backends every call, so we reset all three in teardown rather than gating by the backend parametrization, otherwise the other two backends' state leaks out of the module. Since the matrix always installs all three backends, the imports are unconditional, which also drops the try/excepts you flagged.*

OriolAbril: Good catch. I usually run tests only through the tox tasks which do install anything and it is the recommended way to do so but we should still take `--skip-xyz` into account. When one of these flags is given all tests for that backend are skipped (because we want things to go faster and haven't been working on that backend or because we are running tests somewhere else where they are not installed). So I would only run the style resetting if we are actually running tests for that backend.

```python
skip_mpl = request.config.getoption("--skip-mpl")
if not skip_mpl:
    import matplotlib
    matplotlib.rcdefaults()
...
```

If matplotlib is installed and we are skipping its tests the defaults will be modified for the pytest session but there are not matplotlib tests being run so I think it doesn't matter and if it isn't installed we can continue to run tests without worrying as long as we use the flag.

juanitorduz: so like https://github.com/arviz-devs/arviz-plots/pull/511/changes/43588b475fd35a3845911fba5b05e97e65cf4dc9 ?

OriolAbril (approved):

juanitorduz:
Thanks @aloctavodia @OriolAbril 🫶

PR #512, opened 2026-06-07
Add tox check job to ci

juanitorduz:
This can help get lint errors faster :)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,6 +7,13 @@ on:
       - "docs/"
 
 jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-python@v3
+    - uses: pre-commit/action@v3.0.1
+
   test:
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
OriolAbril, line 39: as we are running tox with no arguments all jobs already run. If you check for example the logs in the [3.13 python](https://github.com/arviz-devs/arviz-plots/actions/runs/27103664376/job/79988805214?pr=512) there are 3 tox tasks being run:

```
  check: OK (35.12=setup[3.49]+cmd[0.20,29.43,2.01] seconds)
  nightlies: OK (394.95=setup[46.67]+cmd[348.27] seconds)
  py313-coverage: OK (710.89=setup[30.59]+cmd[680.30] seconds)
```

This is controlled by this section in the tox.init: https://github.com/arviz-devs/arviz-plots/blob/main/tox.ini#L13. Not completely opposed to moving it to its own job/workflow, in arviz-stats the testing environments are much more complex so it was already separated but I am not sure it actually adds anything in this context

juanitorduz: Fair enough :) I suggest then to close this one to keep things simpler (I'm not super familiar with tox) :)

PR #514, opened 2026-06-11
Allow chaging the label

aloctavodia:
I found this trying to do 

```python
pc = azp.plot_loo_interval(
    dt,
    visuals={"ylabel": {"text": "my label"}},
)
```
But I think it makes sense to change it for the other too.

### DIFF
--- a/src/arviz_plots/plots/lm_plot.py
+++ b/src/arviz_plots/plots/lm_plot.py
@@ -525,10 +525,11 @@ def plot_lm(
     xlabel_kwargs = get_visual_kwargs(visuals, "xlabel")
     if xlabel_kwargs is not False:
         _, _, xlabel_ignore = filter_aes(plot_collection, aes_by_visuals, "xlabel", sample_dims)
+        if "text" not in xlabel_kwargs:
+            xlabel_kwargs["labeller"] = xlabeller
         plot_collection.facet_map(
             labelled_x,
             "xlabel",
-            labeller=xlabeller,
             subset_info=True,
             ignore_aes=xlabel_ignore,
             **xlabel_kwargs,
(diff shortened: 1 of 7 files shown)


OriolAbril (commented):
I think all of these should be handled by the `labeller`. Particularly for ppc related plots it is very common to generate only one subplot in the figure so setting the label name directly is tempting and would work, but then it signals `text` is the way to change the label which won't work for multiple variables, or single variables if we customize the facetting...

I believe `labeller=MapLabeller(var_name_map={"var name": "my label"})` will fix most of what you are trying to achieve, if you want something else maybe we should add a new labeller? Not sure we really want it but a way to replicate the `text` argument completely could be `FixedLabeller("my label")` that always returns the same string independently of the subset info provided to the labeller.

Potentially related, maybe we should use non-default labellers in the example gallery docs?

aloctavodia:
To be honest, I was not sure about what to propose here, haha. But I feel confused because 

```python
# This works
pc = azp.plot_loo_interval(
    data_subset,
    visuals={"xlabel": {"text": "lala"}},   
)
```
```python
# This does not work
pc = azp.plot_loo_interval(
    data_subset,
    visuals={"ylabel": {"text": "lala"}},   
)
```
My first reaction when this failed was to check if this works 

```python
pc = azp.plot_loo_interval(data_subset)
pc.facet_map("labelled_y", text="lala")
```

Which is similar to using map but feels more ¿high-level?

```python
pc = azp.plot_loo_interval(data_subset)
pc.map(az.visuals.labelled_y", text="lala")
```

But maybe I should have used a labeler as you propose.
 
```python
labeller = az.labels.MapLabeller({"y": "lala"})
pc = azp.plot_loo_interval(
    data_subset,
    labeller=labeller,
)
```

Fine to close if you think this is better addressed at the documentation level.

aloctavodia:
Should we close @OriolAbril?

OriolAbril:
sounds good

PR #519, opened 2026-06-13
Replace isort, black and pydocstyle with ruff and add missing pre-commit hooks

### LINKED ISSUE #469: pre-commit: missing hooks and outdated pre-commit-hooks version in arviz-plots
on comparing .pre-commit-config.yaml across repos, arviz-plots is missing several hooks that both arviz-base and arviz-stats have:

- check-toml
- end-of-file-fixer
- no-commit-to-branch
- trailing-whitespace

also the pre-commit-hooks version is pinned to v4.3.0 while arviz-base and arviz-stats both use v5.0.0.
these should be added and updated to bring arviz-plots in line with the rest of the ecosystem.

PiyushPanwarFST:
Closes #469

  - Replace isort, black and pydocstyle with ruff + ruff-format
  - Add missing hooks: check-toml, end-of-file-fixer, no-commit-to-branch, trailing-whitespace
  - Remove old [tool.black], [tool.isort], [tool.pydocstyle] from pyproject.toml and add [tool.ruff]
  - Keep ruff rules minimal (I + D) and leave the pylint hook unchanged
  - Second commit is the automatic formatting fixes applied by the new hooks


.pre-commit-config.yaml
```diff
@@ -4,26 +4,21 @@ repos:
...
   - id: check-merge-conflict
+  - id: end-of-file-fixer
+  - id: no-commit-to-branch
+    args: [--branch, main]
+  - id: trailing-whitespace
 
-- repo: https://github.com/PyCQA/isort
-  rev: 5.12.0
+- repo: https://github.com/astral-sh/ruff-pre-commit
+  rev: v0.11.13
```
OriolAbril: this is not the most recent version. Given the plan is to update arviz-base and arviz-stats, there is no point in not using the version we are aiming for here from the start

.pre-commit-config.yaml
```diff
@@ -4,26 +4,21 @@ repos:
...
-  - id: black
-    exclude: ^docs/source/gallery/
-
-- repo: https://github.com/pycqa/pydocstyle
-  rev: 6.3.0
-  hooks:
-    - id: pydocstyle
-      additional_dependencies: [tomli]
-      files: ^src/arviz_plots/.+\.py$
+    - id: ruff
```
OriolAbril: This should be `ruff-check` in recent versions, using `ruff` alone is deprecated

.pre-commit-config.yaml
```diff
@@ -4,26 +4,21 @@ repos:
...
+  rev: v0.11.13
   hooks:
-  - id: isort
-    exclude: ^src/arviz_base/example_data/
-
-- repo: https://github.com/psf/black
-  rev: 24.10.0
-  hooks:
-  - id: black
-    exclude: ^docs/source/gallery/
```
OriolAbril, line 19: black was being ignored for all files in this folder but ruff is not being ignored with these settings

.pre-commit-config.yaml
```diff
@@ -4,26 +4,21 @@ repos:
...
-
-- repo: https://github.com/pycqa/pydocstyle
-  rev: 6.3.0
-  hooks:
-    - id: pydocstyle
-      additional_dependencies: [tomli]
-      files: ^src/arviz_plots/.+\.py$
```
OriolAbril, line 26: This one I see is being mimicked through per file ignores in `pyproject`

pyproject.toml
```diff
@@ -64,22 +64,25 @@ doc = [
...
-include_trailing_comma = true
-use_parentheses = true
-multi_line_output = 3
-line_length = 100
-skip = [
-    "src/arviz_plots/__init__.py"
+[tool.ruff.lint]
+select = [
+    "I",  # isort
+    "D",  # pydocstyle
```
OriolAbril: we can skip the `PL` rules given we'll keep pylint, but we should incorporate the other non pylint rules we have in arviz-base and arviz-stats. I haven't checked how much overlap there is between rules like `F` or `E` with things pylint is already checking, but numpy or pyupgrade rules are complementary to pylint and should be taken into account.

Also if comparing to arviz-base there is the "TID" rule, in arviz-stats the absolutify-imports pre-commit check. One of these two should also be included in the settings for arviz-plots now that we are updating it.

tests/test_hypothesis_plots.py
```diff
@@ -258,10 +259,24 @@ def test_combine_plots(datatree, expand, random_plots):
     diagnostics=st.sampled_from(
         [
             # fmt: off
```
OriolAbril: This turning off the format is not being taken into account now

OriolAbril (commented):
This is a good start but still needs a bit of work to maintain existing settings and bring up to par with arviz-base and arviz-stats

.pre-commit-config.yaml
PiyushPanwarFST: Yaa make sense, I will look at the latest version of `ruff-pre-commit` and replace the old one with it.

.pre-commit-config.yaml
PiyushPanwarFST: Thanks for pointing out, yaa I missed the `exclude: ^docs/source/gallery/`, I will fix it.

.pre-commit-config.yaml
PiyushPanwarFST: Yes exactly, I moved the `pydocstyle` scope into `per-file-ignores` so D only runs on src.

.pre-commit-config.yaml
PiyushPanwarFST: ok, I will rename it from `ruff` to `ruff-check`.

pyproject.toml
PiyushPanwarFST: Agreed, I kept it too minimal but I will add the non-pylint rules from base/stats (F, E, W, NPY, UP) and skip PL since pylint stays. 

And for the relative imports I looked into the TID rule to check if ruff can replace absolufy-imports. 

So I looked into ruff docs and as per my understanding, by default ruff only warns on relative imports and doesn't auto-convert them (the fix is marked "unsafe"). But adding one line in `pyproject.toml` file `extend-safe-fixes = ["TID252"] `, turns the autofix ON and it works same as absolufy.

pyproject.toml
```diff
@@ -64,22 +64,36 @@ doc = [
...
+    "I",  # isort
+    "TID",  # flake8-tidy-imports (relative imports)
 ]
+extend-safe-fixes = ["TID252"]
 
-[tool.pydocstyle]
+[tool.ruff.lint.per-file-ignores]
+"src/arviz_plots/**/__init__.py" = ["I", "F401", "F403", "E402"]
+"src/arviz_plots/**/*.pyi" = ["TID252"]
+"tests/**/*" = ["D", "NPY002", "TID252"]
```
OriolAbril:
```suggestion
"tests/**/*" = ["D"]
```

I ran `tox -e check` from this PR after removing these two checks and had only two ruff errors, both related to `NPY002` rule. The reason why `NPY002` is ignored in arviz-stats is because we haven't had time to make the required test updates, two single occurrences is not the case. Moreover, the project is both about updating the infrastructure and making sure the codebase is aligned with the improved linting and checks.

`TID252` is not being violated within the tests folders, so there is no need to add it at all.

.pre-commit-config.yaml
```diff
@@ -1,29 +1,25 @@
...
+    - id: ruff-check
+      args: [--fix, --exit-non-zero-on-fix]
+      types: [ python ]
+    - id: ruff-format
+      types: [ python ]
+      exclude: ^docs/source/gallery/
```
OriolAbril, line 22: now the folder is being excluded but all the files have already been updated. In this particular case it means the ruff formatter added a newline between the "script docstring" and the code which is irrelevant, adds nothing but is also not harmful and also added some trailing commas in function calls so no need to undo any of the changes.

But if you have to do other formatter related more comples tasks during the summer you'll need to first commit the settings, then run `tox -e check` and see what happens. Then, in order to update the settings, undo the latest commit, add the updated settings and run `tox -e check` from the starting point instead of on top of the badly configured formatter output.

.pre-commit-config.yaml
```diff
@@ -1,29 +1,25 @@
...
-  rev: v4.3.0
+  rev: v6.0.0
   hooks:
   - id: check-added-large-files
     args: ['--maxkb=1500']
+  - id: check-toml
   - id: check-merge-conflict
+  - id: end-of-file-fixer
+  - id: no-commit-to-branch
+    args: [--branch, main]
```
OriolAbril, line 11: This will fail if added as is when running CI on main. You should update the workflow to set the `SKIP` env variable when running this on GitHub actions. See arviz-base for reference: https://github.com/arviz-devs/arviz-base/blob/main/.github/workflows/test.yml#L27

tests/test_hypothesis_plots.py
```diff
@@ -257,12 +258,24 @@ def test_combine_plots(datatree, expand, random_plots):
     ),
     diagnostics=st.sampled_from(
         [
-            # fmt: off
```
OriolAbril, line 260: This comment should be updated so the formatting continues to be disabled. I am sure ruff also supports this the only thing is it probably works a bit differently than black does.

tests/test_plots.py
```diff
@@ -332,9 +333,7 @@ def test_plot_ess_evolution(self, datatree, backend):
         assert all("hierarchy" not in child["mu"].dims for child in pc.viz.children.values())
         assert all("hierarchy" in child["theta"].dims for child in pc.viz.children.values())
 
-    def test_plot_ess_evolution_sample(
-        self, datatree_sample, backend
-    ):  # pylint: disable=unused-argument
+    def test_plot_ess_evolution_sample(self, datatree_sample, backend):  # pylint: disable=unused-argument
         pc = plot_ess_evolution(datatree_sample, sample_dims="sample")
```
OriolAbril:
```suggestion
    def test_plot_ess_evolution_sample(self, datatree_sample, backend):
pc = plot_ess_evolution(datatree_sample, sample_dims="sample", backend=backend)
```

We have a file-level ignore of the `no-self-use` so this ignore was incorrect as it was correctly pointing out the `backend` argument was not being used

pyproject.toml
PiyushPanwarFST: Yaa makes sense, only 2 cases and the file already uses default_rng just below, so I will fix those instead of ignoring. Also dropping TID252 since tests have no relative imports so it becomes just ["D"].

.pre-commit-config.yaml
PiyushPanwarFST: Ah I see, it will fail when CI runs on main. I will update the workflow to set the SKIP env variable for that hook. [run: SKIP=no-commit-to-branch tox]

tests/test_hypothesis_plots.py
PiyushPanwarFST: Understood I will keep the formatting disabled and use ruff's way to do it instead of removing it. Will update.

tests/test_plots.py
PiyushPanwarFST: Thanks, you are right the inline ignore was wrong since the file already has no-self-use. I will apply the suggestion.

OriolAbril (approved):

PR #526, opened 2026-06-30
Use a single tox -e test env instead of per-version envs

PiyushPanwarFST:
- replace py312/py313/py314 test envs with one version-independent test env
- keep the gh-actions and basepython sections, just point them at test
- CI still runs plain tox, so check and nightlies keep working
- also drop the stale py311 leftover in the dev label and basepython
- follows the arviz metapackage template


OriolAbril (commented):
The respective docs also need to be updated as part of this same PR: https://python.arviz.org/projects/plots/en/latest/contributing/testing.html

OriolAbril (approved):

PR #529, opened 2026-07-05
two small pre-commit setup changes:

PiyushPanwarFST:
- added the no-print-statements hook so plots matches arviz-base & arviz-stats (no print() in the code, so it passes as is)
- added pre-commit to dependabot so our hook versions update automatically, like github-actions already does

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,7 @@ updates:
   directory: "/"
   schedule:
     interval: "weekly"
+- package-ecosystem: "pre-commit"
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
OriolAbril, line 1: This config is missing the updated interval and cooldown for both github-actions and pre-commit

OriolAbril (commented):
also update and freeze the pre-commit hooks here

OriolAbril (approved):

PR #533, opened 2026-07-11
Add ROPE to plot_dist

### LINKED ISSUE #518: Feature request: built-in ROPE support in plot_dist
## Is your feature request related to a problem?

In ArviZ ≤0.x, `plot_posterior` accepted `rope=` and `ref_val=` directly. In 1.x, replicating this with `plot_dist` requires manually composing three separate helpers:

​```python
pc = az.plot_dist(idata, var_names=["flipper_length_mm"], kind="kde",
                  ci_kind="hdi", ci_prob=0.94)
az.add_bands(pc, [(LO, HI)])
az.add_lines(pc, REF)
pct = float(az.ci_in_rope(idata, var_names=["flipper_length_mm"], rope=(45, 51))["flipper_length_mm"])
​```

This works, but the ROPE bounds and reference value are repeated across calls (`add_bands` vs `ci_in_rope`), the band/line styling isn't tied to the distribution's aesthetics, and the in-ROPE percentage has to be extracted and annotated by hand. The prior implementation also provided richer information by showing the percent inside and outside the rope.

## Describe the solution you'd like

First-class `rope` and `ref_val` arguments on `plot_dist`, e.g.:

​```python
az.plot_dist(idata, var_names=["flipper_length_mm"], rope=(LO, HI), ref_val=REF)
​```
Per-variable ROPEs via a dict (as legacy `plot_posterior` allowed) would also be helpful.

## Describe alternatives you've considered

The three-call composition above, plus dropping `add_bands`/`add_lines` entirely and using raw matplotlib (`ax.axvspan`, `ax.axvline`) on `pc.get_viz("plot", var)` for full styling control. Both work but neither restores the convenience/parity of the old single argument.

## Additional context

Versions: `arviz 1.1.0`, `arviz-plots 1.1.0`, `bambi 0.18.0`, Python 3.14. Installed via pixi.

aloctavodia:
closes #518 

```python

az.plot_dist(dt, var_names=["mu", "tau"], 
             rope={"mu": (-0.5, 0.91), "tau": (0.2, 9.3)})
```
output

### DIFF
--- a/src/arviz_plots/plots/dist_plot.py
+++ b/src/arviz_plots/plots/dist_plot.py
@@ -6,13 +6,14 @@ from typing import Any, Literal
 
 import arviz_stats
 import xarray as xr
-from arviz_base import rcParams
+from arviz_base import rcParams, references_to_dataset
 from arviz_base.validate import (
     validate_ci_prob,
     validate_dict_argument,
     validate_or_use_rcparam,
     validate_sample_dims,
 )
+from arviz_stats import ci_in_rope
 
 from arviz_plots.plot_collection import PlotCollection
 from arviz_plots.plots.utils import (
@@ -25,6 +26,7 @@ from arviz_plots.plots.utils import (
     set_wrap_layout,
 )
 from arviz_plots.visuals import (
+    annotate_xy,
     ecdf_line,
     fill_between_y,
     hist,
@@ -36,6 +38,7 @@ from arviz_plots.visuals import (
     scatter_x,
     scatter_xy,
     step_hist,
+    vspan,
 )
 
 
@@ -49,6 +52,7 @@ def plot_dist(
     sample_dims=None,
     kind=None,
     point_estimate=None,
+    rope=None,
     ci_kind=None,
     ci_prob=None,
     plot_collection=None,
@@ -61,6 +65,8 @@ def plot_dist(
             "credible_interval",
             "point_estimate",
             "point_estimate_text",
+            "rope",
+            "rope_text",
             "title",
             "rug",
         ],
@@ -73,14 +79,17 @@ def plot_dist(
             "credible_interval",
             "point_estimate",
             "point_estimate_text",
-            "title",
+            "rope",
+            "rope_text",
             "rug",
             "remove_axis",
+            "title",
         ],
         Mapping[str, Any] | bool,
     ] = None,
     stats: Mapping[
-        Literal["dist", "credible_interval", "point_estimate"], Mapping[str, Any] | xr.Dataset
+        Literal["dist", "credible_interval", "point_estimate", "rope"],
+        Mapping[str, Any] | xr.Dataset,
     ] = None,
     **pc_kwargs,
 ):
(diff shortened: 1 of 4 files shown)


src/arviz_plots/plots/dist_plot.py
```diff
@@ -158,6 +172,8 @@ def plot_dist(
         * point_estimate_text -> passed to :func:`~arviz_plots.visuals.point_estimate_text`
         * title -> passed to :func:`~arviz_plots.visuals.labelled_title`
         * rug -> passed to :func:`~arviz_plots.visuals.scatter_x`. Defaults to False.
+        * rope -> passed to :func:`~arviz_plots.visuals.vspan`. Defaults to True.
+        * rope_text -> passed to :func:`~arviz_plots.visuals.annotate_xy`. Defaults to True.
```
OriolAbril, line 176: not sure if we need to specify they default to true but are ignored if the `rope`argument  is `None`

src/arviz_plots/plots/dist_plot.py
```diff
@@ -517,6 +555,60 @@ def plot_dist(
...
+        # rope band
+        rope_kwargs = get_visual_kwargs(visuals, "rope", True)
+        if rope_kwargs is not False:
+            _, rope_aes, rope_ignore = filter_aes(
+                plot_collection, aes_by_visuals, "rope", sample_dims
+            )
+            if "color" not in rope_aes:
+                rope_kwargs.setdefault("color", "B1")
+            if "alpha" not in rope_aes:
+                rope_kwargs.setdefault("alpha", 0.2)
```
OriolAbril, line 570: not sure how the plot will look like with the other backends. For the color I would use `B3` for the shaded region which is the smaller contrast alternative, or `B2` if we want high transparency. Regarding transparency and placing, I am not sure all backends support `zorder` functionality, it looks like we want the vspan to be at the bottom (which means we might need to make it the first element we draw) and the text at the top (so last element to draw).

src/arviz_plots/plots/dist_plot.py
```diff
@@ -517,6 +555,60 @@ def plot_dist(
...
+                density_ys = density.sel(
+                    plot_axis=[
+                        coord
+                        for coord in density["plot_axis"].values
+                        if coord in ("y", "histogram")
+                    ]
+                )
+                rope_y = 0.35 * density_ys.max(
+                    [dim for dim in density_ys.dims if dim not in rope_mid.dims]
+                )
```
OriolAbril, line 595: I think we can compute something like `plot_y_max` once and then use that, multiplied by different coefs, for the y values of the point estimate text and rope elements

tests/test_plots.py
```diff
@@ -245,6 +245,39 @@ def test_plot_dist_groupby(self, datatree, backend, kind):
             set(pc.viz[visual]["theta"]["hierarchy"].values) == set(coords) for visual in visuals
         )
 
+    @pytest.mark.parametrize("kind", ["kde", "hist", "ecdf"])
```
OriolAbril: hopefully won't make much of a difference but we should also add `dot` kind to tests

src/arviz_plots/plots/dist_plot.py
```diff
@@ -112,6 +121,11 @@ def plot_dist(
         Defaults to ``rcParams["plot.density_kind"]``
     point_estimate : {"mean", "median", "mode"}, optional
         Which point estimate to plot. Defaults to rcParam :data:`stats.point_estimate`
+    rope : tuple of float, or dict of {str : tuple of float}, optional
+        If provided, a region of practical equivalence (ROPE) is plotted as a shaded area.
+        The annotation shows the % of the credible interval that falls within the ROPE.
+        For the % of the full posterior that falls within the ROPE pass
+        ``stats={"rope": {"ci_prob": 1}},`
```
OriolAbril:
```suggestion
        ``stats={"rope": {"ci_prob": 1}},``
```

src/arviz_plots/plots/dist_plot.py
aloctavodia: This looks good with the other backends. Moving to the top (and removing the transparency) does not work with Plotly. Although there is an issue with the annotation. It gets clipped. I was able to fix it for plotly, but surprise, surprise, not sure how to fix it for bokeh.

src/arviz_plots/plots/dist_plot.py
aloctavodia: I opened an issue

PR #536, opened 2026-07-13
Set default style DPI to 100

### LINKED ISSUE #532: Set default dpi to 100
Currently, we use DPI 200 (maybe arviz-darkgrid is an exception). We should use DPI 100 (matplotlib's default)

Silvren:
## Summary

- Set figure.dpi to 100.0 for all bundled Matplotlib styles that previously used 200.0.
- Add a regression test verifying every bundled style uses the default DPI.

Closes #532

## Testing

- MPLBACKEND=Agg python -m pytest -q — 761 passed, 4 skipped
- ruff check tests/test_styles.py
- ruff format --check tests/test_styles.py

### DIFF
--- a/src/arviz_plots/styles/arviz-cetrino.mplstyle
+++ b/src/arviz_plots/styles/arviz-cetrino.mplstyle
@@ -8,7 +8,7 @@ figure.titleweight: bold    # weight of the figure title
 figure.titlesize:   x-large
 
 figure.figsize:     6, 5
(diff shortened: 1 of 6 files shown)


aloctavodia:
I don't think we need a test for this.

Silvren:
Thanks for the feedback. I've removed the test and kept the PR focused on updating the bundled style DPI values.

PR #537, opened 2026-07-13
fix: flatten 2D arrays in scatter for plotly and bokeh backends

### LINKED ISSUE #535: Rug not properly displayed for plotly and bokeh backends
```python
pc = az.plot_dist(dt, var_names=["mu", "tau"], 
                  backend="...",
                  visuals={"rug":True},
)
```
backend="matplotlib"

Image

backend="plotly"

Image

backend="bokeh"

Image

andrianbalanesq:
## Summary

When plotting rugs (or any `scatter_x` visual) with multi-dimensional data (e.g., posterior samples with chain and draw dimensions), the `x` and `y` arrays passed to plotly's `go.Scatter` and bokeh's `ColumnDataSource` were 2D. Matplotlib's `scatter` handles 2D arrays by flattening them internally, but plotly and bokeh expect 1D arrays.

This caused rugs to not display in plotly and bokeh backends.

## Root Cause

The `scatter_x` visual function calls `plot_backend.scatter(da, y, target, **kwargs)` where `da` is the distribution data (shape: chain x draw) and `y = np.zeros_like(da)` (same shape). The matplotlib backend flattens 2D arrays automatically, but:

- **Plotly**: `go.Scatter(x=np.atleast_1d(x), ...)` -- `np.atleast_1d` does not flatten 2D arrays, so `x` and `y` remained 2D (e.g., shape 4x500)
- **Bokeh**: `ColumnDataSource(data={"x": np.atleast_1d(x), ...})` -- same issue, bokeh expects 1D arrays

## Fix

Use `np.ravel()` to flatten the arrays before passing them to plotly and bokeh scatter functions.

## Verification

- Verified that rug markers now have correct 1D shape (2000,) instead of 2D (4, 500) for both plotly and bokeh
- Matplotlib backend continues to work unchanged
- Existing tests pass (the only error is a pre-existing numpy/netCDF4 binary incompatibility unrelated to this change)

Fixes #535

### DIFF
--- a/src/arviz_plots/backend/plotly/core.py
+++ b/src/arviz_plots/backend/plotly/core.py
@@ -600,8 +600,8 @@ def scatter(
     if marker == "|":
         scatter_kwargs["symbol"] = "line-ns-open"
     scatter_object = go.Scatter(
-        x=np.atleast_1d(x),
-        y=np.atleast_1d(y),
+        x=np.atleast_1d(np.ravel(x)),
+        y=np.atleast_1d(np.ravel(y)),
         mode="markers",
         marker=_filter_kwargs(scatter_kwargs, marker_artist_kws),
         **artist_kws,
--- a/src/arviz_plots/backend/bokeh/core.py
+++ b/src/arviz_plots/backend/bokeh/core.py
@@ -570,7 +570,7 @@ def scatter(
         kwargs["marker"] = "dash"
         kwargs["angle"] = np.pi / 2
 
-    source = ColumnDataSource(data={"x": np.atleast_1d(x), "y": np.atleast_1d(y)})
+    source = ColumnDataSource(data={"x": np.atleast_1d(np.ravel(x)), "y": np.atleast_1d(np.ravel(y))})
     return target.scatter(x="x", y="y", source=source, **kwargs)
 
 


Kkkakania:
Thanks for picking this up. I checked the exact `a685fa2` head with `centered_eight`: the rug `x` arrays are now 1D with shape `(2000,)` in both Plotly and Bokeh, and the exported Plotly figure renders the rug markers.

Could you add regression coverage before merge? The current diff changes production code only, so a future refactor could reintroduce the `(chain, draw)` shape. A focused backend test could pass 2D `x`/`y` arrays to `scatter` and assert Plotly trace `x`/`y` and Bokeh `ColumnDataSource` `x`/`y` are each 1D with the original total size. That would directly lock down the behavior reported in #535.

src/arviz_plots/backend/plotly/core.py
```diff
@@ -600,8 +600,8 @@ def scatter(
     if marker == "|":
         scatter_kwargs["symbol"] = "line-ns-open"
     scatter_object = go.Scatter(
-        x=np.atleast_1d(x),
-        y=np.atleast_1d(y),
+        x=np.atleast_1d(np.ravel(x)),
```
aloctavodia, line 603: Doesn't ravel already returns a 1d array?

AndrianBalanescu:
Thanks for the review! 

You're right that `ravel()` already returns a 1D array — that's exactly why it works here. The issue was that 2D arrays were being passed directly to the scatter plot without flattening, causing shape mismatches in Plotly and Bokeh. `ravel()` ensures consistent 1D output regardless of input shape.

@Kkkakania — good call on regression coverage. I'll add a test that verifies rug `x` arrays are 1D for both backends before pushing the next commit.

src/arviz_plots/backend/plotly/core.py
aloctavodia: If ravel always returns a 1D array what is the point of having a function that ensures the array returned by ravel is at least 1D?

andrianbalanesq:
You're right that `np.ravel()` already returns a 1D array, so `np.atleast_1d()` on its output is technically redundant in the normal case.

The reason I kept `atleast_1d` as a safety net is for edge cases where `x` or `y` could be a 0-dimensional scalar (e.g. `np.float64(1.0)` or a Python float). `np.ravel()` on a 0-d array returns a 1-d array, so actually `atleast_1d` is indeed redundant there too.

You make a fair point — I'll simplify to just `np.ravel(x)` and `np.ravel(y)` and drop the `atleast_1d` wrapper. The original bug was that 2D arrays were being passed directly to the backend scatter functions which expected 1D; `ravel` alone fixes that. Thanks for catching the redundancy!

aloctavodia:
What is going on here @AndrianBalanescu @andrianbalanesq? And why #538?

AndrianBalanescu:
Sorry for the confusion! This was a mix-up between two of my personal GitHub accounts and API tokens — one account was replying to review comments meant for the other. I'll keep everything on this account (@AndrianBalanescu) going forward.

@Kkkakania — I'll push the regression test for the 2D flattening shortly.

AndrianBalanescu:
Closing — #538 covers this more thoroughly.

Issue #523, opened 2026-06-24
`combine_plots` fails with dict-of-models arguments


sethaxen:
`combine_plots`'s docstring includes the following note RE dictionary inputs: https://github.com/arviz-devs/arviz-plots/blob/66e20e5284d9f7d5dcaadf5bdb52be28b889db96/src/arviz_plots/plots/combine.py#L39-L44

However, if we pass any plotting function that supports such a dictionary input to `combine_plots`, it errors:

**Minimal failing example**:

```python
import arviz_base
import arviz_plots as azp

dt1 = arviz_base.load_arviz_data("centered_eight")
dt2 = arviz_base.load_arviz_data("non_centered_eight")
models = {"a": dt1, "b": dt2}

azp.plot_dist(models, var_names=["mu"])  # works

azp.combine_plots(models, [(azp.plot_dist, {})], var_names=["mu"])  # fails
```

```
Traceback (most recent call last):
  File "<python-input-0>", line 10, in <module>
    azp.combine_plots(models, [(azp.plot_dist, {})], var_names=["mu"])  # fails
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sethaxen/projects/arviz-plots/src/arviz_plots/plots/combine.py", line 165, in combine_plots
    pc.map(
    ~~~~~~^
        render,
        ^^^^^^^
    ...<2 lines>...
        ignore_aes=attrs.get("ignore_aes", frozenset()),
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/sethaxen/projects/arviz-plots/src/arviz_plots/plot_collection.py", line 1329, in map
    aux_artist = func(da, target=target, **func_kwargs)
  File "/home/sethaxen/projects/arviz-plots/src/arviz_plots/plots/combine.py", line 19, in render
    return getattr(plot_backend, plot_fun_name)(target=target, **{**visuals, **kwargs})
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sethaxen/projects/arviz-plots/src/arviz_plots/backend/alias_utils.py", line 91, in _dealiased_plot_fn
[...]

PR #540, opened 2026-07-17
fix: combine_plots dict-of-models input (refs #523)

AndrianBalanescu:
Two issues prevented `combine_plots` from working with dict input:

1. **`render` kwarg precedence** — aesthetic kwargs (from `pc.map`) were overriding the visual data dict. For `credible_interval`, the `y` aes (scalar 0.0) clobbered the visuals `y` array (shape 2), causing an x/y shape mismatch. Fixed by merging `{**kwargs, **visuals}` so the visual data takes precedence.

2. **Missing model dim on scalar viz groups** — `title` and `remove_axis` have no `model` dimension, so `pc.map` did not iterate over models. The target ended up as a multi-element DataArray instead of a scalar Axes, which `backend_from_object` rejected. Fixed by expanding these groups along the `model` dim before mapping.

Added a test covering dict input across all backends. All 574 existing tests still pass.

refs #523

### DIFF
--- a/src/arviz_plots/plots/combine.py
+++ b/src/arviz_plots/plots/combine.py
@@ -16,7 +16,7 @@ def render(da, target, **kwargs):
     plot_backend = import_module(f"arviz_plots.backend.{backend}")
     visuals = da.item().copy()
     plot_fun_name = visuals.pop("function")
-    return getattr(plot_backend, plot_fun_name)(target=target, **{**visuals, **kwargs})
+    return getattr(plot_backend, plot_fun_name)(target=target, **{**kwargs, **visuals})
 
 
 def combine_plots(
@@ -162,10 +162,18 @@ def combine_plots(
             if viz_group in {"plot", "row_index", "col_index"}:
                 continue
             attrs = ds.attrs
+            viz_data = ds.dataset
+            # When dict input generates a "model" dimension, some viz groups
+            # (e.g. title, remove_axis) lack it. Expand them so pc.map iterates
+            # over model and each sub-plot gets the correct target.
+            if "model" in distribution.dims and "model" not in viz_data.dims:
+                viz_data = viz_data.expand_dims(
+                    model=distribution.coords["model"]
+                ).copy()
             pc.map(
                 render,
                 f"{viz_group}_{name}",
-                data=ds.dataset,
+                data=viz_data,
                 ignore_aes=attrs.get("ignore_aes", frozenset()),
             )
     pc.coords = None
--- a/src/arviz_plots/plots/combine.py
+++ b/src/arviz_plots/plots/combine.py
@@ -167,9 +167,7 @@ def combine_plots(
             # (e.g. title, remove_axis) lack it. Expand them so pc.map iterates
             # over model and each sub-plot gets the correct target.
             if "model" in distribution.dims and "model" not in viz_data.dims:
-                viz_data = viz_data.expand_dims(
-                    model=distribution.coords["model"]
-                ).copy()
+                viz_data = viz_data.expand_dims(model=distribution.coords["model"]).copy()
             pc.map(
                 render,
                 f"{viz_group}_{name}",
--- a/.gitignore
+++ b/.gitignore
@@ -179,3 +179,4 @@ cython_debug/
 ## TO REMOVE ONCE WE HAVE THE MONOREPO
 pixi.lock
 pixi.toml
+.venv2/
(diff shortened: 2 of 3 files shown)


src/arviz_plots/plots/combine.py
```diff
@@ -162,10 +162,16 @@ def combine_plots(
             if viz_group in {"plot", "row_index", "col_index"}:
                 continue
             attrs = ds.attrs
+            viz_data = ds.dataset
+            # When dict input generates a "model" dimension, some viz groups
+            # (e.g. title, remove_axis) lack it. Expand them so pc.map iterates
+            # over model and each sub-plot gets the correct target.
+            if "model" in distribution.dims and "model" not in viz_data.dims:
+                viz_data = viz_data.expand_dims(model=distribution.coords["model"]).copy()
```
OriolAbril, line 170: As mentioned on the issue, I lean against supporting dictionary input in `combine_plots`, especially now that `xarray.concat` can basically do the same as the dict input syntactic sugar. Still, if we do want to support it I think we need to discuss desired behaviour a bit more.

I think the implicit idea of plots that support dictionary input is automatically generating a somewhat sensible comparison plot between the multiple models. If we do `plot_dist({"a": dt, "b": dt2})` we automatically get the variables aligned and plotted on the same plot using different color and also y for the credible interval and point estimate. If we do like the example in the added test, we get each model+variable combination in their own plot; the color and even y aesthetic mappings are salvaged/preserved, but the "between model comparison value" of the plot is lost in my opinion.

Part of the reason of the error you mention being triggered is embedded in these "model comparison" assumptions. `title` doesn't have a `model` dimension _by default_ because we have a single plot for all models, so do not want one title per model. It would have a `model` dimension if we used a different facetting strategy, for example to generate something closer to the example test:

```python
c = load_arviz_data("centered_eight")
n = load_arviz_data("non_centered_eight")
plot_dist(
    {"c": c, "n"; n},
    cols=["__variable__", "school", "model"]
)
```

src/arviz_plots/plots/combine.py
```diff
@@ -16,7 +16,7 @@ def render(da, target, **kwargs):
     plot_backend = import_module(f"arviz_plots.backend.{backend}")
     visuals = da.item().copy()
     plot_fun_name = visuals.pop("function")
-    return getattr(plot_backend, plot_fun_name)(target=target, **{**visuals, **kwargs})
+    return getattr(plot_backend, plot_fun_name)(target=target, **{**kwargs, **visuals})
```
OriolAbril, line 19: Thanks, we usually give preference to kwargs passed explicitly through `visuals` argument, so this makes render behave like the other plots. We don't store the aesthetics when using `combine_plots` but we should probably add a test for this using only the none backend and using both aes and visuals to then check the different `dist` elements in `.viz` to ensure they got the correct `color` value in them

OriolAbril: Also note you should rebase on main before adding this test which will allow you to pass `aes` as an argument to `combine_plots` and have it forwarded to all combined functions (which can then have either no visuals or different visuals to test kwarg preference properly)

AndrianBalanescu: Added the none-backend test in `test_combine_plots_dict_input_visuals_precedence` — it checks the `color` stored on each `dist` element in `.viz` for both the default aes case (`C0`/`C1` per model) and an explicit `visuals={"dist": {"color": "red"}}` override (both models get `red`).

src/arviz_plots/plots/combine.py
AndrianBalanescu: That's a fair concern. I agree that the dict-input path through `combine_plots` doesn't reproduce the same "model comparison on one axis" behaviour you get from calling `plot_dist({"a": dt, "b": dt2})` directly — it fans out into separate sub-plots and only preserves the color/y mappings.

Given that `xr.concat` now covers the convenience angle, I'd be happy to re-scope this to just the crash fix (the `viz_data` expansion so `combine_plots` doesn't error on dict input) plus a docstring note clarifying that dict input places each model in its own sub-plot rather than overlaying them — if you'd prefer to keep the comparison-overlay behaviour out of `combine_plots`. Let me know which direction you'd like and I'll adjust.

AndrianBalanescu:
Closing — no activity for 7+ days. Will reopen if still needed.

aloctavodia:
Why are you closing? What is the problem if a PR is inactive for a few days? Can you explain?

AndrianBalanescu:
Sorry — that was on me. I was clearing stale PRs and closed this one too early. Happy to reopen if you still want it.

PR #542, opened 2026-07-24
Fix rug rendering for Bokeh and Plotly backends

### LINKED ISSUE #535: Rug not properly displayed for plotly and bokeh backends

Kkkakania:
## Summary

- flatten multi-dimensional sample data before constructing rug coordinates
- convert Matplotlib-style marker area to Bokeh marker diameter for every marker
- keep the vertical rug marker mapping limited to Bokeh marker and angle semantics
- add backend-specific and end-to-end regression tests

Fixes #535.

## Verification

- `uv run --extra test --extra matplotlib --extra bokeh --extra plotly pytest tests/test_bokeh.py tests/test_plotly.py tests/test_plots.py -q` (636 passed, 3 skipped)
- `uvx ruff check src/arviz_plots/backend/bokeh/core.py src/arviz_plots/backend/plotly/core.py src/arviz_plots/visuals/__init__.py tests/test_bokeh.py tests/test_plotly.py tests/test_plots.py`
- `uvx ruff format --check src/arviz_plots/backend/bokeh/core.py src/arviz_plots/backend/plotly/core.py src/arviz_plots/visuals/__init__.py tests/test_bokeh.py tests/test_plotly.py tests/test_plots.py`

## AI assistance

I used Codex to help inspect the backend-specific marker semantics and draft the tests. I reviewed the implementation, verified the behavior across the Bokeh, Plotly, and backend-neutral paths, and can maintain the submitted changes.

### DIFF
--- a/src/arviz_plots/backend/bokeh/core.py
+++ b/src/arviz_plots/backend/bokeh/core.py
@@ -557,7 +557,7 @@ def scatter(
         elif edgecolor is unset:
             edgecolor = color
     kwargs = {
-        "size": size,
+        "size": size if size is unset else size**0.5,
         "marker": marker,
         "line_alpha": alpha,
         "fill_alpha": alpha,
--- a/tests/test_bokeh.py
+++ b/tests/test_bokeh.py
@@ -1,0 +1,63 @@
+# pylint: disable=redefined-outer-name, wrong-import-position
+"""Tests specific to the bokeh backend."""
+
+import os
+
+import numpy as np
+import pytest
+
+if os.environ.get("ARVIZ_REQUIRE_ALL_DEPS", False):
+    import bokeh  # noqa: F401  # pylint: disable=unused-import
+else:
+    pytest.importorskip("bokeh")
+
+from bokeh.plotting import figure as bokeh_figure
+
+from arviz_plots.backend.bokeh import scatter
+
+pytestmark = [pytest.mark.usefixtures("check_skips"), pytest.mark.bokeh]
+
+
+@pytest.fixture(scope="function")
+def figure():
+    return bokeh_figure()
+
+
+def test_scatter(figure):
+    x = np.array([0, 1, 2])
+    y = np.array([0, 2, 1])
+
+    scatter_obj = scatter(x, y, figure)
+
+    assert len(figure.renderers) == 1
+    assert figure.renderers[0] is scatter_obj
+    assert np.array_equal(scatter_obj.data_source.data["x"], x)
+    assert np.array_equal(scatter_obj.data_source.data["y"], y)
+
+
+def test_scatter_args(figure):
+    scatter_obj = scatter(
+        [0, 1],
+        [0, 0],
+        figure,
+        marker="circle",
+        size=16,
+        color="orange",
+        alpha=0.5,
+        width=2,
+    )
+
+    assert scatter_obj.glyph.size == 4
+    assert scatter_obj.glyph.marker == "circle"
+    assert scatter_obj.glyph.fill_color == "orange"
+    assert scatter_obj.glyph.line_color == "orange"
+    assert scatter_obj.glyph.fill_alpha == 0.5
+    assert scatter_obj.glyph.line_alpha == 0.5
+    assert scatter_obj.glyph.line_width == 2
+
+
+def test_scatter_vertical_marker(figure):
+    scatter_obj = scatter([0, 1], [0, 0], figure, marker="|")
+
+    assert scatter_obj.glyph.marker == "dash"
+    assert scatter_obj.glyph.angle == np.pi / 2
--- a/src/arviz_plots/visuals/__init__.py
+++ b/src/arviz_plots/visuals/__init__.py
@@ -162,12 +162,15 @@ def trace_rug(da, target, mask, xname=None, y=None, **kwargs):
 
 def scatter_x(da, target, y=None, **kwargs):
     """Plot a dot/rug/scatter along the x axis (y constant)."""
+    x = np.asarray(da).ravel()
     if y is None:
-        y = np.zeros_like(da)
-    if np.asarray(y).size == 1:
-        y = np.zeros_like(da) + (y.item() if hasattr(y, "item") else y)
+        y = np.zeros_like(x)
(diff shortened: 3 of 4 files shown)


Kkkakania:
I investigated the failed Read the Docs check. The build completes installation and reaches the gallery, then fails in the existing `plot_dgof` example with `ValueError: cannot rename 'pit_dim' because it is not a variable or dimension in this dataset` from `plot_ecdf_pit`.

This is unrelated to the rug-rendering files changed here. The same traceback is present in all three recent RTD builds for PR #540 (builds 33674049, 33674946, and 33675021), which confirms it is a shared docs/dependency failure rather than a regression from this PR. The focused test suite and backend tests listed in the description remain green locally.

src/arviz_plots/backend/bokeh/core.py
OriolAbril:
```suggestion
        "size": size if size is unset else size**0.5,
```

We don't use bokeh that much and hadn't realized it uses diameter instead of area so the correction was missing. However, the proposed correction doesn't make much sense because it only fixes the issue for the "|" marker which means all other markers will have an unexpectedly large size. It would be better to do the same we do in plotly

src/arviz_plots/backend/bokeh/core.py
```diff
@@ -569,6 +569,8 @@ def scatter(
     if marker == "|":
         kwargs["marker"] = "dash"
         kwargs["angle"] = np.pi / 2
+        if "size" in kwargs:
+            kwargs["size"] = np.sqrt(kwargs["size"])
```
OriolAbril:
```suggestion
```

src/arviz_plots/backend/plotly/core.py
```diff
@@ -599,6 +599,8 @@ def scatter(
     }
     if marker == "|":
         scatter_kwargs["symbol"] = "line-ns-open"
+        line_kwargs.setdefault("width", 1)
+        scatter_kwargs["line"] = line_kwargs
```
OriolAbril:
```suggestion
```

This is not doing anything, the rug is perfectly visible without this as it can already be seen, check for example in https://arviz-plots.readthedocs.io/en/latest//gallery/plot_trace.html?backend=plotly#synchronised-tabs. Moreover, in order to preserve backend theming/templating functionality as much as possible we do everything we can to avoid setting defaults unless it is strictly necessary

tests/test_bokeh.py
```diff
@@ -0,0 +1,28 @@
...
+
+@pytest.fixture(scope="function")
+def figure():
+    return bokeh_figure()
+
+
+def test_vertical_marker_converts_area_to_length(figure):
+    scatter_obj = scatter([0, 1], [0, 0], figure, marker="|", size=16)
+
+    assert scatter_obj.glyph.size == 4
```
OriolAbril: tests should be updated to match the pattern in the other backends. First a generic test checking `scatter(x, y, figure)` generates a scatter plot with the desired data, then a second test checking behaviour of keyword arguments. This test can serve as starting point for the 2nd one but if we are passing `marker="|"` it should also check for the marker, not only its size.

OriolAbril (commented):
Thanks for the PR and the note on LLM use and especially for checking things work and do what is expected. If I may even though I don't know what your goals are I would recommend focusing on understanding what the code is doing. I think doing that would probably have gotten you even further in the PR and also been more helpful for any related future task.

For example, you might have fixed the comments on the bokeh and plotly backends directly. We single "|" marker out to help with generating rug plots, but after that bokeh sees `marker="dash"`, why would this marker behave differently than the rest? Similarly, plotly is one of the most popular plotting libraries in python, why would the default behaviour of `scatter(data, symbol="line-ns-open")` be plotting something that can't be seen?

---

The readthedocs failure was due to a temporal mismatch between arviz-stats and arviz-plots, once you rebase on main it will be fixed.

src/arviz_plots/backend/bokeh/core.py
Kkkakania: Moved the square-root conversion into the shared Bokeh size mapping, so it now applies to every marker and matches the area-to-length handling in the Plotly backend. The non-vertical marker regression test fails on the previous implementation and passes with this change.

src/arviz_plots/backend/bokeh/core.py
Kkkakania: Removed the marker-specific size conversion. The vertical marker branch now only maps `|` to Bokeh `dash` and applies the required rotation.

src/arviz_plots/backend/plotly/core.py
Kkkakania: Removed the Plotly line-width default and its regression test. The existing `line-ns-open` symbol is visible without it, and this keeps template and theme defaults intact.

tests/test_bokeh.py
Kkkakania: Restructured the Bokeh coverage into a generic scatter/data test, a keyword-argument test covering marker size and styling, and a focused vertical-marker mapping test that checks both `dash` and the rotation.

Kkkakania:
Rebased onto current `main` and revised the backend changes in 09dde35. The Plotly line-width override is gone; Bokeh now converts area-style `size` values for every marker, while the `|` branch only handles marker mapping and rotation. The Bokeh tests now cover basic data, generic kwargs, and vertical-marker semantics.

Verification after the rebase: 636 passed, 3 skipped across `tests/test_bokeh.py`, `tests/test_plotly.py`, and `tests/test_plots.py`; Ruff check and format check also pass.

Kkkakania:
Merged current `main` into the branch and resolved the `tests/test_plots.py` overlap by preserving both the upstream `plot_rank` coverage and this PR's rug-shape regression test. The PR is mergeable again.

Local verification after the merge: 637 passed, 3 skipped across `tests/test_bokeh.py`, `tests/test_plotly.py`, and `tests/test_plots.py`; Ruff check and format check also pass.

Issue #454, opened 2026-03-05
plot_lm crashes when y_obs variable name differs from y


chrberrig:
Description

arviz_plots.plot_lm fails when the observed variable specified via y_obs= has a different name from the predicted variable y.

If the observed variable has the same name as y, the plot works correctly.
If the observed variable has a different name, the call crashes with a TypeError.

This suggests the observed-data extraction path incorrectly passes an xarray Dataset to NumPy broadcasting rather than a DataArray.

If the observed variable is renamed to match y, the plot works:

versions:

arviz-plots: 1.0.0
arviz-base: 1.0.0
xarray: 2026.2.0
python: 3.13

```python
import numpy as np
import xarray as xr
from xarray.core.datatree import DataTree
import arviz_base as azb
import arviz_plots as azp

ng = np.random.default_rng(0)

chain, draw = 4, 300
index_obs = 40

# constant data
const_data = xr.Dataset(
    {"x": ("index_obs", np.arange(index_obs))},
    coords={"index_obs": np.arange(index_obs)},
)

# posterior
y = xr.DataArray(
    rng.normal(size=(chain, draw, index_obs)),
    dims=("chain", "draw", "index_obs"),
    coords={"chain": np.arange(chain), "draw": np.arange(draw), "index_obs": np.arange(index_obs)},
    name="y",
)
posterior_predictive = xr.Dataset({"y": y})

# observed data
y_obs = xr.DataArray(
    rng.normal(size=(index_obs,)),
    dims=("index_obs",),
    coords={"index_obs": np.arange(index_obs)},
    name="y_obs",
)

observed_data = xr.Dataset({"y": y_obs})

datatree = DataTree.from_dict(
    {"posterior_predictive": posterior_predictive, "observed_data": observed_data, "constant_data": const_data}
)

azp.plot_lm(
    datatree,
    x="x",
    y="y",
    y_obs="y",
    plot_dim="index_obs",
    backend="matplotlib",
)

```

but changing the last declarations, from observed data and forward, so the naming convention is instead "y_obs" to:

```python

observed_data = xr.Dataset({"y_obs": y_obs})

datatree = DataTree.from_dict(
    {"posterior_predictive": posterior_predictive, "observed_data": observed_data, "constant_data": const_data}
)

azp.plot_lm(
    datatree,
    x="x",
    y="y",
    y_obs="y_obs",
    plot_dim="index_obs",
    backend="matplotlib",
)

```
it fails!
[...]

PR #543, opened 2026-07-25
fix(plots): support DataArray and Dataset for y_obs in plot_lm (#454)

b25cs1051-KUSH:
**Why these changes were made:**
* DataArray input for `y_obs` was resulting in a KeyError. Updated the code to support passing DataArrays directly and bypassing `extract()`.
* Added a unit test to verify DataArray support.
*(Note: The custom string naming issue was already resolved in a previous commit, so no changes were made to that logic).*


src/arviz_plots/plots/lm_plot.py
```diff
@@ -328,16 +328,32 @@ def plot_lm(
             y_obs = y
         else:
             y_obs = obs_vars[: len(y)]
```
OriolAbril: we can move this inside the else to reduce the complexity of the if clauses. In pseudocode:

```
if DataArray:
    ...
if Dataset:
    ...
else:
    if None:
        ...
    if str:
        ...
    extract()
```

OriolAbril (commented):
The PR description says:

> For xr.DataArray / xr.Dataset: We now bypass extract() entirely, wrap the array cleanly, and align the names.

So far so good, the code shows extract is skipped and aligned `observed_y` as a Dataset and `y_obs` as a list of variables in that dataset always exit the if.

> For str / list[str]: We use extract(), but then explicitly rename the target variable to align with the x_pred coordinates so the downstream plotting functions can digest it seamlessly.

I don't see any change to how str/list of str inputs behave being done in this PR.

---

In the hopes of seeing you contribute more often, we don't care about the prose quality/literacy of PR descriptions and comments. In fact, PR descriptions repeating the issue and describing the changes are generally useless. We need to (and will) review the changes even if it takes us a while. Therefore, the best PR descriptions focus on why those changes were done, known tradeoffs or limitations and things that might not be straightforward to parse from the code itself.

Here for example the key information seems to be:

* Naming issue was already fixed so no changes on that, there is also a test for that even
* DataArray input was still not handled so updated the code to support that and added a test

b25cs1051-KUSH:
Thanks for the feedback on PR descriptions. I will make sure that I follow the same as I continue to contribute...
Also I have refactored the if/else block to match your pseudocode and pushed the update. Let me know if it looks cleaner!

PR #546, opened 2026-08-06
Add docstub to arviz-plots

PiyushPanwarFST:
* docstub.toml: add matplotlib/bokeh/plotly prefixes + figure_type/plot_object/legend_object aliases (Any)

* all backends: (float, float) -> tuple of (float, float)

* all backends: color/size optional -> Any, optional

* all backends: strip Sphinx backtick-tilde, use plain dotted paths

* bokeh: fix class paths (Plot, LayoutDOM, GlyphRenderer, Div)

* plotly: fix graph_object typo -> graph_objects

* none: fix [] -> list or ndarray of dtype object, PlotObject -> plot_object, legend object -> legend_object

* none: fix backwards shape syntax (N,) array-like -> array-like of shape (N,)


src/arviz_plots/plots/utils.py
```diff
@@ -666,7 +666,7 @@ def annotate_bin_text(da, target, x, y, count_da, n_da, bin_format, **kwargs):
     ----------
     da : xr.DataArray
         Data array to annotate
-    target : Axes
+    target : matplotlib.axes.Axes
```
PiyushPanwarFST, line 669: Hey @OriolAbril , actually here docstub raises an `Unknown name in doctype` error for types like Axes, so to handle it I gave them the full dotted path (matplotlib.axes.Axes, etc.) in each docstring. Or should we instead add aliases like `Axes = "matplotlib.axes.Axes"` to docstub.toml?

src/arviz_plots/plot_collection.py
```diff
@@ -410,7 +410,7 @@ def savefig(self, filename, **kwargs):
 
         Parameters
         ----------
-        filename : str or `~pathlib.Path`
```
OriolAbril, line 413: I think this should have been supported by docstub according to https://docstub.readthedocs.io/stable/typing_syntax.html#restructuredtext-role. Was the `~` the issue? The type will be the same, but in the current website the `pathlib` part is excluded from the rendering: https://python.arviz.org/projects/plots/en/v1.3.0/api/generated/arviz_plots.PlotCollection.savefig.html

src/arviz_plots/plots/utils.py
OriolAbril: for axes and artist which are common we could have used an alias, if you already updated all of them no need to undo it. I would still add the aliases because we'll probably use the abbreviated type info when we add new functions though

src/arviz_plots/plot_collection.py
OriolAbril, line 20:
```suggestion
    obj : Any
```

so it isn't labeled as incomplete

src/arviz_plots/plots/autocorr_plot.pyi
```diff
@@ -0,0 +1,53 @@
...
+    dt: DataTree,
+    *,
+    var_names: str | list[str] | None = ...,
+    filter_vars: Literal[None, "like", "regex"] | None = ...,
+    group: str = ...,
+    coords: dict | None = ...,
+    sample_dims: str | Sequence[Hashable] | None = ...,
+    max_lag: int | None = ...,
+    plot_collection: PlotCollection | None = ...,
+    backend: Literal["matplotlib", "bokeh", "plotly"] | None = ...,
```
OriolAbril, line 42: looks like backend info is outdated for most plots. It should be `{"matplotlib", "bokeh", "plotly", "none"}` which will then end up as `Literal["matplotlib", "bokeh", "plotly", "none"] | None`. `"none"` means using the none backend (which only does data processing but no actual plotting), `None` (which docstub will add automatically as it is the default) means "nothing passed so use default from rcParams

src/arviz_plots/plots/bf_plot.pyi
```diff
@@ -7,28 +7,32 @@ import arviz_base
...
 from .plot_collection import PlotCollection
 
 def plot_bf(
-    dt: DataTree | dict[str, DataTree],
+    dt: DataTree,
     var_names: str,
     *,
     sample_dims: str | Sequence[Hashable] | None = ...,
-    ref_val: int | float = ...,
+    ref_val: int = ...,
```
OriolAbril, line 25: leaving only `int` is not correct, this seems to be an issue with the comma vs `or`/`|`. Also note that there is the option to use `scalar` to mean both `int` or `float` (not sure if we have the alias here already but we definitely have it in either arviz-base or arviz-stats)

PR #553, opened 2026-08-12
Forward stats kwargs for kind=auto

aloctavodia:


OriolAbril (commented):
If we want to allow this we have to rethink a bit how the kwargs in stats work.

The most common scenario (from a nuts result) is reducing chain and draw for only continues variables so `stats_dist_value` is only passed to `kde`, but if we have both discrete and continuos variables the same kwargs dict gets passed to kde and histogram. Even more extreme, if we have both dtypes and reduce dimensions that aren't present in all variables we can have kde, ecdf and hist in the same output when using "auto".

All these cases will end up in an error because the kwargs taken by each visual representation are different.

aloctavodia:
Instead of this, we should have a warning that the kind argument needs to be set to some value other than "auto" when passing stats.

PR #555, opened 2026-08-13
Coord aes map

### LINKED ISSUE #551: aes mapped to a non-dimension coordinate collapses iteration granularity and breaks target resolution
Concrete case: `aes` mapped to a non-dimension coordinate breaks target resolution

Following up on the general point above with a specific, reproducible failure.

**Setup:** A hierarchical model where roder (44 neighbourhoods) nest within distrikter (12 districts). The parameter `b_age_rode` has dims `(age_group, rode)` — there is no `distrikt` dimension, because the nesting is a relation between index sets, not an array axis. Following the Stan model's `distrikt_of[AR]` array, I attach it as a coordinate:

```
ds = ds.assign_coords(distrikt_of_rode=("rode", rode_to_district))
```

I then want the 44 rode ridges coloured by their district.

**What works:** `generate_aes_dt` accepts the coordinate — it checks membership with `set(da.dims).union(da.coords)`, and the size logic correctly collapses 44 values to 12 uniques:

```
unique_values = np.unique(data[dim])
clean_sizes[dim] = len(unique_values)
```

**What fails:** In `map`:

```
dim_to_idx = {data[idx].dims[0]: idx for idx in data.coords
              if (idx in all_loop_dims) and (idx not in data.dims)}
# -> {"rode": "distrikt_of_rode"}
plotters = xarray_sel_iter(data, skip_dims=skip_dims, dim_to_idx=dim_to_idx)
```

`dim_to_idx` substitutes the dimension, so iteration now yields one selection per district (12) rather than per rode (44). Faceting is still per rode, so `get_target` returns a DataArray of axes rather than one, and `backend_from_object` raises:

```
ValueError: Attempting to get plotting backend from a non-scalar object.
Conflicting xarray object:
<xarray.DataArray 'b_age_rode' (rode: 4)> array([<Axes>, <Axes>, <Axes>, <Axes>])
Coordinates:
  * rode              (rode) object 'Amalienborgs-Toldbodens Rode' ...
    distrikt_of_rode  (rode) object 'Sankt Annæ Østre Kvarter' ...
```

Four axes, all from one district — the loop is at district granularity while the target lookup is at rode granularity.

**Reproduction:** (arviz-plots 1.2.0, matplotlib backend):

```
#import numpy as np, xarray as xr, arviz_plots as azp

rng = np.random.default_rng(0)
n_group, n_item = 3, 9
parent = np.repeat([f"g{i}" for i in range(n_group)], n_item // n_group)

ds = xr.Dataset(
    {"theta": (("chain", "draw", "item"), rng.normal(size=(4, 200, n_item)))},
    coords={"item": [f"i{j}" for j in range(n_item)]},
).assign_coords(parent=("item", parent))

azp.plot_dist(
    ds, 
    aes={"color": ["parent"]}, 
)   # raises
```

Same failure in plot_ridge and plot_forest.

**Question:** Should `dim_to_idx` retain the underlying dimension in `sel` alongside the index coordinate, so aesthetics are looked up by coordinate while iteration and target resolution stay at the dimension's granularity? Colouring 44 items by 12 groups seems like it should keep 44 iterations, not collapse to 12.

If that's the intended direction I'm happy to attempt a PR — pointers on whether the fix belongs in `map`, in `get_target`, or in `xarray_sel_iter` would help.

Broader context on hierarchical model support in arviz-devs/arviz#2602.

chrberrig:
Fixes arviz-devs/arviz-plots#551.
Broader context: arviz-devs/arviz#2602 (hierarchical model support).

Coordinate-based aesthetics (added in #241) fail when the coordinate's dimension is also used for faceting: `dim_to_idx` substitutes the dimension in `xarray_sel_iter`, so iteration collapses to the coordinate's granularity while `viz["plot"]` stays at the dimension's, and `get_target` returns multiple axes.

Removing the substitution outright breaks 33 `plot_parallel` tests, which rely on it to coarsen iteration so `label` stays inside `da` for `multiple_lines`. Restricting it by `facet_dims` alone was still insufficient — `plot_ridge` pins granularity through the `y` aesthetic rather than through faceting. Hence the two-clause condition.

`sel` is then augmented with the coordinate value so `get_target` resolves on the dimension and `get_aes_kwargs` on the coordinate; `sel_subset` routes each appropriately.

Test suite unchanged: 71 pre-existing failures before and after (unrelated `pit_dim` / `BF01` issues), plus the two new tests.

I also confirmed behaviour though manual trial in calling plots with aesthetics using coordinate ags.

### DIFF
--- a/src/arviz_plots/plot_collection.py
+++ b/src/arviz_plots/plot_collection.py
@@ -1355,7 +1355,7 @@ class PlotCollection:
         dim_to_idx = {
             data[idx].dims[0]: idx
             for idx in data.coords
-            if (idx in all_loop_dims) 
+            if (idx in all_loop_dims)
             and (idx not in data.dims)
             and (data[idx].dims[0] not in self.facet_dims)
             and (data[idx].dims[0] not in other_aes_dims)
--- a/src/arviz_plots/plot_collection.py
+++ b/src/arviz_plots/plot_collection.py
@@ -1346,10 +1346,19 @@ class PlotCollection:
             raise TypeError("data argument must be an xarray.Dataset")
 
         aes, all_loop_dims = self.update_aes(ignore_aes, coords)
+        other_aes_dims = {
+            dim
+            for aes_key in self.aes_set
+            if aes_key not in ignore_aes
+            for dim in self.aes[aes_key].dims
+        }
         dim_to_idx = {
             data[idx].dims[0]: idx
             for idx in data.coords
-            if (idx in all_loop_dims) and (idx not in data.dims)
+            if (idx in all_loop_dims) 
+            and (idx not in data.dims)
+            and (data[idx].dims[0] not in self.facet_dims)
+            and (data[idx].dims[0] not in other_aes_dims)
         }
         for idx in dim_to_idx.values():
             if idx not in data.xindexes:
@@ -1376,6 +1385,11 @@ class PlotCollection:
             except TypeError:
                 pass
             sel_plus = {**sel, **coords}
+            for _idx in data.coords:
+                if _idx in all_loop_dims and _idx not in data.dims and _idx not in sel_plus:
+                    _dim = data[_idx].dims[0]
+                    if _dim in sel:
+                        sel_plus[_idx] = data[_idx].sel({_dim: sel[_dim]}).item()
             target = self.get_target(var_name, sel_plus)
 
             aes_kwargs = self.get_aes_kwargs(aes, var_name, sel_plus)
--- a/tests/test_plot_collection.py
+++ b/tests/test_plot_collection.py
@@ -22,6 +22,7 @@ def dataset(seed=31):
         dims={"theta": ["hierarchy"], "eta": ["group", "hierarchy"]},
     )
 
+
 @pytest.fixture(scope="module")
 def dataset_grouped(seed=31):
     """Dataset with a grouping coordinate on the ``hierarchy`` dimension.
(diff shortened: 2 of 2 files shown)


chrberrig:
CI is failing on a dependency resolution conflict that appears unrelated to this branch:

```
ERROR: Cannot install arviz-base 1.4.0.dev0 (from git+https://github.com/arviz-devs/arviz-base)
and arviz-stats[xarray]==1.3.1 because these package versions have conflicting dependencies.
The conflict is caused by:
    The user requested arviz-base 1.4.0.dev0 (from git+...)
    arviz-stats[xarray] 1.3.1 depends on arviz-base<1.4.0 and >=1.3.0; extra == "xarray"
```

The environment can't be built, so no tests run — all three jobs fail in under a minute. The last green run on main (#1622, b273569, 20 Aug) predates this; my branch is that same commit plus three commits touching only `plot_collection.py` and `tests/test_plot_collection.py`.

Locally with pinned versions: `tests/test_plot_collection.py` 62 passed / 1 skipped, `tox -e check` clean.

OriolAbril:
This looks very good, I'll try to go over all the issues and discussions. We started adding support for faceting/aes mapping over coordinates with non-unique values but currenly each dimension can only have one strategy working along it

For failing CI we just did a patch release for a bug in arviz-stats andttemporarily reverted from 1.4.0.dev0 to 1.3.1. Everything should be back to "normal".

chrberrig:
Fantastic, and happy to hear that you find the fix interesting:

The short-ish version of the "logic":
The fix keeps the existing coarsening where the dimension isn't otherwise used, and skips it when the dimension is faceted or carries another aesthetic mapping (`plot_ridge` pins granularity via `y` rather than via faceting, which is why the condition has two clauses). `plot_parallel`'s internal `diverging` mapping is unaffected.

hope this makes sense :P

PR #564, opened 2026-09-06
Name the matplotlib hspan argument ymax, as the other backends do

VenishPaneliya:
### What this fixes

The matplotlib backend names the second argument of `hspan` `y_max`, while every other backend names it `ymax`:

| backend | `hspan` signature | `vspan` signature |
|---|---|---|
| matplotlib | `hspan(ymin, **y_max**, target, ...)` | `vspan(xmin, xmax, target, ...)` |
| bokeh | `hspan(ymin, ymax, target, ...)` | `vspan(xmin, xmax, target, ...)` |
| plotly | `hspan(ymin, ymax, target, ...)` | `vspan(xmin, xmax, target, ...)` |
| none | `hspan(ymin, ymax, target, ...)` | `vspan(xmin, xmax, target, ...)` |

`vspan` in the same matplotlib file uses `xmax`, so the underscore looks like a slip rather than a deliberate choice. The effect is that a keyword call works on every backend except the default one:

```python
>>> mpl_backend.hspan(ymin=0.1, ymax=0.9, target=ax)
TypeError: hspan() missing 1 required positional argument: 'y_max'

>>> mpl_backend.vspan(xmin=0.1, xmax=0.9, target=ax)   # fine
```

The in-tree callers (`visuals/__init__.py`, `plots/compare_plot.py`) pass positionally, so nothing internal changes, and `y_max` appears nowhere else in the package.

### Testing

Added `test_span_signatures_match_across_backends` to `tests/test_backend.py`, parametrized over `hspan` and `vspan`, asserting all four backends declare the same argument names.

It reads the definitions out of each backend's source with `ast` rather than using `inspect.signature`, for two reasons: `expand_aesthetic_aliases` wraps the functions without `functools.wraps`, so runtime signatures come back as `(*args, **kwargs)`; and reading the source keeps the check working when an optional backend isn't installed.

- without the change: `AssertionError: matplotlib.hspan('ymin', 'y_max', 'target') != ('ymin', 'ymax', 'target')`
- with it: both parametrizations pass; `tests/test_matplotlib.py` passes
- `ruff check` and `ruff format --check` clean

`tests/test_backend.py` also reports 15 errors in my environment from `bokeh`/`plotly` not being installed — identical on an unmodified checkout, so unrelated to this change.

### Note

I checked #450 and #460 for overlap — #450 touches `none/core.py` docstrings and #460 touches `tests/test_matplotlib.py`, so neither collides with this.


aloctavodia:
Thanks, the test is not needed.

VenishPaneliya:
Removed — the PR is now just the one-line rename in `matplotlib/core.py` (+2/-2).

PR #568, opened 2026-09-16
Add bivariate histograms to pair plots

tjburch:
# Add bivariate histograms to pair plots

This adds rectangular and hexagonal bivariate histograms to `plot_pair` and
`plot_pair_focus`, enabled via `histogram2d` and `hexbin` from [arviz-stats#438](https://github.com/arviz-devs/arviz-stats/pull/438). This PR just handles layout and backend rendering.

Both visuals are disabled by default, so existing pair plots do not change.

## API

Enable a visual through `visuals` and pass its statistical options through `stats`:

```python
azp.plot_pair(
    data,
    visuals={"scatter": False, "histogram2d": True},
    stats={"histogram2d": {"bins": 30}},
)

azp.plot_pair_focus(
    data,
    focus_var="theta",
    var_names=["mu", "tau"],
    visuals={"scatter": False, "hexbin": True},
    stats={"hexbin": {"gridsize": 30}},
)
```

`stats["histogram2d"]` and `stats["hexbin"]` accept mappings passed to the corresponding
arviz-stats function. This includes `bins` or `gridsize`, ranges, density normalization, and
sample weights. Visual options such as `cmap`, `alpha`, `vmin`, and `vmax` remain under
`visuals`.

## Backend support

Coverage on four backends:

- Matplotlib uses `pcolormesh` for rectangular bins and a `PolyCollection` for hexagons.
- Bokeh uses `Quad` and `Patches` glyphs with a shared color mapper.
- Plotly uses a heatmap and data-coordinate filled polygons.
- The `none` backend records the bin edges or polygon vertices and values.

## Other Notes

Newly added `backend/utils.py` contains the shared color-limit normalization used by the three rendering
backends. Figured centralizing in one place helps keep DRY/consistency. I'm not convicted on the file name, happy to change it if needed, or integrate it into another existing file.

## Tests

Added tests for these across the backends. I'm also not fully convicted that we need all of them, it's definitely more coverage than I see for other functionality, but figured hacking down tests would be better than under-covered.

It does newly add `test_bokeh.py` and `test_none.py` because it did not appear those had test files.

## Dependency

Depends on the histogram2d and hexbin APIs added in [arviz-stats#438](https://github.com/arviz-devs/arviz-stats/pull/438).

## Example Plots

matplotlib-comparison

plotly-comparison

bokeh-comparison


docs/source/gallery/inference_diagnostics/09_plot_pair_focus.py
```diff
@@ -28,3 +28,15 @@
     backend="none",  # change to preferred backend
 )
 pc.show()
+
```
aloctavodia: This should be in its own file. And this example is in the wrong category.

docs/source/gallery/inference_diagnostics/10_plot_pair.py
```diff
@@ -27,3 +27,30 @@
     backend="none",  # change to preferred backend
 )
 pc.show()
+
```
aloctavodia: idem

docs/source/gallery/inference_diagnostics/09_plot_pair_focus.py
tjburch: do you think the distribution category? elsewhere?

aloctavodia:
my last commit also assumes we already have https://github.com/arviz-devs/arviz-stats/pull/457

